PR #23, opened 2025-09-25
fix(core): emit core.close before rejecting for error event

matthyk:
This PR ensures that the close action is called and awaitesd before the promise is rejected. This `thymian.ts` file must be ignored for `prettier` because the `with { type: 'json' }` is not supported in version <3.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,4 +3,4 @@
 /coverage
 /.nx/cache
 /.nx/workspace-data
-/core/src/thymian.ts
+
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,3 +3,4 @@
 /coverage
 /.nx/cache
 /.nx/workspace-data
+/core/src/thymian.ts
--- a/core/src/thymian.ts
+++ b/core/src/thymian.ts
@@ -1,19 +1,20 @@
 import { Subject } from 'rxjs';
 import semver from 'semver';
 
-// prettier-ignore
 import packageJson from '../package.json' with { type: 'json' };
 import { validate } from './ajv.js';
 import { corePlugin } from './core-plugin.js';
 import { ThymianEmitter } from './emitter/index.js';
-import {  ThymianFormat } from './format/index.js';
+import { ThymianFormat } from './format/index.js';
 import type { Logger } from './logger/logger.js';
 import { NoopLogger } from './logger/noop.logger.js';
 import { ThymianBaseError } from './thymian.error.js';
 import type { ThymianPlugin } from './thymian-plugin.js';
 import { timeoutPromise } from './utils.js';
 
-export type RegisteredPlugin<T extends Record<PropertyKey, unknown> = Record<PropertyKey, unknown>> = {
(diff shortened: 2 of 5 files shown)


core/src/thymian.ts
```diff
@@ -101,9 +102,15 @@ export class Thymian {
 
   run<T>(fn: (emitter: ThymianEmitter, logger: Logger) => Promise<T> | T): Promise<T> {
     return new Promise((resolve, reject) => {
+      const tryClose = (err: unknown)=> {
+        this.logger.debug('Try closing Thymian...');
```
Markus-Ende, line 123: was this left on purpose?

.prettierignore
```diff
@@ -2,4 +2,5 @@
 /dist
 /coverage
 /.nx/cache
-/.nx/workspace-data
\ No newline at end of file
+/.nx/workspace-data
+/core/src/thymian.ts
```
Markus-Ende: could you instead just ignore the specific part inline? (see https://prettier.io/docs/ignore/#javascript)

core/test/thymian.test.ts
```diff
@@ -110,4 +121,49 @@ describe('Thymian', () => {
...
+  it('should call close action before rejecting promise for error event', async () => {
+    const closedSpy = vitest.fn();
+
+    thymian.register(
+      createPluginFor(async (emitter) => {
+        emitter.onAction('core.close', (_, ctx) => {
+          setTimeout(() => {
+            closedSpy();
+            ctx.reply();
+          }, 1000);
```
Markus-Ende: does the test really have to wait for this? if so, it woul be a good idea to use the jest fake timer

core/test/thymian.test.ts
```diff
@@ -110,4 +121,49 @@ describe('Thymian', () => {
...
+
+    try {
+      await thymian.run(async () => {
+        await thymian.emitter.emitAction(
+          'core.run',
+          new ThymianFormat().export()
+        );
+      });
+    } catch (err) {
+      expect(closedSpy).toHaveBeenCalled();
```
Markus-Ende, line 162: this will never be reached, if no error is thrown at all.
quick fix would be to use done.fail after the await thymian.run

.prettierignore
matthyk: The prettier-ignore comment does not work because the `with { type: 'json' }` syntax causes a parsing error

Markus-Ende: I don't get it, I tried removing both this entry in .prettierignore and also the inline comment in the file, and the linting worked.. see https://github.com/thymianofficial/thymian/pull/33

Markus-Ende: ah, I see, the prettier plugin errors, and the document can't be formatted

Markus-Ende: found a workaround, by extrackting the json import into its own file, then the prettier parser seems to accept it. Will add this to this pr

Markus-Ende: no need for the workaround, I just updated prettier

Markus-Ende (approved):

PR #27, opened 2025-09-25
fix swagger detection and change peer dependencies to real ones

Markus-Ende:
This pull request updates how dependencies are managed in several `package.json` files across the monorepo. The main change is moving various internal packages from `peerDependencies` to `dependencies`, ensuring that required packages are installed automatically rather than relying on consumers to provide them. Additionally, there are minor improvements to build scripts and OpenAPI file detection logic.


distante (approved):
LGTM

PR #29, opened 2025-09-26
markdown report plugin

xquadradpi:
Plugin to create markdown report file


Markus-Ende:
@xquadradpi @matthyk wouldn't it be better to include this functionality into the existing reporter plugin? We could use a plugin option to set the output format.

matthyk:
> @xquadradpi @matthyk wouldn't it be better to include this functionality into the existing reporter plugin? We could use a plugin option to set the output format.

My first idea was to implement separate plugins for each output format but I see the advantages of a single report plugin. We should perhaps also discuss redesigning the report format as part of this PR.

Markus-Ende:
we can discuss in https://github.com/thymianofficial/thymian-internal/issues/47

matthyk:
Closed in favour of thymianofficial/thymian#60

PR #44, opened 2025-11-17
Websocket Proxy Plugin

matthyk:
Closes thymianofficial/thymian-internal#63


plugins/websocket-proxy/src/websocket-proxy-server.ts
```diff
@@ -0,0 +1,408 @@
...
+    this.wss.on('listening', () => {
+      this.logger.debug(
+        `@thymian/websocket-proxy listening on ws://localhost:${this.options.port}`,
+      );
+    });
+
+    return new Promise((resolve, reject) => {
+      this.wss.on('connection', async (ws) => {
+        try {
+          const connectedClient = await timeoutPromise(
```
atennert, line 88: In the future, this should be a connection that is open as long as the app runs for whoever wants to join. (Thinking of the server version :))

atennert (approved):

plugins/websocket-proxy/src/websocket-proxy-server.ts
matthyk: Clients can connect to the Websocket server as long as Thymian is running. The `start` method just waits for all `required` clients. If no client is `required` the promise is resolved immediatly. But the server still listens for new connections

atennert: ah ok, misread that then :D

PR #45, opened 2025-11-18
feat(cli): serve mode

matthyk:
Implements a `serve` mode via the `serve` command. This command loads all plugins and then listens for events and actions. The process can be closed in the following ways:

1. enter `q` in the terminal
2. emitting a `SIGINT` event to the process by pressing CRTL+C
3. emitting a `core.exit` event to Thymian. This possibility can be dangerous because any plugin can use this event to force the process to exit. Nevertheless, it is useful to let plugins have full control over Thymian. Happy to discuss about it.

### DIFF
--- a/cli/docs/commands/serve.md
+++ b/cli/docs/commands/serve.md
@@ -1,0 +1,12 @@
+---
+title: 'Serve'
+---
+
+The serve command runs Thymian as a long‑lived process so that you, your tools, and remote plugins can interact with it continuously. In this mode, Thymian boots the core, loads the configured plugins, and keeps running until you quit the session.
+This is the recommended mode when you want to drive Thymian from an external tool, an editor integration, a CI job, or any remote plugin connected via WebSockets.
+You can terminate the session at any time by pressing the `q` key in the terminal, or by sending a standard SIGINT (Ctrl+C). Internally, the process also listens for a `core.exit` event. When any plugin emits that event, Thymian shuts down gracefully and exits with the provided code.
+This means the lifecycle can be controlled from the keyboard, from scripts that signal the process, or from plugins that decide when the overall task is finished.
+In serve mode, all configured plugins are active for the entire lifetime of the process. By default, the WebSocket Proxy Plugin (@thymian/websocket-proxy) is part of the standard configuration, which opens a WebSocket server to accept connections from remote plugins.
+Those plugins can be written in any language and can fully control Thymian’s flow by emitting actions and events.
+
+In short, `thymian serve` turns Thymian into a controllable service. You can drive the complete end‑to‑end flow from a remote process—starting runs, reacting to reports, and terminating when done—without ever restarting the CLI, while still preserving the convenience of interactive local control.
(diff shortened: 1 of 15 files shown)


cli/docs/commands/serve.md
```diff
@@ -0,0 +1,12 @@
+---
+title: 'Serve'
+---
+
+The serve command runs Thymian as a long‑lived process so that you, your tools, and remote plugins can interact with it continuously. In this mode, Thymian boots the core, loads the configured plugins, and keeps running until you quit the session.
+This is the recommended mode when you want to drive Thymian from an external tool, an editor integration, a CI job, or any remote plugin connected via WebSockets.
```
BaggersIO, line 6: I'd suggest to add "during development" :)

cli/src/commands/serve.ts
```diff
@@ -0,0 +1,38 @@
...
+    process.stdin.resume();
+    process.stdin.on('data', async (data) => {
+      if (data.toString() === 'q') {
+        await close(0);
+      }
+    });
+
+    await this.thymian.ready();
+
+    this.log('Thymian is now in "serve" mode. Press "q" to exit.');
```
BaggersIO: 'q' should be a const. So you can reuse it in line 29 and 36.

BaggersIO (approved):
Thanks @matthyk! Looks good.

Beside the current changes, I checked which applications use `45678` as default port. This is the result:

```markdown
The most prominent software that uses **TCP port 45678** by default for a specific communication channel is the **Dremio** data lake engine.

---

## 🛠️ Software Using Port 45678 by Default

### 1. Dremio
* **Purpose:** Dremio is a high-performance, open-source platform that enables data analysts and data scientists to query data directly from S3, ADLS, and other data sources without needing to move it.
* **Port Use:** It uses port **45678** as the default port for its **Fabric Communication** channel. This is the port used for internal coordination and data transfer between the Dremio coordinator and executor nodes in a distributed cluster.

---

## 💡 Other Common/Observed Uses

Since port 45678 is in the **Ephemeral/Private** range, it is not officially registered by IANA and is often used by custom or less-common applications:

* **Custom Applications:** Many developers and administrators choose ports in the 40000+ range for **custom-built applications**, services, or **ad-hoc testing** to avoid conflicts with common services (like HTTP/80 or Tomcat/8080).
* **Testing Tools (Iperf):** Network testing tools like **Iperf** are frequently configured by users to listen on this port during network performance tests.
* **Configuration Examples:** Port **45678** often appears in documentation or configuration files as an example for custom host ports, such as for **PlatformIO** over-the-air (OTA) updates on IoT devices (e.g., ESP32/ESP8266).
* **Cybersecurity Context:** It has been observed in the wild being used by specific malware or backdoors (e.g., those associated with the RedPenguin threat group) to establish command and control (C2) communication.

---

## 📋 Port Range Context

Port 45678 falls in the range of **Registered** and **Dynamic/Private** ports (1024 to 65535). Applications that use ports in this higher range typically do so because:

1.  They do not require the **root/administrator** privileges needed to bind to **Well-Known Ports** (0-1023).
2.  They are less likely to conflict with standard, major network services.
```
Do you think we should change the default port to something else? However, we can go with this port now until it is widely used.

matthyk:
@BaggersIO, we could use the port 51234. Will change it along with your comments.

PR #65, opened 2025-12-01
Add license checker

BaggersIO:
Adds automated license compliance checking to the project to prevent incompatible licenses (AGPL, GPL) from being introduced through dependencies.

**Changes**
- License Checker Script (scripts/check-licenses.js): Validates all dependencies against a list of banned licenses, providing detailed reports on violations
- Configuration (.license-checker.json): Defines banned licenses (AGPL-3.0, GPL-2.0, GPL-3.0 variants)
- CI Integration (.github/workflows/ci.yaml): New license-check job runs on every CI build
- NPM Scripts (package.json): Added check:licenses command and prepare hook to run license checks on install

**Benefits**
- Proactive protection against viral license contamination (NPM postinstall & CI)
- Automated enforcement in CI/CD pipeline
- Early detection of license compliance issues
- Configurable exclusion list for special cases

Closes thymianofficial/thymian-internal#59

### DIFF
--- a/.license-checker.json
+++ b/.license-checker.json
@@ -1,0 +1,14 @@
+{
+  "bannedLicenses": [
+    "AGPL-3.0",
+    "AGPL-3.0-only",
+    "AGPL-3.0-or-later",
+    "GPL-3.0",
+    "GPL-3.0-only",
+    "GPL-3.0-or-later",
+    "GPL-2.0",
+    "GPL-2.0-only",
+    "GPL-2.0-or-later"
+  ],
+  "excludePackages": []
+}
--- a/package.json
+++ b/package.json
@@ -10,7 +10,9 @@
     "lint": "nx affected:lint",
     "prepare": "husky",
     "local-registry": "nx local-registry",
-    "local-publish": "npm run build && nx release publish --registry http://localhost:4873"
+    "local-publish": "npm run build && nx release publish --registry http://localhost:4873",
+    "check:licenses": "node scripts/check-licenses.js",
+    "postinstall": "npm run check:licenses"
   },
   "lint-staged": {
     "*.{js,ts,jsx,tsx}": [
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -10,7 +10,9 @@
     "lint": "nx affected:lint",
     "prepare": "husky",
     "local-registry": "nx local-registry",
-    "local-publish": "npm run build && nx release publish --registry http://localhost:4873"
+    "local-publish": "npm run build && nx release publish --registry http://localhost:4873",
+    "check:licenses": "node scripts/check-licenses.js",
+    "postinstall": "npm run check:licenses"
```
matthyk, line 15: Do we really need this post-install script? I would avoid post-install scripts if possible

package.json
```diff
@@ -41,8 +43,10 @@
     "eslint-plugin-import": "^2.31.0",
     "eslint-plugin-simple-import-sort": "^12.1.1",
     "eslint-plugin-unused-imports": "^4.1.4",
+    "glob": "^13.0.0",
```
matthyk: I would recommend [tinyglobby](https://www.npmjs.com/package/tinyglobby) instead

.license-checker.json
```diff
@@ -0,0 +1,14 @@
+{
+  "bannedLicenses": [
```
matthyk, line 2: Should we use a whitelist approach? I guess this would be more restrictive, and we could ensure that only reviewed licenses are used

package.json
BaggersIO: Let's discuss this with @Markus-Ende and @atennert :) I think it makes sense to left-shift possible problems into the development process. So we might avoid unnecessary work based on nonproprietary tools.

package.json
BaggersIO: Good point!

.license-checker.json
BaggersIO: That's a first approach. We'll discuss this with our lawyer about what makes the most sense. :)

package.json
BaggersIO: Replaced :)

matthyk (approved):

PR #111, opened 2026-01-23
docs(astro-docs): generate CLI references with oclif

atennert:


matthyk (commented):
I just thought about it again, and I think we should remove the extra content from the HTTP Linter/CLI.md after all. This should be placed in a guide or so.

BaggersIO:
@atennert do you plan to add the JSON schema generation in a seperate pull request? If not, I'd go with a review here and only focus on oclif :)

astro-docs/src/content/docs/references/Plugins/HTTP Linter/cli.md
```diff
@@ -5,266 +5,13 @@ description: 'Command-line tools for managing HTTP linting rules'
...
-| Command    | Purpose                         |
-| ---------- | ------------------------------- |
-| `generate` | Interactively create a new rule |
-| `list`     | Show all loaded rules           |
-| `search`   | Find rules by description       |
-| `overview` | Display rule statistics         |
+  <!-- commands -->
 
-## http-linter:generate
+# Common Workflows
```
matthyk: @BaggersIO and I just talked about it and both agree that the Common Workflow Section should be deleted.

packages/format-validator/project.json
```diff
@@ -5,5 +5,12 @@
...
   "tags": ["scope:plugin", "type:lib"],
   "// targets": "to see all targets run: nx show project @thymian/format-validator --web",
-  "targets": {}
+  "targets": {
+    "generate-docs": {
+      "command": "npx oclif readme --no-aliases --readme-path ../../astro-docs/src/content/docs/references/Plugins/Format\\ Validator/cli.md",
+      "options": {
+        "cwd": "packages/format-validator"
+      }
+    }
```
matthyk, line 14:
```suggestion
      },
      "dependsOn": ["build"]
    }
```

Without this dependency the following can happen:

1. I make a change in the description of the command
2. I run `nx build astro-docs`
3. I run `nx serve astro-docs`
4. I expect the changes to be present in the docs, but they are not there, because the `oclif readme` commands reads the commands from the `dist` folder.

This should be added to all `projects.json` which have a `generate-docs` target

matthyk (changes requested):

BaggersIO:
> @atennert do you plan to add the JSON schema generation in a seperate pull request? If not, I'd go with a review here and only focus on oclif :)

OK, got it: https://github.com/thymianofficial/thymian-internal/issues/79 :)

packages/format-validator/project.json
atennert: done

astro-docs/src/content/docs/references/Plugins/HTTP Linter/cli.md
atennert: done

packages/format-validator/project.json
matthyk: Was removed and solved by reading the commands from source

matthyk (approved):

PR #116, opened 2026-01-28
feat(core): add strict mode to emitAciton method

matthyk:
Closes [#thymianofficial/thymian-internal#75](https://github.com/thymianofficial/thymian-internal/issues/75)

## Summary

- Add strict mode to `emitAction` method in `ThymianEmitter`
- Throw error when emitting actions without registered handlers (strict mode enabled by default)
- Add proper TypeScript type guards to return `undefined` when strict mode is disabled

### Core Package (`packages/core`)

- **`thymian-emitter.ts`**:
  - Add `strict` boolean option to `EmitActionOptions` (defaults to `true`)
  - Add `EmitActionReturnType` conditional type that returns optional types when `strict: false`
  - Add validation check before emitting actions to ensure handlers are registered
  - Throw `NoHandlerRegisteredError` when no handler exists and strict mode is enabled
  - Return `undefined` when no handler exists and strict mode is disabled

- **`thymian-emitter.test.ts`**:
  - Add test case for strict mode error throwing behavior
  - Add test case for non-strict mode undefined return behavior

### DIFF
--- a/packages/core/src/emitter/thymian-emitter.ts
+++ b/packages/core/src/emitter/thymian-emitter.ts
@@ -18,6 +18,7 @@ import {
 
 import type { ThymianActionName, ThymianActions } from '../actions/index.js';
 import type { ThymianEventName } from '../events/index.js';
+import { ThymianFormat } from '../format/index.js';
 import type { Logger } from '../logger/logger.js';
 import { NoopLogger } from '../logger/noop.logger.js';
 import {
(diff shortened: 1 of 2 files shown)


packages/core/src/emitter/thymian-emitter.ts
```diff
@@ -326,23 +339,36 @@ export class ThymianEmitter {
...
+      strict: true,
       ...(options ?? {}),
     };
 
+    if (!this.#listeners.has(name) && !name.startsWith('core.')) {
+      if (opts.strict) {
+        this.emitError(
+          new ThymianBaseError(`No listener for action "${name}" registered.`, {
+            suggestions: [
+              'Did you forget to call register handler via "onAction" for it?',
```
Markus-Ende, line 362: do we already have docs for this? Could we link them here?

Markus-Ende (approved):
see comment :)

packages/core/src/emitter/thymian-emitter.ts
matthyk: No but we could think about an `Error Reference` section?

matthyk: See https://github.com/thymianofficial/thymian-internal/issues/100

PR #118, opened 2026-01-29
feat(core): replace circular references with json pointers

matthyk:
Closes https://github.com/thymianofficial/thymian-internal/issues/21

## Summary
Replace circular references in ThymianFormat with JSON pointers to enable safe serialization and deserialization.

## Changes
* Added circular reference handling: Implemented resolveCircularReferences() and resolveThymianPointers() functions in packages/core/src/format/circular-references.ts
* Integrated with ThymianFormat: Updated export() to replace circular references with JSON pointers ($thymian-ref) and import() to restore them
* Comprehensive test coverage: Added 303 lines of tests covering primitives, nested objects, arrays, self-references, and complex circular graphs

## Technical Details

The implementation uses JSON pointer syntax (#/path/to/object) to track object references during serialization. When a circular reference is detected, it's replaced with { "$thymian-ref": "#/path" }. During deserialization, these pointers are resolved back to the actual object references, fully restoring the circular structure.

### DIFF
--- a/packages/core/src/format/circular-references.ts
+++ b/packages/core/src/format/circular-references.ts
@@ -1,0 +1,84 @@
+const REF_KEY = '$thymian-ref';
+
+export function resolveCircularReferences<T>(obj: T): T {
+  const activeStack = new Map();
+  const REF_KEY = '$thymian-ref';
+
+  function recurse(current: any, path: string): any {
+    if (typeof current !== 'object' || current === null) {
+      return current;
+    }
+
+    if (activeStack.has(current)) {
+      return { [REF_KEY]: activeStack.get(current) };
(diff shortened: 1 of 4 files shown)


packages/core/src/format/circular-references.ts
```diff
@@ -0,0 +1,84 @@
+const REF_KEY = '$thymian-ref';
+
+export function resolveCircularReferences<T>(obj: T): T {
+  const activeStack = new Map();
+  const REF_KEY = '$thymian-ref';
```
atennert: I think one `REF_KEY` is enough

atennert (approved):

PR #123, opened 2026-02-04
docs(astro-docs): generate plugin option docs from schema JSON

atennert:


BaggersIO:
@atennert I added some small changes. Plugin options always have order -100 to be at the top of the plugin docs. CLI comes with -50 to be on the second place. Everything else comes afterward.

BaggersIO (approved):

PR #124, opened 2026-02-04
chore(request-dispatcher): http request error log

xquadradpi:
- add propper log, if server is not reachable on http request
grafik


packages/test-utils/thymian.config.yaml
atennert, line 1: Don't check in the thymian.config.yaml :)

packages/request-dispatcher/src/index.ts
```diff
@@ -25,6 +25,14 @@ declare module '@thymian/core' {
   }
 }
 
+export interface UndiciError extends Error {
```
atennert: I advice against the name Undici here. Might be a fun reference, but might also cause some headaches in terms of maintainability for those who don't get it.

atennert (approved):

Markus-Ende:
closes https://github.com/thymianofficial/thymian-internal/issues/32

PR #134, opened 2026-02-13
docs(repo): consistent error naming and referencing

matthyk:
This PR:

* introduces error reference pages for each currently thrown error 
* each pages described the error and how it can be resolved
* Each error have no a distinct name and references its reference page

Closes https://github.com/thymianofficial/thymian-internal/issues/100


atennert (approved):
Generally it might be good to use some subtitles like **Cause** and **How to solve**, as those are recurring parts in the descriptions.

matthyk:
> Generally it might be good to use some subtitles like **Cause** and **How to solve**, as those are recurring parts in the descriptions.

I like the idea, let's do it

PR #147, opened 2026-02-17
feat(repo): add e2e test

atennert:


matthyk (commented):
Cool! Maybe we should split the test of the different commands into separate files? This would scale better if we add more tests for other commands. Or is there a reason why everything is in one file?

.github/workflows/ci.yaml
```diff
@@ -71,4 +71,4 @@ jobs:
       - name: Build
         run: npx nx affected:build
       - name: Test
-        run: npx nx affected:test
+        run: npx nx affected:test --exclude=e2e-tests
```
Markus-Ende: rename the test target of e2e tests (e.g. into e2e) instead, then you dont need to exclude it here. Otherwise we need to do that locally also

e2e-tests/README.md
```diff
@@ -0,0 +1,7 @@
+# e2e-tests
+
+This library was generated with [Nx](https://nx.dev).
```
Markus-Ende: we need some documentation ;)

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,57 @@
...
+              echo "Verdaccio is ready!"
+              exit 0
+            fi
+            echo "Attempt $i/30: Verdaccio not ready yet..."
+            sleep 2
+          done
+          echo "Verdaccio failed to start within 60 seconds"
+          exit 1
+      - name: Publish to local registry
+        run: npm run local-publish -- --version 0.0.1 --no-dry-run
```
Markus-Ende: 0.0.1 is dangerous, because it really exists, so if publishing would fail silently, we would test the remote 0.0.1. I would propose a non-exising version, e.g. 1337.0.0

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,57 @@
...
+        with:
+          node-version: 22
+      - name: install
+        run: npm ci
+      - name: Build packages
+        run: npx nx build thymian
+      - name: Start Verdaccio
+        run: |
+          npm run local-registry &
+      - name: Wait for Verdaccio
```
Markus-Ende: The verdaccio handling should be integrated into the test global setup, so that it easily works in every environment.

You can basically copy functionality from nx: https://github.com/nrwl/nx/blob/master/e2e/utils/global-setup.ts

e2e-tests/project.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "e2e-tests",
+  "$schema": "../node_modules/nx/schemas/project-schema.json",
+  "sourceRoot": "e2e-tests/src",
+  "projectType": "library",
+  "tags": ["type:e2e"],
+  "// targets": "to see all targets run: nx show project e2e-tests --web",
+  "targets": {}
```
Markus-Ende: Please change default test target to e2e, this breaks running unit tests locally, see https://github.com/thymianofficial/thymian/pull/147/changes#r2857317775

Markus-Ende (changes requested):
we dont need to cover all before merging this, but at least please change the target name please

e2e-tests/README.md
atennert: :see_no_evil:

Markus-Ende (approved):

PR #148, opened 2026-02-17
Mk/feat/feedback command

matthyk:
Implements:
* error caching (e.g. for macos in the `/Library/Caches/thymian` folder)
* the `feedback` command
* suggests the feedback command for every error and with a probability of 20% for each command

How can you test the new functionality?

1. checkout the the branch
2. build thymian
3. from root dir run `./packages/thymian/bin/run.js init --yes` 
5. then `./packages/thymian/bin/run.js run`
6. You should then see an error and the tip to run the `feedback command`. Do this via `./packages/thymian/bin/run.js feedback`

Closes https://github.com/thymianofficial/thymian-internal/issues/111

### DIFF
--- a/packages/cli-common/src/feedback.ts
+++ b/packages/cli-common/src/feedback.ts
@@ -18,7 +18,7 @@ export class Feedback {
       return;
     }
 
-    if (Math.random() < 0.2) {
+    if (this.shouldPrintFeedback()) {
       this.wasRun = true;
       await this.onRun();
     }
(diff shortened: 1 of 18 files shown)


packages/cli-common/src/feedback.ts
```diff
@@ -0,0 +1,43 @@
...
+  ) {
+    this.wasRun = initial;
+  }
+
+  async run(): Promise<void> {
+    if (this.wasRun) {
+      return;
+    }
+
+    if (Math.random() < 0.2) {
```
BaggersIO: What's that random code for?

BaggersIO (dismissed):

packages/cli-common/src/feedback.ts
matthyk: I wanted the feedback message to appear only occasionally. This essentially means that the message appears in 20 percent of cases. This should be commented more clearly ;)

BaggersIO (approved):

BaggersIO:
@matthyk feel free to merge!

PR #149, opened 2026-02-17
Mk/feat/rfc9110 range requests

matthyk:
Closes https://github.com/thymianofficial/thymian-internal/issues/92

I targeted `rfc9110-content-negotiation` because its the base branch of the current feature branch and we get a cleaner diff.


packages/rfc-9110-rules/src/rules/range-requests/content-range/utils.ts
BaggersIO, line 1: I'd propose to rename this to `utils.ts` in the folder context of `content-range`, since we use this pattern at other parts of the repo.

packages/rfc-9110-rules/src/rules/range-requests/range/client-should-list-multiple-ranges-in-ascending-order.rule.ts
BaggersIO, line 1: Nice implementation :)

packages/rfc-9110-rules/src/rules/range-requests/range/utils.ts
BaggersIO, line 1: Imho, this should be a utils.ts, too :)

BaggersIO (changes requested):
The rule implementation looks very good so far. However, I'd like to re-organize the util functions as stated in the comments.

packages/rfc-9110-rules/src/rules/range-requests/content-range/utils.ts
matthyk: done

packages/rfc-9110-rules/src/rules/range-requests/range/utils.ts
matthyk: done

BaggersIO (approved):

PR #160, opened 2026-02-24
chore(repo): add opencode config

matthyk:


opencode.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "https://opencode.ai/config.json",
+  "skills": {
+    "paths": [".github/skills"]
```
Markus-Ende, line 4: can you also point agents to .github/agents and prompts to .github/prompts?

Markus-Ende (approved):

PR #173, opened 2026-03-11
test(repo): try to get websocket test running

atennert:


matthyk (approved):
Maybe we should split the tests into separate files?  This would scale better if we add new tests

PR #206, opened 2026-04-08
feat(rules-rfc-9110): move http identifiers rules into correct rule set

matthyk:
Commit 1e09c3a442c86347997806a69e967308334f7c25 add http rule for the http identifiers chapter but they were merged into the wrong directory


packages/rules-rfc-9110/src/rules/identifiers/userinfo/sender-must-not-generate-userinfo-in-uri.rule.ts
```diff
@@ -1,5 +1,5 @@
 import { or, protocol } from '@thymian/core';
-import { httpRule } from '@thymian/http-linter';
+import { httpRule } from '@thymian/core';
```
Markus-Ende: Should probably be merged into line 1

Markus-Ende (approved):

PR #248, opened 2026-04-23
fix(rules-rfc-9110): make rule informational

matthyk:
We need to make this rule `informational` for now. In the future, we should add the functionality to send invalid HTTP requests to test implementations and see how they respond to invalid requests.


Markus-Ende (approved):
please add the reason as code comment

PR #269, opened 2026-05-26
fix(rules-rfc-9110): reclassify content-negotiation rule execution contexts

BaggersIO:
## Summary

Reviewed and updated execution context assignments for all 11 content-negotiation rules in `@thymian/rules-rfc-9110` to align with the intended lint/test/analyze execution model.

Closes part of thymianofficial/thymian-internal#327

## Rationale

Content-negotiation rules deal with headers like `Accept`, `Accept-Encoding`, `Vary`, and quality values. Whether a rule belongs in `test` depends on whether it validates **response-side** (server) behavior or **request-side** (client/sender) behavior:

- **Response-side rules** belong in `test` because Thymian sends a request and can observe the server's response. The server's behavior is what's being validated.
- **Request-side rules** do not belong in `test` because Thymian generates requests from the OpenAPI spec — it is the sender. Validating Thymian's own generated requests is meaningless. These rules belong in `analyze`, where recorded traffic contains requests made by real clients (developers, third-party consumers) who had full control over what they sent.

Three rules were assigned to executable contexts (`static` or `analytics`) despite having no `.rule()` function and describing behaviors that cannot be observed from outside the system. Keeping these as executable types is misleading — they will never produce findings, and their presence in rule listings for those modes sets false expectations.

The `origin-server-should-generate-vary-header` rule is security-relevant: a missing `Vary` header on cacheable responses can enable cache poisoning. Expanding it to all three contexts ensures it will be flagged in lint (spec review), test (live validation of server responses), and analyze (production traffic audit) once the rule function is implemented.

### Rules expanded to include `test` context (3 rules)

These rules validate **server/response-side behavior** and have existing `.rule()` functions. In `test` mode, Thymian sends a request and the server's response can be checked — adding `test` is correct because the server's behavior is what's under validation.

- `server-must-not-include-accept-encoding-for-non-content-coding-415-errors` — `static, analytics` → `static, analytics, test`
  **Why**: Validates that a server returning 415 for non-content-coding reasons does not include `Accept-Encoding` in the response. This is observable in live server responses.
- `cache-must-not-use-response-without-matching-vary-headers` — `analytics` → `analytics, test`
  **Why**: Validates caching correctness by checking Vary-related headers in responses. Observable when testing against a live server with caching infrastructure.
- `proxy-must-not-generate-vary-wildcard` — `analytics` → `analytics, test`
  **Why**: Validates that `Vary: *` is not present in responses. Observable in live responses from servers behind proxies.

### Rules reclassified to `informational` (3 rules)

These rules had no `.rule()` function and describe requirements that depend on internal implementation details not observable through HTTP traffic or API descriptions. Automatic validation is not feasible.

- `origin-server-should-send-response-without-content-coding` — `static` → `informational`
  **Why**: Determining whether "none of the available representations have an acceptable content coding" requires knowledge of the server's full representation set, which is not available from a spec or traffic alone.
- `user-agent-must-not-send-accept-language-without-user-control` — `analytics` → `informational`
  **Why**: Whether a user agent exposes language preference controls to the user is an implementation detail that cannot be inferred from the presence or absence of `Accept-Language` in traffic.
- `recipient-should-process-q-parameter-as-weight` — `static` → `informational`
  **Why**: How a recipient internally interprets the `q` parameter is unobservable — there is no external signal that distinguishes correct from incorrect processing.

### Rules expanded with TODO for implementation (1 rule)

- `origin-server-should-generate-vary-header` — `static` → `static, test, analytics`
  **Why**: Vary header presence on cacheable responses is a **response-side** check, valid in all three contexts. In lint, the API description can declare Vary on negotiated endpoints. In test, live server responses reveal whether Vary is actually sent. In analyze, recorded traffic exposes gaps introduced by infrastructure (proxies, CDNs) that may strip or fail to add Vary. This is security-relevant because missing Vary enables cache poisoning. A TODO stub is added; the rule function implementation will follow.

### No change (4 rules)

- `user-agent-may-associate-quality-value-with-charset` — stays `static, analytics`
  **Why**: Validates request-side behavior (`Accept-Charset` header). Does not belong in `test` because Thymian generates the request — validating its own output is not meaningful. Correctly present in `analytics` where real client requests can be checked.
- `codings-value-may-be-given-quality-value` — stays `analytics, static`
  **Why**: Validates request-side behavior (`Accept-Encoding` header). Same reasoning — request originates from Thymian in test mode, not from a real client.
- `sender-should-send-q-parameter-last` — stays `analytics, static`
  **Why**: Validates request-side behavior (`Accept` header q-parameter ordering). The analytics override inspects actual header values from real client traffic, which is where this check is meaningful.
- `user-agent-may-send-preference-headers-for-proactive-negotiation` — stays `informational`
  **Why**: General allowance ("a user agent MAY send preference headers") with no specific condition to validate.

## Verification
- `nx run rules-rfc-9110:test` — 64/64 passing
- `nx run rules-rfc-9110:lint` — 0 errors

### DIFF
--- a/.opencode/package-lock.json
+++ b/.opencode/package-lock.json
@@ -1,0 +1,115 @@
+{
+  "name": ".opencode",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "dependencies": {
+        "@opencode-ai/plugin": "1.3.17"
+      }
+    },
+    "node_modules/@opencode-ai/plugin": {
+      "version": "1.3.17",
+      "resolved": "https://registry.npmjs.org/@opencode-ai/plugin/-/plugin-1.3.17.tgz",
+      "integrity": "sha512-N5lckFtYvEu2R8K1um//MIOTHsJHniF2kHoPIWPCrxKG5Jpismt1ISGzIiU3aKI2ht/9VgcqKPC5oZFLdmpxPw==",
+      "license": "MIT",
+      "dependencies": {
+        "@opencode-ai/sdk": "1.3.17",
+        "zod": "4.1.8"
+      },
+      "peerDependencies": {
+        "@opentui/core": ">=0.1.96",
+        "@opentui/solid": ">=0.1.96"
+      },
+      "peerDependenciesMeta": {
+        "@opentui/core": {
+          "optional": true
+        },
+        "@opentui/solid": {
+          "optional": true
+        }
+      }
+    },
+    "node_modules/@opencode-ai/sdk": {
+      "version": "1.3.17",
+      "resolved": "https://registry.npmjs.org/@opencode-ai/sdk/-/sdk-1.3.17.tgz",
+      "integrity": "sha512-2+MGgu7wynqTBwxezR01VAGhILXlpcHDY/pF7SWB87WOgLt3kD55HjKHNj6PWxyY8n575AZolR95VUC3gtwfmA==",
+      "license": "MIT",
+      "dependencies": {
+        "cross-spawn": "7.0.6"
+      }
+    },
+    "node_modules/cross-spawn": {
+      "version": "7.0.6",
+      "resolved": "https://registry.npmjs.org/cross-spawn/-/cross-spawn-7.0.6.tgz",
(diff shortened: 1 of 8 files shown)


BaggersIO:
Closing as part of restarting the RFC 9110 execution-context work from scratch. #327 has been rescoped to rule-level hardening (no framework changes); each category will be re-derived against the new four-outcome triage. This branch is left intact as reference and is not being deleted.

PR #276, opened 2026-06-03
fix(core): make hash generation independant from source (location)

matthyk:
# Summary

 This PR makes Thymian format hashing deterministic across machines by excluding non-semantic attribution fields from hash calculation.

 The format hash now ignores:

 - label
 - sourceLocation
 - sourceName

 These fields can vary depending on where or how the format was generated, for example due to absolute file paths or machine-specific source names. They should not affect
 whether the actual Thymian graph changed.

 ## Why this is necessary

 Previously, the hash of a Thymian format could change depending on the physical machine where it was calculated. This happened because source attribution metadata was included
 in the node/edge hashing inputs.

 That meant two semantically identical formats could produce different hashes if, for example:

 - the OpenAPI source file lived under different absolute paths
 - source names differed between environments
 - generated labels differed while request/response/security graph semantics stayed the same

 This made the hash unreliable as a signal for actual format changes.

 The hash should only change when semantic graph content changes, such as request, response, security, sample, or edge data.

 ## Changes

 - Introduced semantic hashing for ThymianFormat
 - Excluded label, sourceLocation, and sourceName from hash inputs
 - Applied semantic hashing to generated node and edge IDs
 - Added recursive stripping of ignored attribution fields
 - Preserved topology in format hashing by including semantic source/target node hashes for edges
 - Added regression tests covering:
     - different machine-specific source paths
     - stable IDs
     - exported hash values
     - imported formats
     - nested metadata
     - security schemes and HTTP links

### DIFF
--- a/packages/core/src/format/thymian-format.ts
+++ b/packages/core/src/format/thymian-format.ts
@@ -131,6 +131,11 @@ export class ThymianFormat {
   readonly graph: ThymianGraph;
 
   static readonly ignoredNodeProperties: string[] = ['x', 'y'];
+  private static readonly ignoredHashProperties = [
+    'label',
+    'sourceLocation',
+    'sourceName',
+  ] as const;
 
   constructor(graph: ThymianGraph = new MultiDirectedGraph()) {
     this.graph = graph;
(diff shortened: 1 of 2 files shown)


atennert (approved):
LGTM

PR #284, opened 2026-06-19
Story/333 map plugin execution data

atennert:


matthyk (approved):
let's fix the e2e tests when we are done with this epic

PR #298, opened 2026-06-29
feat(rules-rfc-9110): align content-negotiation rule execution contexts (#327)

BaggersIO:
## Summary

Resolves the **`content-negotiation/` category (11 rules)** of thymianofficial/thymian-internal#327. Every rule was reviewed exactly once and resolved into one of the four triage outcomes, using **only the existing rule framework** — no changes to `packages/core`, the linter/tester/analyzer plugins, the builder, runner, or loader. Changes are confined to `packages/rules-rfc-9110/src/rules/content-negotiation/**`.

Tutorials under `astro-docs/src/content/docs/tutorials/` were used as the source of truth for context semantics:
- **lint** — design-time validation of the description, no HTTP.
- **test** — live responses from a running server; **response/server-side rules only** (Thymian is the request sender).
- **analyze** — recorded traffic from real clients; **both request- and response-side** rules apply.

## Coverage matrix

| # | Rule | Before | After | Outcome | Notes |
|---|------|--------|-------|---------|-------|
| 1 | accept-charset/`user-agent-may-associate-quality-value-with-charset` | `static, analytics` (presence flag) | `informational` | 2 | Permissive MAY, no violation condition; old presence-flag was noise. Request-side. |
| 2 | accept-encoding/`codings-value-may-be-given-quality-value` | `analytics, static` (presence flag) | `informational` | 2 | Permissive MAY, no violation condition. Request-side. |
| 3 | accept-encoding/`origin-server-should-send-response-without-content-coding` | `static` (no fn) | `test, analytics` **+ impl** | 1 | Now checks the response `Content-Encoding` is acceptable per the request `Accept-Encoding` (q=0, identity, `*`). Response/server-side ⇒ valid in test. Not observable from a static description. |
| 4 | accept-encoding/`server-must-not-include-accept-encoding-for-non-content-coding-415-errors` | `static, analytics` | `static, test, analytics` | 1/4 | Response-side server behavior ⇒ added `test`. Shared `.rule()` over the common projection works in all three. |
| 5 | accept-language/`user-agent-must-not-send-accept-language-without-user-control` | `analytics` (no fn) | `informational` | 2/4 | Trigger ("no user control") is internal UA state, unobservable from any context. Was a declared-but-empty analytics rule. |
| 6 | accept/`recipient-should-process-q-parameter-as-weight` | `static` (no fn) | `informational` | 2/4 | Recipient parsing logic; not present in any message/description. |
| 7 | accept/`sender-should-send-q-parameter-last` | `analytics, static` (+ analytics override) | `analytics` | 4 | **Request-side ⇒ dropped `static`, analyze-only.** Consolidated parser into `.rule()`; **fixed a false positive** that flagged values carrying no `q` parameter. |
| 8 | vary/`cache-must-not-use-response-without-matching-vary-headers` | `analytics` (broken heuristic) | `analytics` (no fn) | 3 | Infra-deferred. Detecting it needs cache-hit linkage + selecting-header correlation not available from a single transaction/HAR. Removed a heuristic that flagged ~all traffic with a common header as an `error`. Scoped to `cache`. |
| 9 | vary/`origin-server-should-generate-vary-header` ⭐ | `static` (no fn) | `static, test, analytics` **+ impl** | 1 | **Golden reference.** Flags cacheable (GET/HEAD, 2xx) responses to negotiated requests (`Accept*`) that omit `Vary` — a cache-poisoning hazard. Shared `.rule()` runs in all three contexts. |
| 10 | vary/`proxy-must-not-generate-vary-wildcard` | `analytics` + impl | unchanged | 1 | Already correct: proxy-scoped, parses `Vary: *`. Reviewed, no change. |
| 11 | proactive-negotiation/`user-agent-may-send-preference-headers-for-proactive-negotiation` | `informational` | `informational` | 2 | Already correct; added written justification. |

## Per-PR checklist

- [x] Every rule in the category reviewed exactly once (11/11)
- [x] No declared active context left without a rule function, except the one documented infrastructure rule (#8, outcome 3)
- [x] No request-side rule is assigned to `test` (#1, #2 → informational; #7 → analyze-only)
- [x] Every `informational` and infrastructure-deferred rule carries a written reason (in-file comments)
- [x] Coverage matrix included (above)
- [x] Security-relevant cache rules explicitly addressed: `Vary` generation (#9, implemented across 3 contexts), `Vary: *` from proxies (#10), cache reuse (#8, documented deferral)
- [x] PR explains why each reclassification makes sense and the consequence of the previous classification

## Security focus

The `Vary` cluster is the cache-security core. #9 now detects negotiated-but-un-`Vary`'d cacheable responses in lint, test, **and** recorded traffic (the cache-poisoning weak spot called out in the issue). #10 keeps catching proxy-generated `Vary: *`. #8 is honestly deferred to infrastructure rather than shipping a heuristic that errored on nearly all traffic.

## Validation

`nx build rules-rfc-9110` ✓ · `nx lint rules-rfc-9110` ✓ (0 errors) · prettier ✓

## Out of scope / follow-up

No framework changes were needed for this category. The role-filtering / `appliesTo` interaction in `analyze` and the lack of automated invariant enforcement remain framework concerns tracked by the issue's follow-up comment.


packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-charset/user-agent-may-associate-quality-value-with-charset.rule.ts
```diff
@@ -1,17 +1,13 @@
-import { requestHeader } from '@thymian/core';
 import { httpRule } from '@thymian/core';
 
 export default httpRule(
   'rfc9110/user-agent-may-associate-quality-value-with-charset',
 )
   .severity('hint')
-  .type('static', 'analytics')
+  .type('informational')
```
BaggersIO: Informational: this is a permissive "MAY" granting the user agent the option to attach a quality value to an Accept-Charset charset. There is no condition under which a transaction is non-conformant — associating (or not associating) a weight is always allowed — so no execution context can produce a meaningful pass/fail. The previous `static`/`analytics` classification merely flagged the presence of the request header, which is noise rather than a conformance check. (Accept-Charset is also deprecated by RFC 9110 Section 12.5.2.)

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/codings-value-may-be-given-quality-value.rule.ts
```diff
@@ -1,14 +1,11 @@
-import { requestHeader } from '@thymian/core';
 import { httpRule } from '@thymian/core';
 
 export default httpRule('rfc9110/codings-value-may-be-given-quality-value')
   .severity('hint')
-  .type('analytics', 'static')
+  .type('informational')
```
BaggersIO: Informational: this is a permissive "MAY" granting the sender the option to attach a quality value to an Accept-Encoding codings value. There is no condition under which a transaction is non-conformant, so no execution context can produce a meaningful pass/fail. The previous `static`/`analytics` classification merely flagged the presence of the request header, which is noise rather than a conformance check. (Well-formedness of the weight syntax is covered by separate field-syntax rules, not this one.)

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/server-must-not-include-accept-encoding-for-non-content-coding-415-errors.rule.ts
```diff
@@ -11,15 +11,15 @@ export default httpRule(
   'rfc9110/server-must-not-include-accept-encoding-for-non-content-coding-415-errors',
 )
   .severity('error')
-  .type('static', 'analytics')
+  .type('static', 'test', 'analytics')
```
BaggersIO, line 14: appliesTo justification: includes 'origin server' so the analyze path matches recorded traffic: HAR import tags responses with the 'origin server' role, and this rule validates a response the server emits, so role-filtering on 'server' alone would never select default-imported HAR transactions.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-language/user-agent-must-not-send-accept-language-without-user-control.rule.ts
```diff
@@ -3,8 +3,8 @@ import { httpRule } from '@thymian/core';
 export default httpRule(
   'rfc9110/user-agent-must-not-send-accept-language-without-user-control',
 )
-  .severity('error')
-  .type('analytics')
+  .severity('hint')
+  .type('informational')
```
BaggersIO, line 13: Informational: the precondition that triggers this MUST NOT — that the user agent "does not provide user control" over linguistic preference — is internal user-agent state. It cannot be derived from the API description, a live response, or recorded traffic: observing an Accept-Language header says nothing about whether the agent exposes user control over it. Genuinely unobservable, so it is informational rather than a declared-but-empty `analytics` rule.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept/recipient-should-process-q-parameter-as-weight.rule.ts
```diff
@@ -3,10 +3,11 @@ import { httpRule } from '@thymian/core';
 export default httpRule(
   'rfc9110/recipient-should-process-q-parameter-as-weight',
 )
-  .severity('warn')
-  .type('static')
+  .severity('hint')
+  .type('informational')
```
BaggersIO, line 12: Informational: this constrains the recipient's internal parsing behavior — how it interprets a "q" parameter when selecting a representation. That decision logic is not present in any HTTP message, the API description, or recorded traffic, so it is genuinely unobservable. The previous declared-but-empty `static` classification could not have validated it. (The related observable, sender-side ordering, is covered by rfc9110/sender-should-send-q-parameter-last.)

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept/sender-should-send-q-parameter-last.rule.ts
```diff
@@ -5,15 +5,23 @@ import {
...
+ * scope a media range (not as Accept extension parameters). If one of these
+ * appears after the "q" weight in an Accept value, the sender most likely
+ * mis-ordered it and it will be silently treated as an accept-ext parameter.
+ */
+const mediaRangeParameterNames = new Set(['charset', 'level', 'boundary']);
+
 export default httpRule('rfc9110/sender-should-send-q-parameter-last')
   .severity('warn')
-  .type('analytics', 'static')
+  .type('analytics')
```
BaggersIO, line 20: Request-side rule: it constrains how the *sender* orders parameters in the Accept request header. During `test` the request is generated by Thymian, so the sender's ordering is not under user control and the rule cannot meaningfully apply; during `lint` there is no real sender to observe. It is therefore only validated against recorded traffic (`analyze`), where the request originates from a real client.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/vary/cache-must-not-use-response-without-matching-vary-headers.rule.ts
```diff
@@ -25,9 +13,4 @@ export default httpRule(
     'Cache MUST NOT use this response to satisfy a later request unless the later request has the same values for the listed Vary header fields.',
   )
   .appliesTo('cache')
```
BaggersIO, line 23: Infrastructure-deferred (outcome 3): this MUST NOT governs a cache's reuse decision — it forbids serving a stored response to a later request whose values for the Vary-listed header fields differ from the original request. Detecting a violation requires correlating two requests AND knowing that the second was answered from cache (a cache hit) with the first request's stored representation. That cache-hit linkage is not available from a single transaction and is not reliably reconstructable from a HAR; it depends on the deployed cache. It *could* be validated from traffic recorded at a cache that annotates hits and the selecting header values, so the rule stays typed `analytics` and scoped to the `cache` role rather than being downgraded to informational. The previous implementation flagged any transaction whose request merely carried a common selecting header (Accept-Encoding, Cookie, Authorization, …) as an error. That is not the rule: it does not observe cache reuse or mismatched Vary values, and it fires on virtually all normal traffic. It has been removed in favor of an honest, documented deferral (no rule function) per the outcome-3 contract. (Anchored at the .appliesTo line because the .type line is not within this PR's diff hunk for this file.)

BaggersIO (commented):
Recording the execution-context triage rationale here (moved out of inline code comments).

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-charset/user-agent-may-associate-quality-value-with-charset.rule.ts
matthyk: I would say that the purpose of the rules marked with `hint` is to inform users about HTTP features they might not think of. But since the `Accept-Charset` header is actual deprecated, I think we have three options:

1. stick with `static` and `analytics` context with the previous logic
2. stick with `static` and `analytics` but print that this header is deprecated
3. keep this rule totally `informational`

Personally, I'd go with one of the first two options.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/codings-value-may-be-given-quality-value.rule.ts
matthyk: The same applies to this rule. The purpose of `hints` is to inform about mechanisms provided by the HTTP protocol not to inform about non-conformance.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/codings-value-may-be-given-quality-value.rule.ts
```diff
@@ -1,14 +1,11 @@
...
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#name-accept-encoding')
   .description(
     'Each codings value MAY be given an associated quality value (weight) representing the preference for that encoding, as defined in Section 12.4.2.',
   )
-  .rule((ctx) =>
-    ctx.validateCommonHttpTransactions(requestHeader('accept-encoding')),
-  )
+  .appliesTo('client', 'user-agent')
```
matthyk:
```suggestion
```

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/origin-server-should-send-response-without-content-coding.rule.ts
```diff
@@ -13,4 +93,46 @@ export default httpRule(
...
+              message: `The response was encoded with content coding(s) ${createList(
+                unacceptable,
+              )} that the request's Accept-Encoding (${acceptEncoding.trim()}) does not list as acceptable. When no acceptable coding is available, the origin server SHOULD send the response without any content coding (unless identity is indicated as unacceptable).`,
+            },
+            findings: [],
+          },
+        ];
+      },
+    ),
+  )
```
matthyk, line 137: nice

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/server-must-not-include-accept-encoding-for-non-content-coding-415-errors.rule.ts
```diff
@@ -11,15 +11,15 @@ export default httpRule(
...
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#name-accept-encoding')
   .description(
     'In order to avoid confusion with issues related to media types, servers that fail a request with a 415 status for reasons unrelated to content codings MUST NOT include the Accept-Encoding header field.',
   )
   .summary(
     'Servers that fail a request with a 415 status for reasons unrelated to content codings MUST NOT include the Accept-Encoding header field.',
   )
-  .appliesTo('server')
+  .appliesTo('server', 'origin server')
```
matthyk: Actual, the `server` tag is a superset of `origin server`. Server includes all component accepting incoming requests. But actually, I am not 100% sure if this implemented properly. We should check this and create a dedicated ticket, if not.
```suggestion
  .appliesTo('server')
```

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-charset/user-agent-may-associate-quality-value-with-charset.rule.ts
matthyk: Decision: Keep the current implementation and use a combination of the first two options. Provide a `hint` that a `user-agent` can provide a quality value for a charset AND provide the information that this header is actual deprecated, and that is should be used `UTF-8` as charset. However, adapt the implementation so that the actual header value is checked in the `analytics` context, and the `hint` is only outputted if no quality value is provided.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/codings-value-may-be-given-quality-value.rule.ts
matthyk: Decision: Keep this rule as it is but override the rule for the `analytics` context to check if the actual header values already contain a quality value or not.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/server-must-not-include-accept-encoding-for-non-content-coding-415-errors.rule.ts
matthyk: That's a tough one. I am actually thinking about making the rule informational. It's hard to test if a 415 response was triggered by a problem with the corresponding media type or by problems with the content coding. If the severity would be `hint` it would not be a problem I guess, but the severity is `error`. We could change the severity accordingly, but this would not reflect the RFC properly. What do you think @BaggersIO ?

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/server-must-not-include-accept-encoding-for-non-content-coding-415-errors.rule.ts
matthyk: Decision: remove the `origin server` tag. I will create a dedicated issue for this more general problem.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-language/user-agent-must-not-send-accept-language-without-user-control.rule.ts
```diff
@@ -3,8 +3,8 @@ import { httpRule } from '@thymian/core';
 export default httpRule(
   'rfc9110/user-agent-must-not-send-accept-language-without-user-control',
 )
-  .severity('error')
-  .type('analytics')
+  .severity('hint')
```
matthyk:
```suggestion
  .severity('error')
```

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept/recipient-should-process-q-parameter-as-weight.rule.ts
```diff
@@ -3,10 +3,11 @@ import { httpRule } from '@thymian/core';
 export default httpRule(
   'rfc9110/recipient-should-process-q-parameter-as-weight',
 )
-  .severity('warn')
-  .type('static')
+  .severity('hint')
```
matthyk:
```suggestion
  .severity('warn')
```

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept/sender-should-send-q-parameter-last.rule.ts
```diff
@@ -5,15 +5,23 @@ import {
...
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#name-accept')
   .summary(
     'Senders using weights SHOULD send "q" last (after all media-range parameters).',
   )
-  .rule((ctx) => ctx.validateCommonHttpTransactions(requestHeader('accept')))
-  .overrideAnalyticsRule((ctx) =>
+  .appliesTo('client', 'user-agent')
```
matthyk:
```suggestion
```

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/vary/cache-must-not-use-response-without-matching-vary-headers.rule.ts
matthyk: I agree, but then the `.type('analytics')` must be removed.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/vary/origin-server-should-generate-vary-header.rule.ts
```diff
@@ -11,4 +47,36 @@ export default httpRule('rfc9110/origin-server-should-generate-vary-header')
     'An origin server SHOULD generate a Vary header field on a cacheable response when it wishes that response to be selectively reused for subsequent requests.',
   )
   .appliesTo('origin server')
+  .rule((ctx) =>
```
matthyk: This implementation could be improved. While we can test this rule for all three contexts we should split the implementation for the `static` context and the `test` + `analytics` context. For the `static` context we can only check if there is a `Vary` header if at least one of the `negotiationRequestHeaders` are set. For the other two contexts we can check the actual value(s) of the `Vary` header.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/vary/origin-server-should-generate-vary-header.rule.ts
```diff
@@ -11,4 +47,36 @@ export default httpRule('rfc9110/origin-server-should-generate-vary-header')
...
+    ctx.validateCommonHttpTransactions(
+      and(
+        // Only safe, cacheable methods can yield a stored, reusable response.
+        or(method('GET'), method('HEAD')),
+        // A 2xx response is the cacheable, negotiable case this rule targets.
+        statusCodeRange(200, 299),
```
matthyk: This can be extended. Responses with the following status codes are at least heuristically cacheable:

```ts
[
  200,
  203,
  204,
  206,
  300,
  301,
  308,
  404,
  405,
  410,
  414,
  501,
];
```

matthyk (changes requested):

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/server-must-not-include-accept-encoding-for-non-content-coding-415-errors.rule.ts
matthyk: Decision: Let's use all three types (`static`, `test` and `analytics`) and keep the `error` severity. We add this rule to our future recommend thymian/rules config and will set this rule's severity to `hint` to reduce its noice since it will produce some false-positives.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-charset/user-agent-may-associate-quality-value-with-charset.rule.ts
BaggersIO: Done in 2230f7a8. Reverted to `.type('static', 'analytics')` and kept `hint`. The description/summary now call out that `Accept-Charset` is deprecated (Section 12.5.2) and that senders should rely on UTF-8. The `static` context emits the hint whenever the header is present (names only), and the `analytics` override reads the actual value and only emits the hint when no charset already carries a `q` weight.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/codings-value-may-be-given-quality-value.rule.ts
BaggersIO: Done in 2230f7a8. Reverted to `.type('static', 'analytics')` and added an `overrideAnalyticsRule` that inspects the real `Accept-Encoding` value and only surfaces the hint when none of the codings already carry a `q` weight. The `static` context keeps the presence-based hint.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-language/user-agent-must-not-send-accept-language-without-user-control.rule.ts
BaggersIO: Done in 2230f7a8 — severity is now `error`. Kept the rule `informational`: the trigger (a user agent that lacks user control) is internal UA state we cannot observe from any context, but `error` correctly reflects the MUST NOT.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept/recipient-should-process-q-parameter-as-weight.rule.ts
BaggersIO: Done in 2230f7a8 — severity is now `warn`. Kept `informational` since recipient-side parsing behaviour is not observable from the message, but `warn` reflects the SHOULD.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept/sender-should-send-q-parameter-last.rule.ts
BaggersIO: Done in 2230f7a8 — removed `.appliesTo('client', 'user-agent')`.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/vary/cache-must-not-use-response-without-matching-vary-headers.rule.ts
BaggersIO: Agreed — done in 2230f7a8. Dropped the empty `analytics` context; the rule is now `informational` (kept `error` for the MUST NOT) and stays scoped to `cache`.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/vary/origin-server-should-generate-vary-header.rule.ts
BaggersIO: Done in 2230f7a8. Split the implementation: `overrideStaticRule` uses `validateCommonHttpTransactions` and can only check that a `Vary` header is present (the common projection exposes header names, not values); `overrideTest` + `overrideAnalyticsRule` use `validateHttpTransactions` on the real headers and inspect the actual `Vary` value — flagging a missing `Vary`, and (when present and not `*`) a `Vary` that lists none of the negotiation headers the request actually exercised.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/vary/origin-server-should-generate-vary-header.rule.ts
BaggersIO: Done in 2230f7a8 — replaced the 2xx range with the explicit heuristically-cacheable set you listed: 200, 203, 204, 206, 300, 301, 308, 404, 405, 410, 414, 501.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/server-must-not-include-accept-encoding-for-non-content-coding-415-errors.rule.ts
BaggersIO: This decision is already satisfied by the PR as-submitted: the rule ships with `.type('static', 'test', 'analytics')` and `.severity('error')` (lines 13-14), so no code change was needed here. The `recommend` config override that lowers this rule to `hint` to reduce false-positive noise is future work in the thymian/rules recommended config, outside the scope of this PR. (The separate `origin server` -> `server` appliesTo change from the other thread is being tracked in your dedicated issue.)

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/codings-value-may-be-given-quality-value.rule.ts
BaggersIO: Done in 41c6ba0c — removed `.appliesTo('client', 'user-agent')`.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/server-must-not-include-accept-encoding-for-non-content-coding-415-errors.rule.ts
BaggersIO: Done in 41c6ba0c — now `.appliesTo('server')`. Correcting my note in the other thread: the tag removal itself belongs in this PR (applied now); thymian-internal#355 tracks the general analyzer problem that `appliesTo('server')` only matches the literal server role. Until that lands, this rule will not select HAR-imported `origin server` transactions in analyze — the accepted trade-off per your decision.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/vary/cache-must-not-use-response-without-matching-vary-headers.rule.ts
```diff
@@ -1,22 +1,10 @@
...
-  'X-Requested-With',
-];
-
 export default httpRule(
   'rfc9110/cache-must-not-use-response-without-matching-vary-headers',
 )
   .severity('error')
-  .type('analytics')
+  .type('informational')
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#name-vary')
```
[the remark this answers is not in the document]
BaggersIO: Restored in 7c5100de — the rule now carries a standalone in-file rationale (cache-hit linkage and request correlation are not available from a single transaction or HAR).

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-language/user-agent-must-not-send-accept-language-without-user-control.rule.ts
```diff
@@ -4,7 +4,7 @@ export default httpRule(
   'rfc9110/user-agent-must-not-send-accept-language-without-user-control',
 )
   .severity('error')
-  .type('analytics')
+  .type('informational')
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#name-accept-language')
```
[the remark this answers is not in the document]
BaggersIO: Restored in 7c5100de — in-file rationale added ('no user control' is internal user-agent state, unobservable from the wire).

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept/recipient-should-process-q-parameter-as-weight.rule.ts
```diff
@@ -4,9 +4,10 @@ export default httpRule(
   'rfc9110/recipient-should-process-q-parameter-as-weight',
 )
   .severity('warn')
-  .type('static')
+  .type('informational')
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#name-accept')
   .description(
```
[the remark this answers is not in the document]
BaggersIO: Restored in 7c5100de — in-file rationale added (recipient parsing behavior is not present in any message; sender-side ordering is covered by rfc9110/sender-should-send-q-parameter-last).

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept/sender-should-send-q-parameter-last.rule.ts
```diff
@@ -29,22 +36,43 @@ export default httpRule('rfc9110/sender-should-send-q-parameter-last')
...
 
         if (malformedValues.length > 0) {
           return [
             {
               location,
               violation: {
-                message: `The weight parameter "q" SHOULD be send as last parameter. Malformed value(s) sent: ${malformedValues.join(
-                  ', ',
-                )}`,
+                message: `The weight parameter "q" SHOULD be sent last, after all media-range parameters. The following value(s) place a media-range parameter after "q", where it is reinterpreted as an accept-extension parameter: ${malformedValues
```
matthyk: This is fine, but you could use the `createList` utility function

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/vary/origin-server-should-generate-vary-header.rule.ts
```diff
@@ -1,8 +1,143 @@
...
+  );
+
+  const varyValue = headerToString(getHeader(res.headers, 'vary'));
+
+  if (!varyValue || !varyValue.trim()) {
+    return [
+      {
+        location,
+        violation: {
+          message: `A cacheable ${req.method} response was returned for a request carrying proactive-negotiation header field(s) ${createList(
```
matthyk: I would prefer a shorter message. Something like "Response SHOULD contain Vary header with values: ${createList(presentHeaders)}." This message should not explain the whole rule.

matthyk (approved):
For me non-blocking changes.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept/sender-should-send-q-parameter-last.rule.ts
BaggersIO: Done in 16e532dd — the malformed-value list now goes through `createList` instead of the manual `map(...).join(', ')`.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/vary/origin-server-should-generate-vary-header.rule.ts
BaggersIO: Done in 16e532dd — shortened all three Vary messages to state the fix rather than restate the rule. The missing-Vary case (both the static and the value-based paths) now reads `Response SHOULD contain Vary header with values: <list>.`, and the insufficient-Vary case reads `Response Vary header (<value>) SHOULD contain at least one of the negotiated values: <list>.` The full rationale stays in the rule description/summary.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-charset/user-agent-may-associate-quality-value-with-charset.rule.ts
```diff
@@ -1,17 +1,80 @@
...
+          getHeader(req.headers, 'accept-charset'),
+        );
+
+        if (!acceptCharset || hasQualityValue(acceptCharset)) {
+          return [];
+        }
+
+        return [
+          { location, violation: { message: hintMessage }, findings: [] },
+        ];
```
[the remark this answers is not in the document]
BaggersIO: Done in c880b331. Split the hint into two parts: the deprecation note now surfaces whenever Accept-Charset is present, and the "you may attach a quality value" note is added only when no charset already carries a `q` weight. Previously the whole hint (deprecation included) was suppressed once a q-value appeared — dropping the more valuable half. Updated the misleading comment as well.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/accept-encoding/origin-server-should-send-response-without-content-coding.rule.ts
```diff
@@ -13,4 +93,46 @@ export default httpRule(
...
+        }
+
+        return [
+          {
+            location,
+            violation: {
+              message: `The response was encoded with content coding(s) ${createList(
+                unacceptable,
+              )} that the request's Accept-Encoding (${acceptEncoding.trim()}) does not list as acceptable. When no acceptable coding is available, the origin server SHOULD send the response without any content coding (unless identity is indicated as unacceptable).`,
+            },
```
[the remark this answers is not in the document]
BaggersIO: Done in c880b331 — the message no longer asserts "When no acceptable coding is available" (which the rule doesn't verify). It now states only the observed condition — an unacceptable content coding was applied per the request's Accept-Encoding — followed by the SHOULD to fall back to identity.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/vary/origin-server-should-generate-vary-header.rule.ts
```diff
@@ -1,8 +1,139 @@
...
+  );
+
+  const varyValue = headerToString(getHeader(res.headers, 'vary'));
+
+  if (!varyValue || !varyValue.trim()) {
+    return [
+      {
+        location,
+        violation: {
+          message: `Response SHOULD contain Vary header with values: ${createList(presentHeaders)}.`,
```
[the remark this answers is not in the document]
BaggersIO: Done in c880b331 — reworded to "field name(s)" since Vary lists header field names, not values.

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/vary/origin-server-should-generate-vary-header.rule.ts
```diff
@@ -1,8 +1,139 @@
...
+
+  if (presentHeaders.some((header) => varyFields.includes(header))) {
+    return [];
+  }
+
+  return [
+    {
+      location,
+      violation: {
+        message: `Response Vary header (${varyValue.trim()}) SHOULD contain at least one of the negotiated values: ${createList(presentHeaders)}.`,
```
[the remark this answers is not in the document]
BaggersIO: Done in c880b331 — this message now reads "negotiated field name(s)" instead of "negotiated values".

packages/rules-rfc-9110/src/rules/content-negotiation/content-negotiation-fields/vary/origin-server-should-generate-vary-header.rule.ts
```diff
@@ -11,4 +142,44 @@ export default httpRule('rfc9110/origin-server-should-generate-vary-header')
...
+      ) => {
+        if (hasHeader(res.headers, 'vary')) {
+          return [];
+        }
+
+        return [
+          {
+            location,
+            violation: {
+              message: `Response SHOULD contain Vary header with values: ${createList(presentNegotiationHeaders(req))}.`,
```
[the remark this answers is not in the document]
BaggersIO: Done in c880b331 — the static-context message now says "field name(s)" too, consistent with the other two Vary messages.

PR #299, opened 2026-06-29
feat(rules-rfc-9110): align authentication rule execution contexts (#327)

BaggersIO:
## Summary

Aligns every rule in the **authentication** category (`packages/rules-rfc-9110/src/rules/authentication/`) with the execution context(s) that can *actually and meaningfully* run, using only the existing rule framework (issue #327). Closes the gap where rules declared an active context but shipped no rule function (appearing "covered" while doing nothing), and makes every classification honest with a written rationale.

**12 rules reviewed** (11 `*.rule.ts` plus one orphaned `.ts` that the rule-set glob `rules/**/*.rule.js` never loaded).

## Coverage matrix

|---|---|---|---|---|
| `realm-parameter-must-use-quoted-string-syntax` | `test, analytics` | `test, analytics` | **1 (implement)** | `realm` is a challenge param (response-only). Dropped dead request-side scan; scoped to response auth headers + descriptive message; `appliesTo('origin server','proxy')` so it fires on HAR. |
| `authentication-parameter-name-must-occur-once-per-challenge` | `test, analytics` | `test, analytics` | **1 (implement)** | Syntactic check on auth-param values (`getHeader`). Test slot validates server challenges; analyze covers request credentials. Added `appliesTo('origin server','proxy','client','user-agent')` + violation message. |
| `proxy-must-not-modify-authorization` | `analytics` (no fn) | `analytics` (no fn) | **3 (infra-deferred)** | Per-hop modification detection needs correlated inbound/outbound at the proxy; documented rationale; `appliesTo('proxy')`. |
| `proxy-must-not-modify-www-authenticate` | `analytics` (no fn) | `analytics` (no fn) | **3 (infra-deferred)** | Same per-hop linkage requirement; documented rationale; `appliesTo('proxy')`. |
| `proxy-must-not-modify-authentication-info` | `informational` | `analytics` (no fn) | **4 → 3 (reclassify, infra-deferred)** | Mis-typed informational; same shape as its MUST-NOT-modify siblings; promoted to analytics with rationale; `appliesTo('proxy')`. |
| `server-may-send-www-authenticate-in-other-responses` | `static, analytics, test` (broken fn) | `informational` | **4 → 2 (reclassify, informational)** | Permissive MAY. Old fn flagged every response *lacking* the optional header as a violation (inverted, fired on all traffic). Removed fn + active contexts. |
| `origin-server-should-send-401-for-invalid-credentials` | `informational` | `informational` | **2 (informational)** | Validity-of-credentials / protected-resource status is server-internal ground truth; no observable non-conformant condition. Added rationale. |
| `proxy-should-send-407-for-invalid-proxy-credentials` | `informational` | `informational` | **2 (informational)** | Proxy-internal ground truth (requires-auth + credential validity); also a proxy behavior so `test` N/A. Added rationale. |
| `authentication-scheme-must-accept-token-and-quoted-string` | `informational` | `informational` | **2 (informational)** | Property of a scheme *definition*, not of any wire message. Added rationale. |
| `proxy-authenticate-applies-to-next-client` | `informational` | `informational` | **2 (informational)** | Statement of fact about field scope; no requirement. Added rationale. |
| `proxy-may-relay-credentials` | `informational` | `informational` | **2 (informational)** | Permissive MAY, proxy-internal. Added rationale. |
| `user-agent-may-reuse-same-credentials-for-identical-protection-space` | `informational` (not loaded; wrong filename + name) | `informational` | **2 + repair** | Permissive MAY, user-agent-internal. Renamed `.ts` → `.rule.ts` so the glob loads it; rule name aligned to filename. Added rationale. |

**Outcome counts:** Outcome 1 (implemented): **2** · Outcome 2 (informational): **6** · Outcome 3 (infra-deferred): **3** · Outcome 4 (reclassify): **2** (both then land in 2/3, counted above).

## Security-relevant rules

Authentication is security-sensitive throughout; each was handled explicitly:

- **proxy-must-not-modify-{authorization, www-authenticate, authentication-info}** — integrity of credentials and challenges across proxy hops. Kept as honest `analytics`/`appliesTo('proxy')` infra-deferred (outcome 3) rather than faked or downgraded, since they *are* validatable from proxy-recorded traffic. No active context ships without a function except these documented cases.
- **realm-parameter-must-use-quoted-string-syntax** — now actually enforces the MUST quoted-string syntax on server challenges (was a no-op-style check spread over the wrong headers).
- **authentication-parameter-name-must-occur-once-per-challenge** — now emits a descriptive violation; duplicate param names are matched case-insensitively per RFC 9110 §11.2.
- **server-may-send-www-authenticate-in-other-responses** — removed an actively wrong check that penalised servers for *not* using an optional permission (a noise/false-positive hazard), reclassified to informational.
- **origin-server-should-send-401 / proxy-should-send-407 for invalid credentials** — left informational because "invalid credentials" requires server-internal ground truth not present on the wire; documented.

## Reclassification reasoning (non-trivial)

- **server-may-send-www-authenticate-in-other-responses** (`static,test,analytics` → `informational`): the field declared three active contexts and a `validateCommonHttpTransactions(not(responseHeader('www-authenticate')))` call with no validation callback, so every transaction passing the filter (i.e. every response without WWW-Authenticate) was reported as a violation. The RFC text is a permissive MAY with no non-conformant condition, so the correct outcome is informational.
- **proxy-must-not-modify-authentication-info** (`informational` → `analytics`, no fn): identical in nature to the other two proxy MUST-NOT-modify rules (per-hop integrity, validatable only from proxy-recorded traffic). It was inconsistently typed `informational`; unified to the infra-deferred analytics shape.
- **realm scope narrowing**: `realm` only appears in challenges (response auth headers); the previous request-side scan was dead code. Scoped to response headers and `appliesTo('origin server','proxy')` so response-side analyze fires on HAR (HAR response role defaults to `origin server`).
- **orphaned protection-space file**: lacked the `.rule.ts` suffix the rule-set glob requires, so it never loaded, and its internal name (`protection-space-cannot-extend-beyond-origin`) did not match the file. Renamed to `.rule.ts` and aligned the name; kept informational (permissive MAY, user-agent-internal).

## Per-PR checklist

- [x] Every rule in the category reviewed exactly once (12 rules).
- [x] No declared active context (`static`/`test`/`analytics`) left without a rule function, except the documented outcome-3 infrastructure rules.
- [x] No request-side rule assigned to `test` (realm scoped to responses; param-once test slot validates server challenges, request side is inert there).
- [x] Every informational and infra-deferred rule carries a written reason in a code comment near `.type(...)`.
- [x] Header comparisons case-insensitive (`toLowerCase`/`equalsIgnoreCase`); header VALUES read via `validateHttpTransactions` + `getHeader`.
- [x] `appliesTo` uses `'origin server'` (not bare `'server'`) for response-side analyze rules so they fire on HAR.
- [x] Security-relevant rules (auth, proxy integrity) addressed explicitly.
- [x] `npx nx lint rules-rfc-9110` shows 0 new errors/warnings in authentication files (the only lint errors are a pre-existing, package-wide `@nx/enforce-module-boundaries`/`dependency-checks` failure present identically on the clean tree — 568 errors before and after — affecting all categories incl. untouched `content-negotiation`).
- [x] `prettier --check` clean for the category.
- [x] Edited only files under `packages/rules-rfc-9110/src/rules/authentication/**`.
- [x] PR kept as draft.

A prior commit on this branch (`38027243`) removed the execution-context triage rationale comments from every authentication rule. This round restores them and upgrades three infra-deferred rules to real implementations.

### FIX-1 — proxy MUST-NOT-modify rules (outcome-3 → implemented)
`proxy-must-not-modify-authorization`, `proxy-must-not-modify-www-authenticate`, and `proxy-must-not-modify-authentication-info` were `type('analytics')` + `appliesTo('proxy')` with **no rule function and no rationale**. Both are now fixed:
- Restored the outcome-3 rationale comment near `.type('analytics')` explaining that detecting per-hop credential/challenge modification requires correlating the proxy's inbound vs outbound hop, available only from proxy-recorded traffic.
- Implemented the check via `validateCapturedHttpTraces`, comparing the header VALUE (`Authorization` on the request side; `WWW-Authenticate` / `Authentication-Info` on the response side) across adjacent hops whose request/response `meta.role === 'proxy'`. This follows the existing sibling pattern in `routing/message-transformations/`. So these are now genuine implemented checks rather than bare outcome-3 declarations.

### FIX-2 — `authentication-parameter-name-must-occur-once-per-challenge` (mixing / request-side-in-test)
Previously `type('test','analytics')` with a **single** `.rule()` that scanned BOTH request auth headers (`authorization` / `proxy-authorization`) and response challenge headers. The request-side scan is inert in `test` (the request is Thymian-generated). Split into:
- `overrideTest`: scans ONLY response challenge headers (`getHeader` on the response) for duplicate challenge-parameter names.
- `overrideAnalyticsRule`: scans BOTH request-side credentials and response-side challenges against recorded traffic.
Both share a new `scanForDuplicateParameterNames` helper and reuse the existing `parseChallenges` util and the `requestAuthenticationHeaders` / `responseAuthenticationHeaders` name lists.

### FIX-3 — restored informational rationale
Restored the outcome-2 informational rationale comments removed by `38027243` on: `origin-server-should-send-401-for-invalid-credentials`, `proxy-should-send-407-for-invalid-proxy-credentials`, `server-may-send-www-authenticate-in-other-responses`, `authentication-scheme-must-accept-token-and-quoted-string`, `proxy-authenticate-applies-to-next-client`, `proxy-may-relay-credentials`, and `user-agent-may-reuse-same-credentials-for-identical-protection-space`; plus the outcome-1 rationale on `realm-parameter-must-use-quoted-string-syntax`.

Build gate: `tsc -b packages/rules-rfc-9110/tsconfig.lib.json` exits 0. Files formatted with prettier. PR remains a draft.


packages/rules-rfc-9110/src/rules/authentication/origin-server/authentication-info/proxy-must-not-modify-authentication-info.rule.ts
```diff
@@ -2,7 +2,7 @@ import { httpRule } from '@thymian/core';
 
 export default httpRule('rfc9110/proxy-must-not-modify-authentication-info')
   .severity('error')
-  .type('informational')
+  .type('analytics')
```
BaggersIO, line 13: Infrastructure-deferred (outcome 3): this MUST NOT governs a *proxy hop* — a proxy must forward the Authentication-Info response field byte-for-byte. Detecting a violation requires correlating the message a proxy received with the message it forwarded (the inbound and outbound sides of the same hop) and observing that the field value changed. That two-sided, per-hop linkage is not available from a single transaction and is not reliably reconstructable from a typical HAR; it depends on traffic captured at the proxy itself. Because it *is* validatable from proxy-recorded traffic, the rule stays typed `analytics` and scoped to the `proxy` role rather than being downgraded to informational. Previously mis-typed `informational`; reclassified to `analytics` to match its sibling MUST-NOT-modify rules.

packages/rules-rfc-9110/src/rules/authentication/origin-server/www-authenticate/server-may-send-www-authenticate-in-other-responses.rule.ts
```diff
@@ -1,19 +1,15 @@
-import { not, responseHeader } from '@thymian/core';
 import { httpRule } from '@thymian/core';
 
 export default httpRule(
   'rfc9110/server-may-send-www-authenticate-in-other-responses',
 )
   .severity('hint')
-  .type('static', 'analytics', 'test')
+  .type('informational')
```
BaggersIO: Informational (outcome 2): this is a permissive MAY with no non-conformant condition. A server is allowed — but never required — to add a WWW-Authenticate header field to responses other than 401 to hint that (different) credentials might change the result. Neither presence nor absence of the header is a violation, so there is nothing to validate. The previous implementation declared `static, test, analytics` and called `validateCommonHttpTransactions(not(responseHeader('www-authenticate')))` with NO validation callback, which flags every response that does *not* carry WWW-Authenticate as a violation — i.e. it penalised servers for not exercising an optional permission. That inverts the rule and fires on essentially all traffic. The active contexts and rule function have been removed in favour of an honest informational classification.

packages/rules-rfc-9110/src/rules/authentication/parameters/authentication-parameter-name-must-occur-once-per-challenge.rule.ts
```diff
@@ -24,6 +24,10 @@ export default httpRule(
   .description(
     'Authentication parameter names are matched case-insensitively and each parameter name MUST only occur once per challenge.',
   )
+  .summary(
+    'Each authentication parameter name MUST only occur once per challenge (matched case-insensitively).',
+  )
+  .appliesTo('origin server', 'proxy', 'client', 'user-agent')
```
BaggersIO: Implementable (outcome 1): this is a syntactic conformance check on the auth-param syntax carried in authentication header field values, readable via `validateHttpTransactions` + `getHeader` (header VALUES, not just names). It applies to challenges (WWW-Authenticate / Proxy-Authenticate, server-sent) and to credentials (Authorization / Proxy-Authorization, client-sent). The `test` slot validates the server-generated challenges in live responses (a server-behavior check); the request-side credentials are validated against recorded traffic (`analyze`). In `test` the request is generated by Thymian and is always well-formed, so the request-side scan cannot produce a false positive there.

appliesTo: Fire on HAR for both directions: response challenges (origin server / proxy) and request credentials (client / user-agent).

(Anchored to the `.appliesTo` line at line 30 rather than the `.type` line: in the final PR diff this rule's `.type` line falls between hunks, so `.appliesTo` is the nearest in-hunk changed line.)

packages/rules-rfc-9110/src/rules/authentication/realm/realm-parameter-must-use-quoted-string-syntax.rule.ts
```diff
@@ -24,29 +19,33 @@ export default httpRule('rfc9110/realm-parameter-must-use-quoted-string-syntax')
   .description(
     'For historical reasons, a sender MUST only generate the quoted-string syntax for realm parameter values.',
   )
+  .summary(
+    'A sender MUST only generate the quoted-string syntax for realm parameter values.',
+  )
+  .appliesTo('origin server', 'proxy')
```
BaggersIO: Implementable (outcome 1): the `realm` auth-param is a *challenge* parameter — it appears only in the server-generated WWW-Authenticate / Proxy-Authenticate response headers, never in request credentials (Authorization carries no `realm`). It is therefore a server-behavior check: meaningful in `test` (validate the live response Thymian receives) and in `analyze` (recorded traffic). The earlier implementation also scanned request auth headers, which was dead code for this rule; that scan has been dropped so the rule is scoped to the response challenges it actually governs.

appliesTo: Challenges are emitted by the origin server (401) or a proxy (407); include 'origin server' so response-side analyze fires on HAR.

(Anchored to the `.appliesTo` line at line 25 rather than the `.type` line: in the final PR diff this rule's `.type` line falls between hunks, so `.appliesTo` is the nearest in-hunk changed line.)

packages/rules-rfc-9110/src/rules/authentication/realm/user-agent-may-reuse-same-credentials-for-identical-protection-space.rule.ts
```diff
@@ -1,6 +1,8 @@
 import { httpRule } from '@thymian/core';
 
-export default httpRule('rfc9110/protection-space-cannot-extend-beyond-origin')
+export default httpRule(
+  'rfc9110/user-agent-may-reuse-same-credentials-for-identical-protection-space',
+)
   .severity('hint')
   .type('informational')
```
BaggersIO, line 14: Informational (outcome 2): a permissive MAY describing user-agent-internal state — once a request is authorized, the user agent is allowed (but not required) to reuse the same credentials for other requests in the same protection space, for a duration of its own choosing. Both reusing and not reusing are conformant, the duration is implementation-defined, and the protection space is user-agent-internal, so there is no observable non-conformant condition to validate. This file previously lacked the `.rule.ts` suffix required by the rule-set glob (`rules/**/*.rule.js`), so it was never loaded; its rule name also did not match the file. It has been renamed to `.rule.ts` and the rule name aligned to the filename so the (informational) rule is honestly registered.

BaggersIO (commented):
Recording the execution-context triage rationale here (moved out of inline code comments).

Inline line-anchored comments below carry the rationale for the rules whose lines remain in this PR's diff. The rules below are now identical to `main` after the triage comments were removed (their only PR change was the comment, or a `.type` already matching `main`), so they have no diff line to anchor to; their rationale is recorded here in the review body instead:

**`packages/rules-rfc-9110/src/rules/authentication/credentials/origin-server-should-send-401-for-invalid-credentials.rule.ts`** (`.type` line): Informational (outcome 2): the SHOULD is conditioned on the request having omitted, invalid, or partial credentials for a *protected* resource. Whether supplied credentials are valid, and whether a resource is protected, is origin-server-internal ground truth that is not observable from the wire: a 200 with credentials might be a public resource or a successful auth, and a non-401 with bad credentials is indistinguishable from a request the server simply did not treat as protected. Without that ground truth there is no reliable non-conformant condition to detect, so this stays informational.

**`packages/rules-rfc-9110/src/rules/authentication/credentials/proxy-should-send-407-for-invalid-proxy-credentials.rule.ts`** (`.type` line): Informational (outcome 2): the SHOULD is conditioned on the proxy requiring authentication and the request having omitted, invalid, or partial proxy credentials. Whether a proxy requires authentication and whether supplied proxy credentials are valid is proxy-internal ground truth not observable from the wire, so there is no reliable non-conformant condition to detect. (It is also a proxy behavior; Thymian is not the proxy, so `test` is not applicable either.) This stays informational.

**`packages/rules-rfc-9110/src/rules/authentication/origin-server/authorization/proxy-must-not-modify-authorization.rule.ts`** (`.type` line): Infrastructure-deferred (outcome 3): this MUST NOT governs a *proxy hop* — a proxy forwarding a request must not alter its Authorization header field. Detecting a violation requires correlating the request a proxy received with the request it forwarded and observing the field value changed. That two-sided, per-hop linkage is not available from a single transaction and is not reliably reconstructable from a typical HAR; it depends on traffic captured at the proxy itself. Because it *is* validatable from proxy-recorded traffic, the rule stays typed `analytics` and scoped to the `proxy` role, without a rule function, per the outcome-3 contract.

**`packages/rules-rfc-9110/src/rules/authentication/origin-server/www-authenticate/proxy-must-not-modify-www-authenticate.rule.ts`** (`.type` line): Infrastructure-deferred (outcome 3): this MUST NOT governs a *proxy hop* — a proxy forwarding a response must not alter its WWW-Authenticate header field. Detecting a violation requires correlating the response a proxy received with the response it forwarded and observing the field value changed. That two-sided, per-hop linkage is not available from a single transaction and is not reliably reconstructable from a typical HAR; it depends on traffic captured at the proxy itself. Because it *is* validatable from proxy-recorded traffic, the rule stays typed `analytics` and scoped to the `proxy` role, without a rule function, per the outcome-3 contract.

**`packages/rules-rfc-9110/src/rules/authentication/parameters/authentication-scheme-must-accept-token-and-quoted-string.rule.ts`** (`.type` line): Informational (outcome 2): this constrains how an authentication *scheme definition* is specified — a scheme's grammar must accept both token and quoted-string notations for parameter values, for senders and recipients alike. It is a property of the scheme specification, not of any individual HTTP message on the wire, so there is nothing in a request, response, or recorded transaction to validate.

**`packages/rules-rfc-9110/src/rules/authentication/proxy/proxy-authenticate/proxy-authenticate-applies-to-next-client.rule.ts`** (`.type` line): Informational (outcome 2): this is a statement of fact about the *scope* of the Proxy-Authenticate field (it applies only to the next outbound client on the response chain, unlike WWW-Authenticate). It expresses no requirement and defines no non-conformant condition, so there is nothing to validate on the wire.

**`packages/rules-rfc-9110/src/rules/authentication/proxy/proxy-authorization/proxy-may-relay-credentials.rule.ts`** (`.type` line): Informational (outcome 2): a permissive MAY describing internal proxy behavior — a proxy is allowed (but not required) to relay client credentials to the next proxy when proxies cooperatively authenticate. There is no non-conformant condition (relaying and not relaying are both permitted), and the relaying decision is proxy-internal, so there is nothing to validate.

packages/rules-rfc-9110/src/rules/authentication/origin-server/authentication-info/proxy-must-not-modify-authentication-info.rule.ts
matthyk: If this rule is marked as `analytics`, the rule requires a rule imlementation.

packages/rules-rfc-9110/src/rules/authentication/origin-server/www-authenticate/server-may-send-www-authenticate-in-other-responses.rule.ts
matthyk: This behavior is intended. `hints` are meant to be provide such optional HTTP functionalities. I would keep the rule's implementation, type and the `server` tag.

packages/rules-rfc-9110/src/rules/authentication/parameters/authentication-parameter-name-must-occur-once-per-challenge.rule.ts
matthyk:
```suggestion
```

packages/rules-rfc-9110/src/rules/authentication/realm/realm-parameter-must-use-quoted-string-syntax.rule.ts
matthyk:
```suggestion
```

`sender` applies to every component which sends requests/responses.

matthyk (changes requested):

packages/rules-rfc-9110/src/rules/authentication/origin-server/authentication-info/proxy-must-not-modify-authentication-info.rule.ts
BaggersIO: Implemented in 5dca4b32: the rule now correlates the two sides of a proxy hop via `validateCapturedHttpTraces` and reports any change to the field, including additions and removals (the Authorization and WWW-Authenticate siblings got the same treatment). ea330e10 additionally makes the comparison robust against representation differences (string vs. one-element array, unstable order of repeated field lines).

packages/rules-rfc-9110/src/rules/authentication/origin-server/www-authenticate/server-may-send-www-authenticate-in-other-responses.rule.ts
BaggersIO: Reverted in 0c31cfae: implementation, `type("static", "analytics", "test")` and the `server` tag are restored — the file is identical to `main` again.

packages/rules-rfc-9110/src/rules/authentication/parameters/authentication-parameter-name-must-occur-once-per-challenge.rule.ts
BaggersIO: Applied in 0c31cfae — the `.appliesTo(...)` line is removed.

packages/rules-rfc-9110/src/rules/authentication/realm/realm-parameter-must-use-quoted-string-syntax.rule.ts
BaggersIO: Applied in 0c31cfae — the `.appliesTo(...)` line is removed. Good point on the sender scope: since some schemes carry `realm` in request credentials as well (e.g. Digest, RFC 7616), f2f9d08e extends the `analytics` context to also validate request-side Authorization / Proxy-Authorization values, mirroring the test/analytics split of the duplicate-parameter rule. `test` still validates only the server-generated challenges, since Thymian generates the request there.

packages/rules-rfc-9110/src/rules/authentication/parameters/authentication-parameter-name-must-occur-once-per-challenge.rule.ts
```diff
@@ -13,48 +14,101 @@ import {
...
+        if (seenParams.has(lowerName)) {
+          return [
+            {
+              location,
+              violation: {
+                message: `The authentication parameter "${param.name}" occurs more than once in a single challenge of "${headerValue}". Each authentication parameter name MUST only occur once per challenge.`,
+              },
+              findings: [],
+            },
+          ];
```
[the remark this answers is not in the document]
BaggersIO: Fixed in 0c31cfae: the violation message now names only the duplicated parameter; header values are no longer echoed.

packages/rules-rfc-9110/src/rules/authentication/realm/realm-parameter-must-use-quoted-string-syntax.rule.ts
```diff
@@ -2,51 +2,61 @@ import {
...
             for (const param of challenge.parameters) {
               const isRealmParam = equalsIgnoreCase(param.name, 'realm');
               if (isRealmParam && !param.isQuoted) {
-                return [{ location, violation: {}, findings: [] }]; // Violation: realm used token syntax instead of quoted-string
+                return [
+                  {
+                    location,
+                    violation: {
+                      message: `The realm parameter in "${headerValue}" uses token syntax (realm=${param.value}). For historical reasons a sender MUST only generate the quoted-string syntax for realm parameter values (realm="${param.value}").`,
+                    },
```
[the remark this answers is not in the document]
BaggersIO: Fixed in 0c31cfae: the message now reports only the offending `realm` parameter and the required quoted-string form, not the full header value.

packages/rules-rfc-9110/src/rules/authentication/origin-server/authorization/proxy-must-not-modify-authorization.rule.ts
```diff
@@ -10,4 +18,41 @@ export default httpRule('rfc9110/proxy-must-not-modify-authorization')
...
+        }
+
+        const forwarded = getHeader(prev.request.data.headers, 'authorization');
+        const received = getHeader(curr.request.data.headers, 'authorization');
+
+        // A proxy that never received an Authorization field cannot have
+        // modified one; only compare when it is present on the inbound side.
+        if (received === undefined) {
+          continue;
+        }
```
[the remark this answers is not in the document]
BaggersIO: Fixed in 0c31cfae: the guard now skips only when the field is absent on both sides, so removals (and additions) are reported as modifications.

packages/rules-rfc-9110/src/rules/authentication/origin-server/www-authenticate/proxy-must-not-modify-www-authenticate.rule.ts
```diff
@@ -10,4 +19,47 @@ export default httpRule('rfc9110/proxy-must-not-modify-www-authenticate')
...
+        const received = getHeader(
+          curr.response.data.headers,
+          'www-authenticate',
+        );
+
+        // A proxy that never received a WWW-Authenticate field cannot have
+        // modified one; only compare when it is present on the inbound side.
+        if (received === undefined) {
+          continue;
+        }
```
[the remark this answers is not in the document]
BaggersIO: Fixed in 0c31cfae: the guard now skips only when the field is absent on both sides, so additions (and removals) are reported as modifications.

packages/rules-rfc-9110/src/rules/authentication/origin-server/authentication-info/proxy-must-not-modify-authentication-info.rule.ts
```diff
@@ -1,13 +1,65 @@
...
+        const received = getHeader(
+          curr.response.data.headers,
+          'authentication-info',
+        );
+
+        // A proxy that never received an Authentication-Info field cannot have
+        // modified one; only compare when it is present on the inbound side.
+        if (received === undefined) {
+          continue;
+        }
```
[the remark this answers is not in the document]
BaggersIO: Fixed in 0c31cfae: the guard now skips only when the field is absent on both sides, so additions (and removals) are reported as modifications.

packages/rules-rfc-9110/src/rules/authentication/origin-server/www-authenticate/proxy-must-not-modify-www-authenticate.rule.ts
```diff
@@ -10,4 +15,48 @@ export default httpRule('rfc9110/proxy-must-not-modify-www-authenticate')
...
+        );
+
+        // Nothing to compare only when neither hop carried the field. Present
+        // on exactly one side means the proxy added or removed it — itself a
+        // modification this MUST NOT forbids.
+        if (forwarded === undefined && received === undefined) {
+          continue;
+        }
+
+        if (JSON.stringify(forwarded) !== JSON.stringify(received)) {
```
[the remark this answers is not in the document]
BaggersIO: Fixed in ea330e10: both sides are normalized via the shared `equalHeaderValues` helper (coerced to a sorted array, empty meaning absent) before comparison, so representation and ordering differences no longer produce false violations.

packages/rules-rfc-9110/src/rules/authentication/origin-server/authorization/proxy-must-not-modify-authorization.rule.ts
```diff
@@ -10,4 +15,42 @@ export default httpRule('rfc9110/proxy-must-not-modify-authorization')
...
+        const received = getHeader(curr.request.data.headers, 'authorization');
+
+        // Nothing to compare only when neither hop carried the field. Present
+        // on exactly one side means the proxy added or removed it — itself a
+        // modification this MUST NOT forbids.
+        if (forwarded === undefined && received === undefined) {
+          continue;
+        }
+
+        if (JSON.stringify(forwarded) !== JSON.stringify(received)) {
```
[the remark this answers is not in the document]
BaggersIO: Fixed in ea330e10: both sides are normalized via the shared `equalHeaderValues` helper (coerced to a sorted array, empty meaning absent) before comparison, so representation and ordering differences no longer produce false violations.

packages/rules-rfc-9110/src/rules/authentication/origin-server/authentication-info/proxy-must-not-modify-authentication-info.rule.ts
```diff
@@ -1,13 +1,63 @@
...
+        );
+
+        // Nothing to compare only when neither hop carried the field. Present
+        // on exactly one side means the proxy added or removed it — itself a
+        // modification this MUST NOT forbids.
+        if (forwarded === undefined && received === undefined) {
+          continue;
+        }
+
+        if (JSON.stringify(forwarded) !== JSON.stringify(received)) {
```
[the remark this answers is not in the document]
BaggersIO: Fixed in ea330e10: both sides are normalized via the shared `equalHeaderValues` helper (coerced to a sorted array, empty meaning absent) before comparison, so representation and ordering differences no longer produce false violations.

packages/rules-rfc-9110/src/rules/authentication/realm/realm-parameter-must-use-quoted-string-syntax.rule.ts
```diff
@@ -2,51 +2,54 @@ import {
...
+import { responseAuthenticationHeaders } from '../utils/authentication-header-names.js';
 
 export default httpRule('rfc9110/realm-parameter-must-use-quoted-string-syntax')
   .severity('error')
+  // The `realm` auth-param is a *challenge* parameter — it appears only in the
+  // server-generated WWW-Authenticate / Proxy-Authenticate response headers,
+  // never in request credentials (Authorization carries no `realm`). It is
+  // therefore a server-behavior check: meaningful in `test` (validate the live
+  // response Thymian receives) and in `analyze` (recorded traffic).
   .type('test', 'analytics')
```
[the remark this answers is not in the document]
BaggersIO: The role restriction was removed deliberately at maintainer request (see matthyk’s thread on this file): RFC 9110 binds the quoted-string requirement to any *sender*, not only origin servers and proxies. As of f2f9d08e the `analytics` context validates request-side credentials too (e.g. Digest carries `realm` in Authorization), so an origin-server/proxy role filter would be wrong here.

matthyk (approved):

packages/rules-rfc-9110/src/rules/authentication/utils/header-value-equality.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export function equalHeaderValues(
+  a: string | string[] | undefined,
+  b: string | string[] | undefined,
+): boolean {
+  return (
+    JSON.stringify(normalizedHeaderValue(a)) ===
+    JSON.stringify(normalizedHeaderValue(b))
+  );
+}
```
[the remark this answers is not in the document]
matthyk: Let's skip such performance optimizations. In the future we should some benchmarks with high volume data/ large API descriptions and identify bottlenecks.

PR #301, opened 2026-06-29
feat(rules-rfc-9110): align message-abstraction rule execution contexts (#327)

BaggersIO:
## Summary

Aligns every rule in the **message-abstraction** category (`packages/rules-rfc-9110/src/rules/message-abstraction/`) with the execution context(s) that can *actually and meaningfully* run under the existing rule framework (issue #327). Each of the 26 rules was reviewed exactly once and assigned exactly one of the four triage outcomes. No framework, `core`, plugin, or other-category files were touched.

The category is dominated by RFC 9110 §6 material that is definitional, version-negotiation-related, or describes internal recipient/sender behavior — most of it is genuinely unobservable through the rule framework and correctly lands as `informational` (now with written justifications). The observable, message-level requirements (Date presence on 2xx/3xx/4xx, forbidden fields in trailers, Trailer-header-with-trailers) are implemented or were re-pointed at the right context, and two permissive rules that previously shipped functions flagging the *conformant* case were reclassified.

## Coverage matrix (rule × before → after × outcome)

|---|---|---|---|---|
| client-must-retain-knowledge-of-request | informational | informational | 2 | Internal client parse/cache behavior; unobservable. Rationale added. |
| control-data/client-may-send-lower-version-for-broken-servers | informational | informational | 2 | HTTP version token not exposed by framework; conditional MAY. Rationale added. |
| control-data/client-must-not-send-non-conformant-version | informational | informational | 2 | Version token unobservable; conformance is internal client state. Rationale added. |
| control-data/client-should-send-highest-conformant-version | informational | informational | 2 | Version token unobservable. Rationale added. |
| control-data/recipient-should-process-higher-minor-version-as-highest-known | informational | informational | 2 | Internal recipient processing; version unobservable. Rationale added. |
| control-data/server-must-not-send-non-conformant-version | informational | informational | 2 | Response version token not exposed by framework. Rationale added. |
| control-data/server-should-send-response-version-equal-to-highest-conformant | informational | informational | 2 | Response version token unobservable. Rationale added. |
| framing/each-http-version-defines-own-framing | informational | informational | 2 | Pure definition; no requirement. Rationale added. |
| framing/implicit-framing-allowed-for-backwards-compatibility | informational | informational | 2 | Transport/connection framing, below the observed abstraction. Rationale added. |
| framing/message-complete-when-framed-octets-available | informational | informational | 2 | Definitional (message "complete"); transport-level. Rationale added. |
| header-fields/header-field-term-for-header-section-only-fields | informational | informational | 2 | Terminology definition. Rationale added. |
| header-fields/header-fields-sent-before-content | informational | informational | 2 | Terminology definition. Rationale added. |
| header-fields/header-section-consists-of-field-lines | informational | informational | 2 | Structural/terminology definition. Rationale added. |
| message-metadata/date/origin-server-may-generate-date-for-1xx-5xx | analytics (with fn) | informational | 2 | **Reclassified.** Permissive MAY — omitting Date is conformant; old fn flagged that conformant case. Rule fn removed. |
| message-metadata/date/origin-server-with-clock-must-generate-date-for-2xx-3xx-4xx | static,analytics,test (with fn) | static,analytics,test (with fn) | 1 | Implemented presence check over common projection; `appliesTo('origin server')`. Rationale added. |
| message-metadata/date/origin-server-without-clock-must-not-generate-date | informational | informational | 2 | Depends on server having no clock (internal state). Rationale added. |
| message-metadata/date/recipient-may-replace-invalid-date | informational | informational | 2 | Internal recipient MAY. Rationale added. |
| message-metadata/date/recipient-with-clock-must-add-date-if-missing | informational | informational | 2 | Internal recipient/cache/forward behavior. Rationale added. |
| message-metadata/date/sender-should-generate-date-at-message-generation | analytics,informational | informational | 4 | **Resolved mixed type.** True generation time unobservable. |
| message-metadata/date/user-agent-may-send-date-header-in-request | analytics (with fn) | informational | 2 | **Reclassified.** Request-side permissive MAY; old fn flagged requests *without* Date (the conformant case). Rule fn removed. |
| message-metadata/trailer/sender-should-generate-trailer-header-when-sending-trailers | analytics,test (with fn) | analytics,test (with fn) | 1 | Response-side; added `appliesTo('server','origin server')`; tidied callback to use provided `location`. |
| trailer-fields/recipient-may-treat-the-set-of-received-trailer-fields-as-name-value-pairs | informational | informational | 2 | Internal recipient MAY. Rationale added. |
| trailer-fields/recipient-must-not-merge-trailers-unsafely | informational | informational | 2 | Intermediary-internal merge decision; unobservable. Rationale added (security note). |
| trailer-fields/sender-must-not-generate-trailer-unless-permitted | analytics (with fn) | analytics (with fn) | 1 | **Bug fix** + `appliesTo('server','origin server')`. Now iterates trailer-NAME list directly (was `Object.keys` over a `string[]`, yielding indices). |
| trailer-fields/server-should-not-generate-necessary-trailers | informational | informational | 2 | Depends on server's belief a trailer is "necessary" (intent). Rationale added. |
| trailer-fields/trailer-fields-must-be-defined-as-list-if-repeatable | informational | informational | 2 | Constraint on field *definitions*, not on any message. Rationale added. |

### Outcome counts

- **Outcome 1 — implemented now:** 3 (origin-server-with-clock-must-generate-date-for-2xx-3xx-4xx; sender-should-generate-trailer-header-when-sending-trailers; sender-must-not-generate-trailer-unless-permitted)
- **Outcome 2 — informational (justified):** 22
- **Outcome 3 — infrastructure-deferred (analytics, no fn):** 0 (no rule in this category requires cross-transaction or proxy/cache-internal correlation while remaining HAR-recoverable)
- **Outcome 4 — reclassified/resolved:** 1 explicit mixed-type resolution (sender-should-generate-date-at-message-generation). Two additional MAY rules were also moved off active contexts to informational (counted under outcome 2 above as their honest end state).

## Per-PR checklist

- [x] Every rule in the category reviewed exactly once.
- [x] No declared active context (`static`/`test`/`analytics`) left without a rule function (no outcome-3 deferrals in this category).
- [x] No request-side rule assigned to `test` (the two request-side MAY rules became informational).
- [x] Every `informational` rule carries a written reason in a code comment near `.type(...)`.
- [x] Header/trailer name comparisons are case-insensitive (`.toLowerCase()`).
- [x] Response-side analyze rules use `appliesTo('origin server')` / `('server','origin server')` so they fire on HAR (whose responses default to the `origin server` role).
- [x] Common projection used only for header/trailer **names** + status (Date presence; forbidden trailer names); live `validateHttpTransactions` used where the actual trailer object is needed.
- [x] Security-relevant rules addressed explicitly (see below).
- [x] `npx prettier --write` applied; re-checked clean.
- [x] `eslint` over the category: 0 warnings and no new errors. (The package-wide `@nx/enforce-module-boundaries` "Imports of apps are forbidden" error is a pre-existing worktree path-resolution artifact present on every file in the package — including untouched ones — and is unrelated to these changes; confirmed it persists with this PR's changes stashed.)

## Security-relevant rules

- **trailer-fields/recipient-must-not-merge-trailers-unsafely** — unsafe trailer→header merging is a request-smuggling / header-injection vector. The decision happens *inside* an intermediary/recipient and is not visible in the observed transaction, so it stays `informational` with an explicit security note rather than being given a misleading active context.
- **trailer-fields/sender-must-not-generate-trailer-unless-permitted** — keeps a real, implemented `analytics` check that flags forbidden fields (`authorization`, `set-cookie`, `transfer-encoding`, `www-authenticate`, etc.) appearing in trailers, where they could be silently dropped or mishandled. Bug fixed so the check actually inspects trailer field names; scoped to the response sender.
- **message-metadata/date** rules touch caching freshness; the only observable one (Date presence on 2xx/3xx/4xx) remains implemented. The cache/forward-side Date behaviors are internal recipient state and are honestly `informational`.

## Reasoning for non-trivial reclassifications

- **origin-server-may-generate-date-for-1xx-5xx** and **user-agent-may-send-date-header-in-request** were typed `analytics` with functions that reported every transaction lacking the (optional) Date header. Since both are permissive MAYs, omitting Date is fully conformant — the implementations flagged the conformant case as a violation. Both are now `informational` with no function.
- **sender-should-generate-date-at-message-generation** mixed `analytics` with `informational`. Verifying "Date ≈ generation time" requires the true message-origination instant, which a recorder does not have (and the RFC explicitly lets the sender choose any instant during origination). Resolved to a single `informational` classification.
- **sender-must-not-generate-trailer-unless-permitted** was correct in spirit but iterated `Object.keys(res.trailers)` where, in the common projection, `res.trailers` is already the trailer-**name** array (`string[]`) — so `Object.keys` returned `['0','1',…]` and the forbidden-field check never matched. Fixed to filter the name list directly; added `appliesTo('server','origin server')` so it evaluates HAR responses.

A prior commit on this branch (`4df1103d`, "remove execution-context triage rationale from message-abstraction rules") had deleted the inline per-rule triage rationale from all 26 changed rule files, moving that reasoning into the PR body only. This directly violated the #327 hard requirement that every informational (and every infra-deferred) rule carry a written reason **in the code** adjacent to `.type(...)`, and it made the checklist item "Every `informational` rule carries a written reason in a code comment near `.type(...)`" false in the tree even though it was still checked here.

**What changed this round:**

- Restored the inline rationale comments to all 26 message-abstraction rule files (revert of `4df1103d`). This re-establishes the written reason near `.type(...)` for:
  - all **23 informational** rules in the category (outcome 2, plus the outcome-4 mixed-type resolution on `sender-should-generate-date-at-message-generation`), each stating why the requirement is unobservable (permissive MAY, terminology/definitional statement, transport control data not surfaced by the framework, or internal client/recipient/server state), and
  - the **3 implemented** (outcome 1) rules (`origin-server-with-clock-must-generate-date-for-2xx-3xx-4xx`, `sender-should-generate-trailer-header-when-sending-trailers`, `sender-must-not-generate-trailer-unless-permitted`), whose comments document the chosen execution context and `appliesTo` scoping.
- Build gate re-verified: `npx tsc -b packages/rules-rfc-9110/tsconfig.lib.json` exits 0. Prettier re-applied to the restored files (no formatting drift).

**Coverage-matrix reconciliation (FIX-2):** the outcome counts above (23 informational + implemented + 1 outcome-4 resolution) reflect the actual end state of the category. The checklist item "Every `informational` rule carries a written reason in a code comment near `.type(...)`" is now **accurate again in code** — verified that all 23 files matching `type('informational'` contain an `// Informational` rationale block adjacent to the rule declaration. No rule logic was changed in this round; this is a documentation-integrity fix only.


BaggersIO (commented):
Recording the execution-context triage rationale here (moved out of inline code comments) per issue #327.

**Line-anchored comments below** cover the 5 rules whose `.type(` line is part of this PR's diff.

**The remaining rules in this category had no change other than the triage comment**, so after removing it they are byte-identical to `main` and are no longer in the PR diff — GitHub cannot line-anchor a comment to them. Their rationale is recorded here:

- **client-must-retain-knowledge-of-request** (`rfc9110/client-must-retain-knowledge-of-request`, `.type('informational')`): Informational (outcome 2): this MUST governs how a client internally parses and caches a response (it must keep the corresponding request in mind, e.g. HEAD responses look like GET responses but must not be parsed as such). That is internal client behavior with no observable on-the-wire condition the rule framework can validate.
- **client-may-send-lower-version-for-broken-servers** (`rfc9110/client-may-send-lower-version-for-broken-servers`, `.type('informational')`): Informational (outcome 2): a permissive "MAY" about the HTTP protocol version a client sends. The HTTP version token lives on the request/status line and is not exposed by the rule framework (HttpRequest/HttpResponse carry no version field, and the common projection covers only header names, status, media type, and body). There is also no observable condition — it is a conditional allowance after a prior failed request.
- **client-must-not-send-non-conformant-version** (`rfc9110/client-must-not-send-non-conformant-version`, `.type('informational')`): Informational (outcome 2): a MUST NOT about the HTTP protocol version a client sends. The HTTP version token is transport control data not surfaced by the rule framework (no version field on HttpRequest; common projection is header names / status / media type / body only). Determining "conformance to a version" is also internal client state. Not observable.
- **client-should-send-highest-conformant-version** (`rfc9110/client-should-send-highest-conformant-version`, `.type('informational')`): Informational (outcome 2): a SHOULD about which HTTP protocol version a client sends. The version token is transport control data the rule framework does not expose, and "highest conformant version" depends on internal client and server capabilities. Not observable from a transaction.
- **recipient-should-process-higher-minor-version-as-highest-known** (`rfc9110/recipient-should-process-higher-minor-version-as-highest-known`, `.type('informational')`): Informational (outcome 2): a SHOULD about how a recipient internally processes a message whose minor version exceeds what it implements. This is internal recipient processing logic, and the HTTP version is transport control data not surfaced by the rule framework. Not observable.
- **server-must-not-send-non-conformant-version** (`rfc9110/server-must-not-send-non-conformant-version`, `.type('informational')`): Informational (outcome 2): a MUST NOT about the HTTP protocol version a server sends in its response. The response version token is transport control data not exposed by the rule framework (HttpResponse has no version field; the common projection is status / header names / media type / body / trailer names only), and "conformance to a version" is internal server state. Not observable.
- **server-should-send-response-version-equal-to-highest-conformant** (`rfc9110/server-should-send-response-version-equal-to-highest-conformant`, `.type('informational')`): Informational (outcome 2): a SHOULD about which HTTP protocol version a server puts in its response. The response version token is transport control data not surfaced by the rule framework, and "highest conformant version" is internal server capability. Not observable from a transaction.
- **each-http-version-defines-own-framing** (`rfc9110/each-http-version-defines-own-framing`, `.type('informational')`): Informational (outcome 2): a purely definitional statement (each HTTP version defines its own framing mechanism). It states no requirement on any message, so there is nothing to validate.
- **implicit-framing-allowed-for-backwards-compatibility** (`rfc9110/implicit-framing-allowed-for-backwards-compatibility`, `.type('informational')`): Informational (outcome 2): a permissive, definitional statement about connection-closure (implicit) framing. Framing is a transport/connection concern below the abstraction the rule framework observes, and the statement imposes no testable requirement.
- **message-complete-when-framed-octets-available** (`rfc9110/message-complete-when-framed-octets-available`, `.type('informational')`): Informational (outcome 2): a definitional statement of when a message is considered "complete" (a transport-level framing concept). It defines a term rather than imposing a requirement, and message completeness is below the abstraction the rule framework observes. Nothing to validate.
- **header-field-term-for-header-section-only-fields** (`rfc9110/header-field-term-for-header-section-only-fields`, `.type('informational')`): Informational (outcome 2): a terminology definition (what the spec means by "header field"). It imposes no requirement on any message, so there is nothing to validate.
- **header-fields-sent-before-content** (`rfc9110/header-fields-sent-before-content`, `.type('informational')`): Informational (outcome 2): a terminology definition ("header fields" are the fields sent before content). No requirement to validate.
- **header-section-consists-of-field-lines** (`rfc9110/header-section-consists-of-field-lines`, `.type('informational')`): Informational (outcome 2): a structural/terminology definition of the "header section". It states no requirement on any message, so there is nothing to validate.
- **origin-server-with-clock-must-generate-date-for-2xx-3xx-4xx** (`rfc9110/origin-server-with-clock-must-generate-date-for-2xx-3xx-4xx`): Implemented (outcome 1): a response-side server-behavior MUST that needs only the *presence* of the Date header on 2xx/3xx/4xx responses. Header-name presence is available in every context via the common projection, so a single shared `.rule()` runs identically at lint (described responses), test (live responses Thymian elicits), and analyze (recorded responses). The "with a clock" qualifier is not observable, but a missing Date on these status classes is the dominant, honest non-conformance signal. `appliesTo('origin server')` keeps it scoped to the responding origin and lets it fire on HAR (whose responses default to the `origin server` role). The multi-type union (static,analytics,test) infers the base ApiContext, so only the common projection is available — which is all this presence check requires.
- **origin-server-without-clock-must-not-generate-date** (`rfc9110/origin-server-without-clock-must-not-generate-date`, `.type('informational')`): Informational (outcome 2): conformance is conditioned on whether the origin server has a clock, which is internal server state not exposed in any transaction. Without knowing the server lacks a clock, a present (or absent) Date header cannot be classified as a violation, so there is no observable non-conformant condition.
- **recipient-may-replace-invalid-date** (`rfc9110/recipient-may-replace-invalid-date`, `.type('informational')`): Informational (outcome 2): permissive "MAY" describing internal recipient behavior (replacing an invalid Date value with the reception time). It is a local processing choice with no non-conformant condition observable in the transaction, so there is nothing to validate.
- **recipient-with-clock-must-add-date-if-missing** (`rfc9110/recipient-with-clock-must-add-date-if-missing`, `.type('informational')`): Informational (outcome 2): this MUST governs the internal behavior of a recipient with a clock (caches/forwards a response, adding Date if missing). It depends on internal recipient state (having a clock) and on what the recipient subsequently does downstream, neither of which is observable from a single recorded transaction, so there is no non-conformant condition to flag.
- **recipient-may-treat-the-set-of-received-trailer-fields-as-name-value-pairs** (`rfc9110/recipient-may-treat-the-set-of-received-trailer-fields-as-name-value-pairs`, `.type('informational')`): Informational (outcome 2): permissive "MAY" describing how a recipient MAY internally model received trailers. It is internal recipient processing with no observable non-conformant condition in any transaction.
- **recipient-must-not-merge-trailers-unsafely** (`rfc9110/recipient-must-not-merge-trailers-unsafely`, `.type('informational')`): Informational (outcome 2): this MUST NOT constrains an internal recipient decision — whether it merges a received trailer field into the header section. That merge happens inside the recipient/intermediary and is not visible in the transaction Thymian observes, so there is no non-conformant condition to detect. (Security-relevant: unsafe trailer-to-header merging is a request-smuggling / header-injection vector, but it remains an intermediary-internal behavior that the rule framework cannot observe.)
- **server-should-not-generate-necessary-trailers** (`rfc9110/server-should-not-generate-necessary-trailers`, `.type('informational')`): Informational (outcome 2): conformance turns on the server's belief that a trailer is "necessary for the user agent to receive". That intent is internal server state, not derivable from the wire, so no observable trailer pattern distinguishes a conformant from a non-conformant response. (The related, observable constraint — never putting *forbidden* fields in trailers — is enforced by `sender-must-not-generate-trailer-unless-permitted`.)
- **trailer-fields-must-be-defined-as-list-if-repeatable** (`rfc9110/trailer-fields-must-be-defined-as-list-if-repeatable`, `.type('informational')`): Informational (outcome 2): this MUST is a constraint on how a *field is defined* in its specification (repeatable trailer fields MUST be defined as list-based), not on the content of any individual message. There is no per-transaction observation — and no OpenAPI modeling of trailer-field definitions — that the rule framework can use to detect a violation.

packages/rules-rfc-9110/src/rules/message-abstraction/message-metadata/date/origin-server-may-generate-date-for-1xx-5xx.rule.ts
```diff
@@ -1,21 +1,12 @@
-import { and, not, or, responseHeader, statusCodeRange } from '@thymian/core';
 import { httpRule } from '@thymian/core';
 
 export default httpRule('rfc9110/origin-server-may-generate-date-for-1xx-5xx')
   .severity('hint')
-  .type('analytics')
+  .type('informational')
```
BaggersIO: Informational (outcome 2): this is a pure permissive "MAY". An origin server MAY generate a Date header field in 1xx and 5xx responses, but it is equally conformant to omit it. There is therefore no non-conformant condition to detect — neither presence nor absence of Date on a 1xx/5xx response is a violation. The previous `analytics` implementation flagged every 1xx/5xx response that lacked a Date header, which is exactly the conformant case and not a violation at all, so it has been reclassified to informational.

packages/rules-rfc-9110/src/rules/message-abstraction/message-metadata/date/user-agent-may-send-date-header-in-request.rule.ts
```diff
@@ -1,14 +1,12 @@
-import { not, requestHeader } from '@thymian/core';
 import { httpRule } from '@thymian/core';
 
 export default httpRule('rfc9110/user-agent-may-send-date-header-in-request')
   .severity('hint')
-  .type('analytics')
+  .type('informational')
```
BaggersIO: Informational (outcome 2): this is a request-side permissive "MAY". A user agent MAY send a Date header field in a request, and generally will not. Both sending and omitting Date are conformant, so there is no non-conformant condition to detect. The previous `analytics` implementation reported every request that lacked a Date header, which is the normal, conformant case and not a violation. It has therefore been reclassified to informational.

packages/rules-rfc-9110/src/rules/message-abstraction/message-metadata/date/sender-should-generate-date-at-message-generation.rule.ts
```diff
@@ -4,7 +4,7 @@ export default httpRule(
   'rfc9110/sender-should-generate-date-at-message-generation',
 )
   .severity('hint')
-  .type('analytics', 'informational')
+  .type('informational')
```
BaggersIO, line 13: Informational (outcome 4 — resolve mixed classification): this rule was declared `('analytics','informational')`, which mixes an active analyze context with the inert informational marker. The requirement is that the Date value approximate the moment of message generation. Verifying that would require knowing the true message-generation instant, which is not recoverable from a recorded transaction (the recorder only sees receipt time, and the RFC explicitly allows the sender to pick any instant during origination). There is no observable non-conformant condition, so the mixed type is resolved to a single coherent `informational` classification.

packages/rules-rfc-9110/src/rules/message-abstraction/message-metadata/trailer/sender-should-generate-trailer-header-when-sending-trailers.rule.ts
```diff
@@ -1,16 +1,12 @@
...
-  type RuleViolationLocation,
-} from '@thymian/core';
+import { not, responseHeader, type RuleViolationLocation } from '@thymian/core';
 import { httpRule } from '@thymian/core';
 
 export default httpRule(
   'rfc9110/sender-should-generate-trailer-header-when-sending-trailers',
 )
   .severity('warn')
   .type('analytics', 'test')
```
BaggersIO, line 15: Implemented (outcome 1): a response-side server-behavior SHOULD. The `analytics,test` union infers LiveApiContext, so the callback reads the live response and can inspect the actual trailer fields. A response that carries trailer fields but no Trailer header in its header section is the observable non-conformance. It is response-side (the server is the sender of the response trailers), so `test` is meaningful and `appliesTo('server','origin server')` lets it fire on HAR (responses default to the `origin server` role).

packages/rules-rfc-9110/src/rules/message-abstraction/trailer-fields/sender-must-not-generate-trailer-unless-permitted.rule.ts
```diff
@@ -23,6 +23,7 @@ export default httpRule(
 )
   .severity('error')
   .type('analytics')
```
BaggersIO, line 31: Implemented (outcome 1): a response-side MUST NOT that needs only the trailer field *names* (does a forbidden field appear in the trailers?). The single `analytics` type infers AnalyzeContext; `validateCommonHttpTransactions` exposes `res.trailers` as the trailer-name list (Object.keys of the trailer section), which is exactly what this check requires. Names are compared case-insensitively. Scoped to the response sender via `appliesTo('server','origin server')` so it fires on HAR responses, whose role defaults to `origin server`.

packages/rules-rfc-9110/src/rules/message-abstraction/message-metadata/date/origin-server-may-generate-date-for-1xx-5xx.rule.ts
matthyk: Let's keep this rule `analytics` and we can discuss if we should deactivate it in our recommended rule set. But this is out of scope for this PR, so keep the current type.

packages/rules-rfc-9110/src/rules/message-abstraction/message-metadata/date/user-agent-may-send-date-header-in-request.rule.ts
matthyk: Also keep this rule `analytics`

packages/rules-rfc-9110/src/rules/message-abstraction/message-metadata/trailer/sender-should-generate-trailer-header-when-sending-trailers.rule.ts
```diff
@@ -1,16 +1,19 @@
...
+// header in its header section is the observable non-conformance. It is
+// response-side (the server is the sender of the response trailers), so `test`
+// is meaningful and `appliesTo('server','origin server')` lets it fire on HAR
+// (responses default to the `origin server` role).
 export default httpRule(
   'rfc9110/sender-should-generate-trailer-header-when-sending-trailers',
 )
   .severity('warn')
   .type('analytics', 'test')
+  .appliesTo('server', 'origin server')
```
matthyk:
```suggestion
  .appliesTo('server')
```

packages/rules-rfc-9110/src/rules/message-abstraction/trailer-fields/sender-must-not-generate-trailer-unless-permitted.rule.ts
```diff
@@ -18,11 +18,19 @@ const FORBIDDEN_TRAILER_FIELDS = new Set([
...
+// trailer-name list (Object.keys of the trailer section), which is exactly what
+// this check requires. Names are compared case-insensitively. Scoped to the
+// response sender via `appliesTo('server','origin server')` so it fires on HAR
+// responses, whose role defaults to `origin server`.
 export default httpRule(
   'rfc9110/sender-must-not-generate-trailer-unless-permitted',
 )
   .severity('error')
   .type('analytics')
+  .appliesTo('server', 'origin server')
```
matthyk:
```suggestion
  .appliesTo('server')
```

packages/rules-rfc-9110/src/rules/message-abstraction/trailer-fields/sender-must-not-generate-trailer-unless-permitted.rule.ts
```diff
@@ -33,9 +41,11 @@ export default httpRule(
   .rule((ctx) =>
     ctx.validateCommonHttpTransactions(
       responseTrailer(),
```
matthyk:
```suggestion
or(...[...FORBIDDEN_TRAILER_FIELDS].map(responseTrailer)),
```

matthyk (changes requested):

matthyk (approved):

PR #303, opened 2026-06-29
feat(rules-rfc-9110): align fields rule execution contexts (#327)

BaggersIO:
## Align `fields` rule execution contexts (#327)

Brings every rule in `packages/rules-rfc-9110/src/rules/fields/` into line with what can *actually and meaningfully* execute, using the existing rule framework only. Each rule lands in exactly one of the four triage outcomes, and every `informational` / infrastructure-deferred rule now carries a written rationale in a code comment (replacing the old `// TODO` placeholders).

**30 rules reviewed.** Counts by outcome: **1 (implemented now): 5** · **2 (informational): 25** · **3 (infra-deferred analytics, no function): 0** · **4 (mis-typed reclassified, distinct from the above): 0**.

> **Update (2026-07-16, review round):** Both proxy rules are now implemented (outcome 1); no outcome-3 (declared active context without a rule function) rules remain. The five implemented rules are the two HTTP-date rules, the empty-list rule, and the two proxy rules. This round also corrected the proxy hop-direction handling (received vs. forwarded leg) with a regression test, added Retry-After coverage to the HTTP-date whitespace rule, and excluded Cache-Control from the empty-list check. Per reviewer feedback the HTTP-date/list `sender-*` rules were left **without** `appliesTo`, since "sender" binds any participant that emits the field, not only origin servers. Outcome-3 / "no fn" wording elsewhere below is superseded.

### Coverage matrix

| Rule | Before → After type | Outcome | Note |
|---|---|---|---|
| defining-field-values/date-time-formats/`sender-must-generate-timestamps-in-imf-fixdate-format` | informational → `test,analytics` | 1 | Implemented. Validates HTTP-date response header VALUES (Date/Expires/Last-Modified/Retry-After) against IMF-fixdate via `validateHttpTransactions`+`getHeader`. No `appliesTo` — the constraint binds any sender of an HTTP-date, not only origin servers. Request-side excluded from test (Thymian is the sender); lint excluded (no concrete timestamps in spec). |
| defining-field-values/date-time-formats/`sender-must-not-generate-additional-whitespace-in-http-date` | informational → `test,analytics` | 1 | Implemented. Flags HTAB / repeated SP / leading-trailing whitespace in HTTP-date response headers. Same context reasoning; no `appliesTo` (binds any HTTP-date sender). |
| defining-field-values/date-time-formats/`recipient-must-accept-all-http-date-formats` | informational → informational | 2 | Recipient-internal parser acceptance; no on-wire signal. |
| defining-field-values/date-time-formats/`recipient-must-interpret-two-digit-years-correctly` | informational → informational | 2 | Recipient-internal interpretation (50-year rule). |
| defining-field-values/lists/`recipient-must-accept-lists-with-empty-elements` | informational → informational | 2 | Recipient-internal parser acceptance. |
| defining-field-values/lists/`recipient-must-parse-and-ignore-empty-list-elements` | informational → informational | 2 | Recipient-internal parsing. |
| defining-field-values/lists/`sender-must-not-generate-empty-list-elements` | informational → `test,analytics` | 1 | Implemented. Flags empty list elements over a curated set of list-typed response headers (Vary, Allow, Content-Encoding, Accept-Ranges); Cache-Control excluded because its quoted-string parameters may contain commas. No `appliesTo` (binds any sender). |
| defining-field-values/quoted-strings/`recipient-must-handle-quoted-pairs-correctly` | informational → informational | 2 | Recipient-internal unquoting. |
| defining-field-values/quoted-strings/`sender-should-not-generate-quoted-pairs-except-for-dquote-backslash` | informational → informational | 2 | Needs per-field grammar to locate quoted-strings; generic scan unsound. |
| defining-field-values/quoted-strings/`sender-should-not-generate-quoted-pairs-in-comments-except-for-parens-backslash` | informational → informational | 2 | Needs per-field grammar to locate comments; generic scan unsound. |
| defining-field-values/whitespace/`implementation-may-remove-bws-before-processing` | informational → informational | 2 | Permissive MAY; also raw whitespace normalized before observation. |
| defining-field-values/whitespace/`implementation-may-replace-ows-or-rws-with-single-sp` | informational → informational | 2 | Permissive MAY; raw whitespace normalized. |
| defining-field-values/whitespace/`recipient-must-parse-and-remove-bws` | informational → informational | 2 | Recipient-internal parsing; raw octets normalized. |
| defining-field-values/whitespace/`sender-must-not-generate-bws` | informational → informational | 2 | Octet-level BWS not preserved through HTTP parsing / HAR. |
| defining-field-values/whitespace/`sender-should-generate-ows-as-single-sp-when-readable` | informational → informational | 2 | Octet-level OWS not preserved. |
| defining-field-values/whitespace/`sender-should-generate-rws-as-single-sp` | informational → informational | 2 | Octet-level RWS not preserved. |
| defining-field-values/whitespace/`sender-should-not-generate-ows-except-when-needed` | informational → informational | 2 | Octet-level OWS not preserved. |
| field-limits/`client-may-discard-oversized-field-lines` | informational → informational | 2 | Permissive MAY; internal client choice. |
| field-limits/`server-must-respond-4xx-for-oversized-fields` | informational → informational | 2 | Cannot craft oversized request; server limit unknown; 4xx not attributable to field size in HAR. |
| field-names/`proxy-must-forward-unrecognized-header-fields` | informational → `analytics` | 1 | Implemented over `validateCapturedHttpTraces`; correlates the proxy's received vs. forwarded leg and flags dropped fields (Connection-listed / hop-by-hop fields excepted; hops with uncaptured headers skipped). `appliesTo('proxy')`. Best-effort — see soundness caveat below. |
| field-names/`recipient-should-ignore-unrecognized-fields` | informational → informational | 2 | "Ignoring" is invisible from outside. |
| field-order/`proxy-must-not-change-field-line-order` | informational → `analytics` | 1 | Implemented over `validateCapturedHttpTraces`; compares same-name field-line order between the proxy's received and forwarded legs. `appliesTo('proxy')`. |
| field-order/`recipient-may-combine-field-lines-with-same-name` | informational → informational | 2 | Permissive MAY; internal recipient choice. |
| field-order/`sender-must-not-generate-multiple-field-lines-unless-allowed` | informational → informational | 2 | Duplicates collapsed by parser; also needs per-field grammar to know when allowed. |
| field-order/`server-must-not-apply-request-until-entire-header-received` | informational → informational | 2 | Server-internal processing-timing property; unobservable. |
| field-values/`new-fields-should-limit-values-to-visible-ascii` | informational → informational | 2 | SHOULD aimed at *spec authors* of new fields; design-time, no traffic condition. |
| field-values/`parser-must-exclude-whitespace-from-field-values` | informational → informational | 2 | Recipient-internal parsing; whitespace already trimmed before observation. |
| field-values/`recipient-may-retain-ctl-characters-in-safe-contexts` | informational → informational | 2 | Permissive MAY; "safe context" is internal app judgement. |
| field-values/`recipient-must-reject-or-replace-invalid-characters` | informational → informational | 2 | **Security-relevant** (CR/LF/NUL injection). Cannot transmit the invalid octets (client lib forbids them); conformant recipient leaves no trace. Pre-existing rationale expanded. |
| field-values/`recipient-should-treat-obs-text-as-opaque-data` | informational → informational | 2 | Recipient-internal handling; invisible from outside. |

### Security-relevant rules

- **`recipient-must-reject-or-replace-invalid-characters` (CR/LF/NUL injection)** — kept `informational` (severity `error`). Validating it would require transmitting a request whose field value carries a raw CR/LF/NUL, which Thymian's HTTP client cannot construct, and a conformant recipient that rejected/replaced the octet leaves no observable trace in recorded traffic. Rationale documented in code; intentionally not silently dropped.
- **`field-names/proxy-must-forward-unrecognized-header-fields`** and **`field-order/proxy-must-not-change-field-line-order`** — security-adjacent (header stripping / reordering at intermediaries). Implemented as `analytics` rules scoped `appliesTo('proxy')` over `validateCapturedHttpTraces`, correlating the proxy's received and forwarded legs. They stay `analytics`-only (not `test`) because the two-leg proxy correlation is observable only from proxy-side recordings, not from a single live transaction.
- **Date-family implementations** read only response header values; no request-side data is sent in `test`.

### Reclassification reasoning (non-trivial)

- **Two date-time `sender-*` rules moved informational → `test,analytics` and implemented.** The origin server is the sender of response timestamps, so the constraint is observable on the response side: present in the live response during `test` and in recorded responses during `analyze`. Lint is excluded because the OpenAPI description does not pin concrete timestamp values. Both read header VALUES, so they use `validateHttpTransactions` + `getHeader` (the common projection exposes header names only). They set **no** `appliesTo`: "sender" binds any participant that emits an HTTP-date, so scoping to `origin server` would wrongly exclude intermediary/cache responses.
- **Two proxy rules moved informational → `analytics` and implemented (outcome 1).** They govern intermediary forwarding behavior that requires correlating the two legs around a proxy; not reconstructable from a single transaction, but observable from proxy-side recordings. Both are implemented over `validateCapturedHttpTraces` with `appliesTo('proxy')`, comparing the proxy's received leg against its forwarded leg.
- All other rules were already `informational`; their `// TODO` placeholders were replaced with a written justification (recipient-internal behavior, permissive MAY, spec-author-directed SHOULD, octet-level whitespace lost to parsing/normalization, or per-field-grammar-registry dependence the framework does not provide).

### Per-PR checklist

- [x] Every rule in the category reviewed exactly once (30/30).
- [x] No declared active context (`static`/`test`/`analytics`) left without a rule function.
- [x] No request-side rule assigned to `test`.
- [x] Every `informational` and every infrastructure-deferred rule carries a written reason near `.type(...)`.
- [x] Header comparisons are case-insensitive (`getHeader` is case-insensitive; date-header lookups use lowercased names).
- [x] Header VALUES read via `validateHttpTransactions`+`getHeader`; the names-only common projection is not misused for value checks.
- [x] Where `appliesTo` is set it uses a concrete role (`'origin server'`, `'proxy'`), never bare `'server'`; the HTTP-date/list `sender-*` rules deliberately set no `appliesTo` so they bind any sender.
- [x] Security-relevant rules addressed explicitly (see section above).
- [x] Only files under `packages/rules-rfc-9110/src/rules/fields/**` changed; no framework/core/plugin/other-category edits.
- [x] `npx prettier --check` clean for the category; no new eslint errors (only the pre-existing package-wide `@nx/enforce-module-boundaries` errors remain).

This round closes the #327 triage gaps for the `fields` category. Every informational and infra-deferred rule now carries a written execution-context rationale adjacent to its `.type(...)`, and two rules were promoted from placeholder to real implementations.

**Implemented (was informational/infra-deferred, now has a fn):**
- `field-order/proxy-must-not-change-field-line-order` — implemented as an outcome-1 analytics rule over `validateCapturedHttpTraces`. On each proxy hop (`prev.request.meta.role === 'proxy'`) it compares the relative order of every same-name multi-line field between the received request (`prev.request`) and the forwarded request (`curr.request`); a reordering is a violation. Fields that were combined/split/dropped are skipped (out of scope for this rule). A rationale comment documents why this is only observable from proxy-recorded traffic.
- `defining-field-values/lists/sender-must-not-generate-empty-list-elements` — implemented (FIX-4) as `type('test','analytics')` reading real response values via `getHeader` over a curated whitelist of list-typed response headers (`Vary`, `Allow`, `Content-Encoding`, `Accept-Ranges`), flagging leading/trailing/doubled commas. Cache-Control is excluded because its quoted-string parameters may contain commas. No `appliesTo` — the constraint binds any sender.

**Implemented, with a documented soundness caveat:**
- `field-names/proxy-must-forward-unrecognized-header-fields` — implemented as `type('analytics')` + `appliesTo('proxy')` over `validateCapturedHttpTraces`, flagging fields present on the received leg but absent from the forwarded leg (Connection-listed / hop-by-hop fields excepted; hops with uncaptured headers are skipped to avoid false positives). Caveat documented in code: from traffic alone a dropped field cannot be *proven* a violation — "unrecognized" is indistinguishable from a field the proxy is "specifically configured to block/transform" (an allowed exception), and Connection-listed/hop-by-hop fields are legitimately removed — so this is a best-effort signal, kept `analytics`-only.

**Restored/expanded unobservable rationale (outcome 2):**
- `field-values/recipient-must-reject-or-replace-invalid-characters` — restored + expanded the previously-deleted rationale: the HTTP client cannot transmit raw CR/LF/NUL in a field value, and a conformant rejecting/replacing recipient leaves no observable trace.
- `field-limits/server-must-respond-4xx-for-oversized-fields` — added outcome-2 rationale and corrected `appliesTo('server')` → `appliesTo('origin server')` (FIX-5).

**Systematic rationale (FIX-3, Theme C):** added concise execution-context rationale comments near `.type(...)` to every remaining informational rule in `fields/**` whose `// TODO` was removed without a replacement — covering the date-time recipient rules, list recipient rules, quoted-string rules, all whitespace rules, and the remaining field-names/field-order/field-values informational rules.

Build gate: `tsc -b packages/rules-rfc-9110/tsconfig.lib.json` passes (exit 0).


packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/recipient-must-accept-all-http-date-formats.rule.ts
```diff
@@ -1,9 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for HTTP-date parsing
-// Requires parsing all three HTTP-date formats:
-//   HTTP-date = IMF-fixdate / obs-date
-//   obs-date = rfc850-date / asctime-date
 export default httpRule('rfc9110/recipient-must-accept-all-http-date-formats')
   .severity('error')
   .type('informational')
```
BaggersIO, line 8: Informational: this MUST constrains the recipient's internal parser — it must accept all three HTTP-date formats (IMF-fixdate, rfc850-date, asctime-date). Whether a recipient accepts a given format is an internal implementation property that produces no observable signal in the request/response that Thymian can lint, test, or analyze. There is no conformant-vs-nonconformant difference visible on the wire, so the rule is recorded for documentation only.

packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/recipient-must-interpret-two-digit-years-correctly.rule.ts
```diff
@@ -1,7 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation and logic for two-digit year interpretation
-// Requires parsing rfc850-date format and applying 50-year rule for 2-digit years
 export default httpRule(
   'rfc9110/recipient-must-interpret-two-digit-years-correctly',
 )
```
BaggersIO, line 5: Informational: this MUST governs how the recipient interprets a two-digit rfc850-date year (the 50-year sliding-window rule). It is an internal interpretation decision inside the recipient; it does not alter the bytes of any message, so there is nothing observable for lint, test, or analyze to validate. Recorded for documentation only.

(Anchored to line 5; the rule's `.type(` is at line 7 but that line is not within this PR's diff hunk after the comment removal.)

packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-generate-timestamps-in-imf-fixdate-format.rule.ts
```diff
@@ -1,20 +1,129 @@
...
+function isDelaySeconds(value: string): boolean {
+  return /^\d+$/.test(value.trim());
+}
+
 export default httpRule(
   'rfc9110/sender-must-generate-timestamps-in-imf-fixdate-format',
 )
   .severity('error')
-  .type('informational')
+  .type('test', 'analytics')
```
BaggersIO, line 74: Sender-side rule made observable on the response side: the origin server is the sender of these response timestamps. During `test` Thymian sends the request, so only the server's response timestamps can be observed; during `analyze` the same response timestamps are present in recorded traffic. (Lint is excluded: the OpenAPI description does not pin concrete timestamp values, so there is nothing to validate at design time.) The check reads header VALUES, so it uses validateHttpTransactions + getHeader rather than the common projection (which exposes header names only).

packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-not-generate-additional-whitespace-in-http-date.rule.ts
```diff
@@ -1,19 +1,84 @@
...
+function hasAdditionalWhitespace(value: string): boolean {
+  return /\t/.test(value) || / {2,}/.test(value) || value !== value.trim();
+}
+
 export default httpRule(
   'rfc9110/sender-must-not-generate-additional-whitespace-in-http-date',
 )
   .severity('error')
-  .type('informational')
+  .type('test', 'analytics')
```
BaggersIO, line 56: Sender-side rule observed on the response side: the origin server is the sender of these response timestamps. During `test` only the server's response is observable (Thymian sends the request); during `analyze` the response is present in recorded traffic. Lint is excluded because the OpenAPI description does not pin concrete timestamp values. The check needs header VALUES, so it uses validateHttpTransactions + getHeader rather than the names-only common projection.

BaggersIO (commented):
Recording the execution-context triage rationale here (moved out of inline code comments per issue #327).

The following rules' only PR change was the triage comment; removing it reverted the file to `main`, so they are not in the diff and cannot be line-anchored. Their rationale is recorded here:

- **field-limits/client-may-discard-oversized-field-lines.rule.ts** (`.type` line 5): Informational: a permissive MAY describing an internal client choice (discard/truncate oversized field lines it does not wish to process) with no non-conformant condition. Whether the client discards a value is an internal decision with no observable signal in any message Thymian can lint, test, or analyze. Recorded for documentation only.
- **field-limits/server-must-respond-4xx-for-oversized-fields.rule.ts** (`.type` line 5): Informational: validating this MUST requires sending a request whose header field line / value / set exceeds the server's limit and observing a 4xx, but the response depends on the server's (unknown, deployment- and resource-specific) limit, and Thymian's HTTP stack does not let a rule craft a deliberately oversized request to provoke it. In recorded traffic the over-limit request that triggered a 4xx is rarely present, and a 4xx alone is not attributable to field size. No reliable observable condition exists, so it is recorded for documentation only.
- **field-names/recipient-should-ignore-unrecognized-fields.rule.ts** (`.type` line 5): Informational: this SHOULD describes an internal recipient behavior (silently ignore unrecognized header/trailer fields). "Ignoring" produces no observable signal — a conformant recipient looks identical to one that never received the field — so there is nothing for lint, test, or analyze to validate. Recorded for documentation only.
- **field-order/recipient-may-combine-field-lines-with-same-name.rule.ts** (`.type` line 7): Informational: a permissive MAY describing an internal recipient choice (combine same-name field lines into one comma-separated value). There is no non-conformant condition to detect, and the choice has no observable effect Thymian can lint, test, or analyze. Recorded for documentation only.
- **field-order/sender-must-not-generate-multiple-field-lines-unless-allowed.rule.ts** (`.type` line 7): Informational: detecting this MUST has two blockers. (1) Observability: HTTP libraries parse repeated same-name field lines into a single combined value (or array), so the fact that a sender emitted multiple lines is not reliably preserved in the parsed messages Thymian sees, nor in HAR. (2) Even if duplicates were visible, the rule only forbids them when the field's definition does NOT allow a comma-separated #(list) recombination — deciding that requires a per-field grammar registry the framework does not provide. Honestly deferred to informational. Recorded for documentation only.
- **field-order/server-must-not-apply-request-until-entire-header-received.rule.ts** (`.type` line 7): Informational: this MUST constrains the server's internal processing timing — it must not act on a request before the full header section is received. This is an internal ordering/timing property of the server with no observable signal in the completed request/response that Thymian can lint, test, or analyze. Recorded for documentation only.

packages/rules-rfc-9110/src/rules/fields/defining-field-values/lists/recipient-must-accept-lists-with-empty-elements.rule.ts
```diff
@@ -1,8 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for #rule recipient syntax
-// Requires accepting lists with empty elements per recipient syntax:
-//   #element => [ element ] *( OWS "," OWS [ element ] )
 export default httpRule(
   'rfc9110/recipient-must-accept-lists-with-empty-elements',
 )
```
BaggersIO, line 5: Informational: this MUST constrains the recipient's internal list parser (it must not reject lists that carry empty elements). Acceptance is an internal parser behavior with no observable effect on the request/response Thymian can inspect — there is no on-the-wire signal distinguishing a conformant from a non-conformant recipient. Recorded for documentation only.

(Anchored to line 5; the rule's `.type(` is at line 7 but that line is not within this PR's diff hunk after the comment removal.)

packages/rules-rfc-9110/src/rules/fields/defining-field-values/lists/recipient-must-parse-and-ignore-empty-list-elements.rule.ts
```diff
@@ -1,8 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for #rule list parsing with empty elements
-// Requires parsing recipient-side #rule syntax and filtering empty elements:
-//   #element => [ element ] *( OWS "," OWS [ element ] )
 export default httpRule(
   'rfc9110/recipient-must-parse-and-ignore-empty-list-elements',
 )
```
BaggersIO, line 5: Informational: this MUST governs the recipient's internal parsing (ignore a reasonable number of empty list elements). It is an internal behavior of the recipient that emits no observable difference in the messages Thymian can lint, test, or analyze. Recorded for documentation only.

(Anchored to line 5; the rule's `.type(` is at line 7 but that line is not within this PR's diff hunk after the comment removal.)

packages/rules-rfc-9110/src/rules/fields/defining-field-values/lists/sender-must-not-generate-empty-list-elements.rule.ts
```diff
@@ -1,9 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for #rule list syntax
-// Requires parsing #rule lists and detecting empty elements:
-//   1#element => element *( OWS "," OWS element )
-// Can be implemented in static context to validate outgoing messages
 export default httpRule('rfc9110/sender-must-not-generate-empty-list-elements')
   .severity('error')
   .type('informational')
```
BaggersIO: Informational: detecting an empty list element (e.g. a stray "a,,b") requires first knowing that a given field uses the #(list) production. Whether a comma is a list separator is field-specific: in Date, Cookie, ETag with weak validators, and many other fields a comma is ordinary data, not a list delimiter. The framework provides no per-field grammar registry to distinguish list-valued fields from comma-bearing single values, so a generic check across captured header values would produce false positives on legitimate traffic. Honestly deferred to informational rather than shipping an unsound check. Recorded for documentation only.

packages/rules-rfc-9110/src/rules/fields/defining-field-values/quoted-strings/recipient-must-handle-quoted-pairs-correctly.rule.ts
```diff
@@ -1,9 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for quoted-pair parsing
-// Requires parsing quoted-string and quoted-pair:
-//   quoted-string = DQUOTE *( qdtext / quoted-pair ) DQUOTE
-//   quoted-pair = "\" ( HTAB / SP / VCHAR / obs-text )
 export default httpRule('rfc9110/recipient-must-handle-quoted-pairs-correctly')
   .severity('error')
   .type('informational')
```
BaggersIO, line 8: Informational: this MUST describes how the recipient must internally interpret a quoted-pair (treat "\x" as the octet x). It is an internal unquoting behavior of the recipient and produces no observable difference in any message Thymian can lint, test, or analyze. Recorded for documentation only.

packages/rules-rfc-9110/src/rules/fields/defining-field-values/quoted-strings/sender-should-not-generate-quoted-pairs-except-for-dquote-backslash.rule.ts
```diff
@@ -1,8 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for unnecessary quoted-pair detection
-// Requires detecting quoted-pairs not needed for DQUOTE (%x22) or backslash (%x5C)
-// Can be implemented in static context to validate outgoing messages
 export default httpRule(
   'rfc9110/sender-should-not-generate-quoted-pairs-except-for-dquote-backslash',
 )
```
BaggersIO, line 5: Informational: flagging an "unnecessary" quoted-pair requires reliably locating quoted-strings inside arbitrary field values, which in turn requires per-field grammar knowledge (only some fields define quoted-string syntax, and a backslash is ordinary data elsewhere). The framework exposes no per-field grammar registry, so a generic scan would misfire on legitimate values. Honestly deferred to informational rather than shipping an unsound check. Recorded for documentation only.

(Anchored to line 5; the rule's `.type(` is at line 7 but that line is not within this PR's diff hunk after the comment removal.)

packages/rules-rfc-9110/src/rules/fields/defining-field-values/quoted-strings/sender-should-not-generate-quoted-pairs-in-comments-except-for-parens-backslash.rule.ts
```diff
@@ -1,9 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for comment syntax and quoted-pair detection
-// Requires parsing comment and detecting unnecessary quoted-pairs:
-//   comment = "(" *( ctext / quoted-pair / comment ) ")"
-// Can be implemented in static context to validate outgoing messages
 export default httpRule(
   'rfc9110/sender-should-not-generate-quoted-pairs-in-comments-except-for-parens-backslash',
 )
```
BaggersIO, line 5: Informational: detecting an unnecessary quoted-pair inside a comment requires locating comments — a construct only some fields (e.g. User-Agent, Server, Via) permit — and parsing their nested parenthesis grammar. The framework provides no per-field grammar registry to know where comments are syntactically valid, so a generic scan would misfire. Honestly deferred to informational rather than shipping an unsound check. Recorded for documentation only.

(Anchored to line 5; the rule's `.type(` is at line 7 but that line is not within this PR's diff hunk after the comment removal.)

packages/rules-rfc-9110/src/rules/fields/defining-field-values/whitespace/implementation-may-remove-bws-before-processing.rule.ts
```diff
@@ -1,7 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for BWS detection
-// Requires identifying BWS positions in grammar for optional removal
 export default httpRule(
   'rfc9110/implementation-may-remove-bws-before-processing',
 )
```
BaggersIO, line 5: Informational: a permissive MAY (an implementation may remove BWS before processing/forwarding) with no non-conformant condition — there is nothing to flag. It also concerns the recipient's internal handling of raw whitespace, which is normalized away by the HTTP layer before Thymian sees parsed header values. Recorded for documentation only.

(Anchored to line 5; the rule's `.type(` is at line 7 but that line is not within this PR's diff hunk after the comment removal.)

packages/rules-rfc-9110/src/rules/fields/defining-field-values/whitespace/implementation-may-replace-ows-or-rws-with-single-sp.rule.ts
```diff
@@ -1,7 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for OWS/RWS normalization
-// Requires detecting OWS/RWS positions and optionally normalizing to single SP
 export default httpRule(
   'rfc9110/implementation-may-replace-ows-or-rws-with-single-sp',
 )
```
BaggersIO, line 5: Informational: a permissive MAY (OWS/RWS may be replaced with a single SP before processing/forwarding) with no non-conformant condition to detect. It also concerns raw inter-token whitespace, which the HTTP layer normalizes before Thymian observes parsed header values. Recorded for documentation only.

(Anchored to line 5; the rule's `.type(` is at line 7 but that line is not within this PR's diff hunk after the comment removal.)

packages/rules-rfc-9110/src/rules/fields/defining-field-values/whitespace/recipient-must-parse-and-remove-bws.rule.ts
```diff
@@ -1,8 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for BWS parsing and removal
-// Requires detecting and removing BWS before processing:
-//   BWS = OWS ; "bad" whitespace
 export default httpRule('rfc9110/recipient-must-parse-and-remove-bws')
   .severity('error')
   .type('informational')
```
BaggersIO, line 8: Informational: this MUST governs the recipient's internal parsing (it must strip bad whitespace before interpreting an element). It is an internal recipient behavior with no observable effect on the messages Thymian can lint, test, or analyze, and the relevant raw octets are normalized by the HTTP layer before Thymian sees parsed values. Recorded for documentation only.

packages/rules-rfc-9110/src/rules/fields/defining-field-values/whitespace/sender-must-not-generate-bws.rule.ts
```diff
@@ -1,10 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for BWS detection
-// Requires parsing BWS (bad whitespace) = OWS in historically allowed positions
-//   BWS = OWS ; "bad" whitespace
-//   OWS = *( SP / HTAB )
-// Can be implemented in static context to validate outgoing messages
 export default httpRule('rfc9110/sender-must-not-generate-bws')
   .severity('error')
   .type('informational')
```
BaggersIO, line 10: Informational: BWS is raw whitespace at specific grammar positions on the wire. By the time Thymian observes a message the HTTP layer has parsed and normalized field values, so the original octet-level BWS the sender did or did not generate is no longer present to inspect; nor does it survive into a HAR's parsed headers. With no preserved on-the-wire representation there is nothing observable to flag. Recorded for documentation only.

packages/rules-rfc-9110/src/rules/fields/defining-field-values/whitespace/sender-should-generate-ows-as-single-sp-when-readable.rule.ts
```diff
@@ -1,9 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for OWS generation
-// Requires detecting OWS positions and validating single SP generation:
-//   OWS = *( SP / HTAB ) ; optional whitespace
-// Can be implemented in static context to validate outgoing messages
 export default httpRule(
   'rfc9110/sender-should-generate-ows-as-single-sp-when-readable',
 )
```
BaggersIO, line 5: Informational: OWS is raw optional whitespace between tokens on the wire. The HTTP layer parses and normalizes field values before Thymian observes them, so the sender's original OWS representation (single SP vs. multiple SP/HTAB) is not preserved for lint, test, or analyze; HAR parsed headers likewise do not carry it. Nothing observable to flag. Recorded for documentation only.

(Anchored to line 5; the rule's `.type(` is at line 7 but that line is not within this PR's diff hunk after the comment removal.)

packages/rules-rfc-9110/src/rules/fields/defining-field-values/whitespace/sender-should-generate-rws-as-single-sp.rule.ts
```diff
@@ -1,9 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for RWS generation
-// Requires detecting RWS positions and validating single SP generation:
-//   RWS = 1*( SP / HTAB ) ; required whitespace
-// Can be implemented in static context to validate outgoing messages
 export default httpRule('rfc9110/sender-should-generate-rws-as-single-sp')
   .severity('warn')
   .type('informational')
```
BaggersIO, line 8: Informational: RWS is raw required whitespace between tokens on the wire. The HTTP layer normalizes field values before Thymian observes them, so the sender's original RWS representation (single SP vs. a longer run of SP/HTAB) is not preserved for lint, test, or analyze, nor in HAR parsed headers. Nothing observable to flag. Recorded for documentation only.

packages/rules-rfc-9110/src/rules/fields/defining-field-values/whitespace/sender-should-not-generate-ows-except-when-needed.rule.ts
```diff
@@ -1,8 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for unnecessary OWS detection
-// Requires detecting OWS where not needed for readability
-// Can be implemented in static context to validate outgoing messages
 export default httpRule(
   'rfc9110/sender-should-not-generate-ows-except-when-needed',
 )
```
BaggersIO, line 5: Informational: this SHOULD constrains where a sender places raw optional whitespace on the wire. That octet-level OWS is normalized by the HTTP layer before Thymian observes parsed field values, and does not survive into HAR parsed headers, so there is no preserved representation to flag for lint, test, or analyze. Recorded for documentation only.

(Anchored to line 5; the rule's `.type(` is at line 7 but that line is not within this PR's diff hunk after the comment removal.)

packages/rules-rfc-9110/src/rules/fields/field-names/proxy-must-forward-unrecognized-header-fields.rule.ts
```diff
@@ -2,7 +2,7 @@ import { httpRule } from '@thymian/core';
 
 export default httpRule('rfc9110/proxy-must-forward-unrecognized-header-fields')
   .severity('error')
-  .type('informational')
+  .type('analytics')
```
BaggersIO, line 26: Infrastructure-deferred (outcome 3): this MUST governs a proxy's forwarding behavior. Detecting a violation requires correlating the message a proxy received with the message it forwarded and observing that an unrecognized field present on the inbound side was dropped on the outbound side. That before/after-the-proxy linkage is not available from a single captured transaction; it depends on traffic recorded at the proxy (both legs). The rule therefore stays typed `analytics` and scoped to the `proxy` role rather than being downgraded to informational, since it is observable from suitable proxy-side recordings even though no function can be written against a single transaction today.

packages/rules-rfc-9110/src/rules/fields/field-order/proxy-must-not-change-field-line-order.rule.ts
```diff
@@ -2,7 +2,7 @@ import { httpRule } from '@thymian/core';
 
 export default httpRule('rfc9110/proxy-must-not-change-field-line-order')
   .severity('error')
-  .type('informational')
+  .type('analytics')
```
BaggersIO, line 23: Infrastructure-deferred (outcome 3): this MUST governs how a proxy forwards repeated same-name field lines. Detecting a violation requires comparing the order of those field-line values as received by the proxy with the order it emits when forwarding — a before/after-the-proxy comparison that is not reconstructable from a single captured transaction (and HTTP libraries typically collapse repeated field lines into one parsed value, discarding ordering). It is observable only from traffic recorded at the proxy that preserves both legs and the per-line order, so the rule stays typed `analytics` and scoped to the `proxy` role rather than being downgraded to informational.

packages/rules-rfc-9110/src/rules/fields/field-values/new-fields-should-limit-values-to-visible-ascii.rule.ts
```diff
@@ -1,8 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for VCHAR character detection
-// Requires validating field values only contain VCHAR (%x21-7E), SP (%x20), and HTAB (%x09)
-// Can be implemented in static context to validate outgoing field values
 export default httpRule(
   'rfc9110/new-fields-should-limit-values-to-visible-ascii',
 )
```
BaggersIO, line 5: Informational: this SHOULD is directed at *specification authors* defining new header fields ("Specifications for newly defined fields SHOULD limit their values to ..."), not at any HTTP message participant. It is a design-time recommendation about future field definitions; it has no observable condition in the request/response traffic or OpenAPI description that Thymian can lint, test, or analyze. Recorded for documentation only.

(Anchored to line 5; the rule's `.type(` is at line 7 but that line is not within this PR's diff hunk after the comment removal.)

packages/rules-rfc-9110/src/rules/fields/field-values/parser-must-exclude-whitespace-from-field-values.rule.ts
```diff
@@ -1,9 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for field-value format (field-content, field-vchar)
-// Requires parsing field-value ABNF to detect and validate whitespace handling:
-//   field-value = *field-content
-//   field-content = field-vchar [ 1*( SP / HTAB / field-vchar ) field-vchar ]
 export default httpRule(
   'rfc9110/parser-must-exclude-whitespace-from-field-values',
 )
```
BaggersIO, line 5: Informational: this MUST governs the recipient's internal field parser (it must strip leading/trailing whitespace before evaluating a value). It is an internal parsing behavior with no observable signal in the messages Thymian can lint, test, or analyze, and the HTTP layer has already trimmed such whitespace before Thymian sees parsed values. Recorded for documentation only.

(Anchored to line 5; the rule's `.type(` is at line 7 but that line is not within this PR's diff hunk after the comment removal.)

packages/rules-rfc-9110/src/rules/fields/field-values/recipient-may-retain-ctl-characters-in-safe-contexts.rule.ts
```diff
@@ -1,8 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for CTL character detection
-// Requires detecting CTL characters (%x00-1F, %x7F) excluding CR/LF/NUL
-// Context-aware validation needed to determine "safe contexts"
 export default httpRule(
   'rfc9110/recipient-may-retain-ctl-characters-in-safe-contexts',
 )
```
BaggersIO, line 5: Informational: a permissive MAY allowing the recipient to retain certain CTL characters "for the sake of robustness" within a safe context. There is no non-conformant condition to flag, and "safe context" is an application-specific judgement internal to the recipient — not observable in the messages Thymian can lint, test, or analyze. Recorded for documentation only.

(Anchored to line 5; the rule's `.type(` is at line 7 but that line is not within this PR's diff hunk after the comment removal.)

packages/rules-rfc-9110/src/rules/fields/field-values/recipient-must-reject-or-replace-invalid-characters.rule.ts
```diff
@@ -1,14 +1,9 @@
...
-// TODO: Implement ABNF validation for invalid character detection
-// Requires detecting CR (%x0D), LF (%x0A), or NUL (%x00) in field values
-// Can be implemented in analytics context to validate message handling
 export default httpRule(
   'rfc9110/recipient-must-reject-or-replace-invalid-characters',
 )
   .severity('error')
-  // We need to make this informational for now since we do not yet have the functionality to craft or send an intentionally invalid HTTP request.
-  // To test this rule against a real HTTP server, we would have to establish a TCP connection ourselves and manually construct the HTTP request.
   .type('informational')
```
BaggersIO, line 14: Informational (security-relevant — CR/LF/NUL injection): validating this MUST requires sending a request whose field value contains a raw CR, LF, or NUL and observing that the recipient rejects the message or replaces the octet with SP. Thymian's HTTP stack cannot craft and transmit such a deliberately invalid request (the client library forbids these octets in header values), so the trigger condition cannot be produced in test; and a conformant recipient that has already rejected/replaced the octet leaves no observable trace of the original invalid input in recorded traffic. With no way to generate the input and no preserved signal to analyze, it is recorded for documentation only.

packages/rules-rfc-9110/src/rules/fields/field-values/recipient-should-treat-obs-text-as-opaque-data.rule.ts
```diff
@@ -1,8 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// TODO: Implement ABNF validation for obs-text detection
-// Requires detecting obs-text octets (%x80-FF) in field values
-//   obs-text = %x80-FF
 export default httpRule(
   'rfc9110/recipient-should-treat-obs-text-as-opaque-data',
 )
```
BaggersIO, line 5: Informational: this SHOULD describes how the recipient internally treats obs-text octets (as opaque data). It is an internal handling decision with no observable effect in the messages Thymian can lint, test, or analyze — a conformant and a non-conformant recipient are indistinguishable from the outside. Recorded for documentation only.

(Anchored to line 5; the rule's `.type(` is at line 7 but that line is not within this PR's diff hunk after the comment removal.)

BaggersIO:
Implemented `proxy-must-forward-unrecognized-header-fields`, which was typed `analytics` without a rule function. It correlates the received/forwarded proxy legs in a `CapturedTrace` and flags end-to-end fields dropped in forwarding, excluding hop-by-hop and Connection-listed fields. Severity lowered `error`→`warn` because an intentional block is an allowed exception (so a dropped field is not provably a violation). Build green; runtime-checked. Commit e415e1c6.

packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-generate-timestamps-in-imf-fixdate-format.rule.ts
```diff
@@ -1,20 +1,129 @@
...
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.7')
   .description(
     'When a sender generates a field that contains one or more timestamps defined as HTTP-date, the sender MUST generate those timestamps in the IMF-fixdate format.',
   )
   .summary(
     'Sender MUST generate HTTP-date timestamps in the IMF-fixdate format.',
   )
+  .appliesTo('origin server')
```
matthyk:
```suggestion
```

packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-generate-timestamps-in-imf-fixdate-format.rule.ts
```diff
@@ -1,20 +1,129 @@
...
+          return [];
+        }
+
+        return [
+          {
+            location,
+            violation: {
+              message: `The response carries HTTP-date timestamp(s) not in the required IMF-fixdate format (e.g. "Sun, 06 Nov 1994 08:49:37 GMT"). Offending field value(s): ${createList(
+                malformed,
+              )}. A sender MUST generate HTTP-date timestamps in IMF-fixdate; the obsolete rfc850 and asctime formats MUST NOT be generated.`,
```
matthyk:
```suggestion
              message: `The response carries HTTP-date timestamp(s) not in the required IMF-fixdate format: ${createList(
                malformed,
              )}. `,
```

packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-not-generate-additional-whitespace-in-http-date.rule.ts
```diff
@@ -1,19 +1,84 @@
...
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.7')
   .description(
     'A sender MUST NOT generate additional whitespace in an HTTP-date beyond that specifically included as SP in the grammar.',
   )
   .summary(
     'Sender MUST NOT generate additional whitespace in HTTP-date beyond the grammar.',
   )
+  .appliesTo('origin server')
   .tags('fields', 'date-time', 'whitespace')
```
matthyk:
```suggestion
```

packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-not-generate-additional-whitespace-in-http-date.rule.ts
```diff
@@ -1,19 +1,84 @@
...
+          return [];
+        }
+
+        return [
+          {
+            location,
+            violation: {
+              message: `The response carries HTTP-date field value(s) with whitespace beyond the single SP characters defined by the IMF-fixdate grammar (no HTAB, no repeated spaces, no leading/trailing whitespace are permitted). Offending field value(s): ${createList(
+                offending,
+              )}.`,
```
matthyk:
```suggestion
              message: `The response carries HTTP-date field value(s) with additional whitespace: ${createList(
                offending,
              )}.`,
```

packages/rules-rfc-9110/src/rules/fields/defining-field-values/lists/sender-must-not-generate-empty-list-elements.rule.ts
```diff
@@ -1,16 +1,89 @@
...
 export default httpRule('rfc9110/sender-must-not-generate-empty-list-elements')
   .severity('error')
-  .type('informational')
+  .type('test', 'analytics')
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.1')
   .description(
     'In any production that uses the list construct, a sender MUST NOT generate empty list elements.',
   )
   .summary('Sender MUST NOT generate empty list elements.')
+  .appliesTo('origin server')
```
matthyk:
```suggestion
```

packages/rules-rfc-9110/src/rules/fields/defining-field-values/lists/sender-must-not-generate-empty-list-elements.rule.ts
```diff
@@ -1,16 +1,89 @@
...
+ * because only these are guaranteed list-typed; scanning arbitrary fields for
+ * commas would misfire on values (e.g. dates) that legitimately contain them.
+ */
+const listTypedResponseHeaders = [
+  'vary',
+  'allow',
+  'content-encoding',
+  'cache-control',
+  'accept-ranges',
+];
```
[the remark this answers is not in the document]
matthyk: +1

packages/rules-rfc-9110/src/rules/fields/defining-field-values/lists/sender-must-not-generate-empty-list-elements.rule.ts
```diff
@@ -1,16 +1,89 @@
...
+          return [];
+        }
+
+        return [
+          {
+            location,
+            violation: {
+              message: `The response carries a list-typed field with empty list element(s) (a leading/trailing comma or a doubled comma). A sender MUST NOT generate empty list elements. Offending field value(s): ${createList(
+                offending,
+              )}.`,
```
matthyk:
```suggestion
              message: `The response carries a list-typed field with empty list element(s): ${createList(
                offending,
              )}.`,
```

packages/rules-rfc-9110/src/rules/fields/field-limits/server-must-respond-4xx-for-oversized-fields.rule.ts
```diff
@@ -10,6 +16,6 @@ export default httpRule('rfc9110/server-must-respond-4xx-for-oversized-fields')
   .summary(
     'Server MUST respond with 4xx status code when receiving oversized fields.',
   )
-  .appliesTo('server')
+  .appliesTo('origin server')
```
matthyk:
```suggestion
  .appliesTo('server')
```

packages/rules-rfc-9110/src/rules/fields/field-names/proxy-must-forward-unrecognized-header-fields.rule.ts
```diff
@@ -1,13 +1,72 @@
...
+const connectionOptionNames = (value: string | string[] | undefined): string[] =>
+  (value === undefined ? [] : Array.isArray(value) ? value : [value])
+    .flatMap((entry) => entry.split(','))
+    .map((entry) => entry.trim().toLowerCase())
+    .filter((entry) => entry.length > 0);
 
 export default httpRule('rfc9110/proxy-must-forward-unrecognized-header-fields')
-  .severity('error')
-  .type('informational')
+  .severity('warn')
```
matthyk:
```suggestion
  .severity('error')
```

packages/rules-rfc-9110/src/rules/fields/field-names/proxy-must-forward-unrecognized-header-fields.rule.ts
```diff
@@ -1,13 +1,72 @@
...
+  .type('analytics')
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1')
   .description(
     'A proxy MUST forward unrecognized header fields unless the field name is listed in the Connection header field (Section 7.6.1) or the proxy is specifically configured to block, or otherwise transform, such fields.',
   )
   .summary('Proxy MUST forward unrecognized header fields unless excepted.')
   .appliesTo('proxy')
   .tags('fields', 'field-names', 'proxy')
```
matthyk:
```suggestion
```

packages/rules-rfc-9110/src/rules/fields/field-names/proxy-must-forward-unrecognized-header-fields.rule.ts
```diff
@@ -1,13 +1,72 @@
...
+          .filter(
+            (name) =>
+              !excepted.has(name) &&
+              getHeader(forwarded.request.data.headers, name) === undefined,
+          );
+        if (dropped.length > 0) {
+          results.push({
+            location,
+            violation: {
+              message: `A proxy did not forward header field(s) (${dropped.join(', ')}) present on the received request. Unless configured to block them, a proxy must forward unrecognized fields.`,
```
matthyk:
```suggestion
              message: `A proxy did not forward header field(s) (${dropped.join(', ')}) present on the received request.`,
```

packages/rules-rfc-9110/src/rules/fields/field-order/proxy-must-not-change-field-line-order.rule.ts
```diff
@@ -12,4 +30,68 @@ export default httpRule('rfc9110/proxy-must-not-change-field-line-order')
   )
   .appliesTo('proxy')
   .tags('fields', 'field-order', 'proxy')
```
matthyk:
```suggestion
```

packages/rules-rfc-9110/src/rules/fields/field-order/proxy-must-not-change-field-line-order.rule.ts
```diff
@@ -12,4 +30,68 @@ export default httpRule('rfc9110/proxy-must-not-change-field-line-order')
...
+          if (before.some((value, idx) => value !== after[idx])) {
+            changed.push(name);
+          }
+        }
+
+        for (const name of changed) {
+          results.push({
+            location,
+            violation: {
+              message: `The proxy changed the relative order of the "${name}" field line values when forwarding the request. The order of same-name field lines is significant and MUST be preserved.`,
```
matthyk:
```suggestion
              message: `The proxy changed the relative order of the "${name}" field line values when forwarding the request.`,
```

matthyk (changes requested):

packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-not-generate-additional-whitespace-in-http-date.rule.ts
```diff
@@ -1,19 +1,104 @@
...
+  .type('test', 'analytics')
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.7')
   .description(
     'A sender MUST NOT generate additional whitespace in an HTTP-date beyond that specifically included as SP in the grammar.',
   )
   .summary(
     'Sender MUST NOT generate additional whitespace in HTTP-date beyond the grammar.',
   )
-  .tags('fields', 'date-time', 'whitespace')
+  .appliesTo('origin server')
```
matthyk:
```suggestion
```

packages/rules-rfc-9110/src/rules/fields/defining-field-values/lists/sender-must-not-generate-empty-list-elements.rule.ts
```diff
@@ -1,16 +1,90 @@
...
 export default httpRule('rfc9110/sender-must-not-generate-empty-list-elements')
   .severity('error')
-  .type('informational')
+  .type('test', 'analytics')
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.1')
   .description(
     'In any production that uses the list construct, a sender MUST NOT generate empty list elements.',
   )
   .summary('Sender MUST NOT generate empty list elements.')
   .tags('fields', 'lists')
```
matthyk:
```suggestion
```

matthyk (changes requested):

packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-not-generate-additional-whitespace-in-http-date.rule.ts
BaggersIO: Done in d25f03fb — dropped .appliesTo('origin server'). (The earlier commit had mistakenly removed .tags instead of this line, which is why it was still here.)

packages/rules-rfc-9110/src/rules/fields/defining-field-values/lists/sender-must-not-generate-empty-list-elements.rule.ts
BaggersIO: Done in d25f03fb — dropped .tags('fields', 'lists').

packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-generate-timestamps-in-imf-fixdate-format.rule.ts
```diff
@@ -1,14 +1,77 @@
...
+  return /^\d+$/.test(value.trim());
+}
+
 export default httpRule(
   'rfc9110/sender-must-generate-timestamps-in-imf-fixdate-format',
 )
   .severity('error')
-  .type('informational')
+  .type('test', 'analytics')
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.7')
```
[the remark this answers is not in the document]
matthyk: this is intended

packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-not-generate-additional-whitespace-in-http-date.rule.ts
```diff
@@ -1,19 +1,103 @@
...
+}
+
 export default httpRule(
   'rfc9110/sender-must-not-generate-additional-whitespace-in-http-date',
 )
   .severity('error')
-  .type('informational')
+  .type('test', 'analytics')
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.7')
   .description(
```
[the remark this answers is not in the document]
matthyk: this is intended because `sender` means that **ANY** sender of a HTTP-date must comply with this rule and not only origin servers

packages/rules-rfc-9110/src/rules/fields/field-names/proxy-must-forward-unrecognized-header-fields.rule.ts
```diff
@@ -1,13 +1,71 @@
...
+            (name) =>
+              !excepted.has(name) &&
+              getHeader(forwarded.request.data.headers, name) === undefined,
+          );
+        if (dropped.length > 0) {
+          results.push({
+            location,
+            violation: {
+              message: `A proxy did not forward header field(s) (${dropped.join(', ')}) present on the received request.`,
+            },
```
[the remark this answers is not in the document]
matthyk: Do not increase the length of the message

matthyk (approved):
I have just left some comments to copilot comments. LGTM

packages/rules-rfc-9110/src/rules/fields/field-names/proxy-must-forward-unrecognized-header-fields.rule.ts
```diff
@@ -1,13 +1,71 @@
...
+            getHeader(received.request.data.headers, 'connection'),
+          ),
+        ]);
+        const dropped = Object.keys(received.request.data.headers ?? {})
+          .map((name) => name.toLowerCase())
+          .filter(
+            (name) =>
+              !excepted.has(name) &&
+              getHeader(forwarded.request.data.headers, name) === undefined,
+          );
```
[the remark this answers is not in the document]
BaggersIO: Fixed in 30406d55. The hop is now skipped when either leg lacks captured headers, so an incomplete trace no longer reports every received field as "dropped". Note `getHeader` defaults `undefined` → `{}` (packages/core/src/utils.ts), so there was never a runtime throw — the issue was purely false positives. Added a regression test in `proxy-rules.test.ts` (forwarded leg with no captured headers ⇒ no findings).

packages/rules-rfc-9110/src/rules/fields/field-names/proxy-must-forward-unrecognized-header-fields.rule.ts
```diff
@@ -1,13 +1,72 @@
...
```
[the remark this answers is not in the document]
BaggersIO: Accepted as a best-effort signal. From traffic capture alone the rule cannot distinguish unrecognized-vs-recognized fields, nor a proxy legitimately configured to block/transform a field, so it flags dropped non-excepted fields and stays `analytics`-only (never a `test` assertion). This limitation is now documented as a caveat both in the code and in the PR description. Reviewed and approved by @matthyk.

packages/rules-rfc-9110/src/rules/fields/defining-field-values/lists/sender-must-not-generate-empty-list-elements.rule.ts
```diff
@@ -1,16 +1,89 @@
...
+    return false;
+  }
+  return elements.some((element) => element.trim() === '');
+}
+
 export default httpRule('rfc9110/sender-must-not-generate-empty-list-elements')
   .severity('error')
-  .type('informational')
+  .type('test', 'analytics')
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.1')
```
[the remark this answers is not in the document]
BaggersIO: Intentionally no `appliesTo` here. Per @matthyk on the sibling `sender-must-not-generate-additional-whitespace-in-http-date` thread, "sender" binds any sender of the field, not only origin servers — so this rule stays unscoped for consistency. The false-positive concern about comma-splitting is separately handled by curating the header whitelist (Cache-Control excluded).

packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-generate-timestamps-in-imf-fixdate-format.rule.ts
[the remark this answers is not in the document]
BaggersIO: Confirmed intended (per @matthyk above) — leaving unscoped. The rule binds any sender of an HTTP-date, so no `appliesTo`. Resolving.

packages/rules-rfc-9110/src/rules/fields/field-names/proxy-must-forward-unrecognized-header-fields.rule.ts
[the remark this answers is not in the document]
BaggersIO: Message left as-is per @matthyk — kept concise, not lengthened. The best-effort nature of the rule is documented in the code comment and PR description instead. Resolving.

packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-not-generate-additional-whitespace-in-http-date.rule.ts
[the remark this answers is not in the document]
BaggersIO: Confirmed — no `appliesTo`, per your note: "sender" means any sender of an HTTP-date, not only origin servers. Resolving.

packages/rules-rfc-9110/src/rules/fields/field-names/proxy-must-forward-unrecognized-header-fields.rule.ts
```diff
@@ -1,13 +1,72 @@
...
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1')
   .description(
     'A proxy MUST forward unrecognized header fields unless the field name is listed in the Connection header field (Section 7.6.1) or the proxy is specifically configured to block, or otherwise transform, such fields.',
   )
   .summary('Proxy MUST forward unrecognized header fields unless excepted.')
   .appliesTo('proxy')
   .tags('fields', 'field-names', 'proxy')
+  .rule((ctx) =>
+    ctx.validateCapturedHttpTraces((trace, location) => {
+      const results: RuleFnResult[] = [];
```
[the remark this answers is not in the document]
BaggersIO: Reconciled on the description side: this rule is intentionally implemented (outcome 1), and the PR description has been updated to remove the stale "no fn / outcome-3" wording. The residual soundness caveat (unrecognized-vs-recognized, configured blocking) is documented — see the reply on the related thread.

packages/rules-rfc-9110/src/rules/fields/field-order/proxy-must-not-change-field-line-order.rule.ts
```diff
@@ -12,4 +30,68 @@ export default httpRule('rfc9110/proxy-must-not-change-field-line-order')
   )
   .appliesTo('proxy')
   .tags('fields', 'field-order', 'proxy')
+  // Outcome 1: observable only from proxy-recorded traffic. Reordering of
+  // same-name field lines is invisible in a single message; it is detectable
+  // only by correlating a proxy's inbound request hop with the request it
+  // forwards, so this runs over captured per-hop traces.
+  .rule((ctx) =>
```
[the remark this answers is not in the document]
BaggersIO: Reconciled — the PR description now consistently describes this as an implemented outcome-1 analytics rule (no remaining outcome-3 / "no fn" wording).

packages/rules-rfc-9110/src/rules/fields/proxy-rules.test.ts
```diff
@@ -0,0 +1,137 @@
...
+  if (!analyze) {
+    throw new Error('rule is missing an analyzeRule');
+  }
+  const ctx = {
+    validateCapturedHttpTraces: (
+      validate: (t: CapturedTrace, location: string) => RuleFnResult[],
+    ) => validate(trace, 'trace-location'),
+  };
+  return analyze(ctx);
+}
```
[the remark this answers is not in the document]
BaggersIO: Addressed in 7d2fda16. `runOnTrace` now invokes the rule with the real `(context, options, logger)` signature — passing `{ mode: 'analytics' }` and a logger stub matching what the rule runner supplies — and throws if a rule returns a `Promise`, so a future async rule fails loudly instead of silently passing.

PR #304, opened 2026-06-29
feat(rules-rfc-9110): align message-context rule execution contexts (#327)

BaggersIO:
## Summary

Aligns the execution context(s) of every rule in the **message-context** category (`packages/rules-rfc-9110/src/rules/message-context/`) with what can *actually and meaningfully* run, using only the existing rule framework (part of #327). Covers the Expect, From, Referer, TE, User-Agent (request side) and Allow, Location, Server (response side) fields.

Guiding rules applied per the shared brief:
- **Request-side rules must not be `test`** (Thymian is the sender, so the user cannot drive the non-conformant request). They become `static` and/or `analytics`.
- **Common projection exposes header NAMES only** — value checks use `validateHttpTransactions` + `getHeader` on the live request/response model (analyze).
- **`appliesTo` role-filtering**: request-side analyze rules use `('client','user-agent')` (HAR request role defaults to `user-agent`); response-side analyze rules keep `'origin server'`.
- Header comparisons are case-insensitive (`getHeader` / `.toLowerCase()`).
- Every `informational` and infra-deferred rule carries a written rationale in a code comment near `.type(...)`.

36 rules reviewed.

## Coverage matrix

|---|---|---|---|---|
| **expect** / client-may-proceed-to-send-content-without-receiving-100-response | informational | informational | 2 | MAY about client timing; unobservable |
| **expect** / client-must-not-generate-100-continue-without-content | static,test,analytics | static,analytics | 4 | request-side → drop test; shared filter-only rule (Expect:100-continue + no body); appliesTo client,user-agent |
| **expect** / client-must-send-expect-header-for-100-response | informational | informational | 2 | conditioned on internal client intent (will-wait); unobservable |
| **expect** / client-should-not-wait-for-an-indefinite-period-for-100-response | informational | informational | 2 | client wait timing; unobservable |
| **expect** / client-should-repeat-request-without-expect-for-417 | informational | informational | 2 | cross-request retry strategy; not reconstructable |
| **expect** / origin-server-must-not-wait-for-content-before-100-continue | informational | informational | 2 | server timing; unobservable |
| **expect** / origin-server-must-respond-immediately-to-100-continue-request | informational | informational | 2 | immediacy + interim 100; unobservable |
| **expect** / proxy-may-generate-immediate-100-response | informational | informational | 2 | permissive MAY on proxy internal state |
| **expect** / proxy-must-handle-100-continue-expectation | informational | informational | 2 | both outcomes conformant; needs cross-hop proxy view |
| **expect** / server-may-omit-sending-100-response-if-already-received-content | informational | informational | 2 | permissive MAY; interim response not captured |
| **expect** / server-may-respond-with-417-response-for-other-expect-than-100-continue | analytics (+rule) | informational | 4 | **inverted MAY removed** — flagged conformant non-417 responses as violations |
| **expect** / server-must-ignore-100-continue-in-http-1.0 | informational | informational | 2 | depends on HTTP version (not exposed) + "ignore" non-action |
| **expect** / server-must-send-final-status-after-100-continue | informational | informational | 2 | connection lifecycle; premature-close exception unobservable |
| **expect** / server-should-indicate-if-closing-the-connection-... | informational | informational | 2 | connection-management framing; unobservable |
| **from** / robotic-user-agent-should-send-valid-from-header | static,analytics (+rule+override) | analytics | 4 | static dropped (can't validate value or detect "robotic"); analytics validates a present From is a valid mailbox; appliesTo client,user-agent |
| **from** / server-should-not-use-from-for-authentication | informational | informational | 2 | server internal authz decision; unobservable |
| **from** / user-agent-should-not-send-from-without-configuration | static,test,analytics | static,analytics | 4 | request-side → drop test; presence check (names only) |
| **referer** / intermediary-should-not-modify-referer-for-same-scheme-and-host | informational | informational | 2 | modification needs cross-hop correlation |
| **referer** / user-agent-may-truncate-parts-other-than-referring-origin | informational | informational | 2 | permissive MAY; also fixed wrong url anchor → #name-referer |
| **referer** / user-agent-must-exclude-referer-or-send-about-blank-for-no-source | informational | informational | 2 | triggering condition (target URI source) not in request |
| **referer** / user-agent-must-not-include-fragment-or-userinfo-in-referer | analytics (+override) | analytics | 1 (kept) | value check kept; added descriptive messages + security note; appliesTo user-agent |
| **referer** / user-agent-must-not-send-referer-in-unsecured-request-from-secure-resource | analytics (filter-only) | analytics | 1 (repaired) | **bug fix**: now requires the Referer *value* to be https before flagging (was flagging all http requests with any Referer) |
| **referer** / user-agent-should-not-send-referer-for-secure-to-insecure | analytics (+override) | analytics | 1 (kept) | kept; hardened null-origin handling, added message |
| **te** / sender-must-send-te-connection-option-with-te-header | static,test,analytics | static,analytics | 4 | request-side → drop test; documented under-approximation (names only) |
| **user-agent** / sender-must-not-generate-advertising-in-product-identifier | analytics (+override) | analytics | 1 (kept) | value check kept; appliesTo client,user-agent; added message |
| **user-agent** / sender-should-limit-generated-product-identifiers-to-necessity | informational | informational | 2 | subjective "necessary" |
| **user-agent** / sender-should-not-generate-non-version-info-in-product-version | analytics (+override) | analytics | 1 (kept) | value check kept; appliesTo client,user-agent; added message |
| **user-agent** / user-agent-should-limit-addition-of-subproducts-by-third-parties | informational | informational | 2 | depends on "who added" + subjective limit |
| **user-agent** / user-agent-should-not-generate-overly-detailed-user-agent | informational | informational | 2 | subjective SHOULD NOT |
| **user-agent** / user-agent-should-send-user-agent-header | analytics (+rule) | analytics | 1 (kept) | presence check (names only) kept; added message |
| **allow** / origin-sever-may-send-allow-header | static,test,analytics (+rule) | informational | 4 | **inverted MAY removed** — flagged responses *without* Allow as violations |
| **allow** / proxy-must-not-modify-allow-header | informational | informational | 2 | modification needs cross-hop correlation |
| **location** / user-agent-must-inherit-fragment-for-3xx-without-fragment | informational | informational | 2 | recipient internal redirect processing |
| **server** / origin-server-may-generate-server-header-field | static,test,analytics (+rule) | informational | 4 | **inverted MAY removed** — flagged responses *without* Server as violations |
| **server** / origin-server-should-limit-addition-of-subproducts-by-third-parties | informational | informational | 2 | depends on "who added" + subjective limit |
| **server** / origin-server-should-not-generate-detailed-server-header | informational | informational | 2 | subjective SHOULD NOT |

**Outcome counts:** Outcome 1 (implementable/kept active) = **6** · Outcome 2 (informational) = **20** · Outcome 3 (infra-deferred analytics, no fn) = **0** · Outcome 4 (reclassified) = **10**.

No outcome-3 rules: the proxy/intermediary "must not modify" and "must handle" cases here require correlating the same message across two hops, which is not reconstructable from a single captured transaction / standard HAR, so they are honestly informational rather than typed analytics without a function.

## Per-PR checklist

- [x] Every rule in the category reviewed exactly once (36).
- [x] No declared active context (`static`/`test`/`analytics`) left without a rule function (no outcome-3 rules in this category).
- [x] No request-side rule assigned to `test` (dropped from 4 rules).
- [x] Every `informational` and infra-deferred rule carries a written reason in a code comment near `.type(...)`.
- [x] Header comparisons are case-insensitive (`getHeader` / `.toLowerCase()`).
- [x] `appliesTo` roles set so analyze rules actually fire on HAR (request-side → `client,user-agent`; response-side → `origin server`).
- [x] Common projection used only for presence/status/media-type/body; value checks use `validateHttpTransactions` + `getHeader`.
- [x] No new metadata fields / framework changes; only files under `message-context/` edited.
- [x] `npx prettier --check` passes on the category.
- [x] Security-relevant rules addressed (below).

## Security-relevant rules

- **Referer / userinfo + fragment leak** (`user-agent-must-not-include-fragment-or-userinfo-in-referer`): kept as an active analytics value check (error). Leaking userinfo (credentials) or a fragment to a third-party origin is a real disclosure risk; added explicit credential/fragment violation messages.
- **Referer / secure→insecure leak** (`user-agent-must-not-send-referer-in-unsecured-request-from-secure-resource`): **repaired** — previously flagged *every* http request that carried any Referer (false positives, and missed the actual security condition). Now flags only when the unsecured request carries a Referer whose value is `https://`, i.e. a secure referring resource leaking into an unsecured request.
- **Referer / cross-origin from secure resource** (`user-agent-should-not-send-referer-for-secure-to-insecure`): kept; hardened to avoid false positives when an origin cannot be parsed, with a descriptive message.
- **Server header disclosure** (`origin-server-may-generate-server-header-field`, `origin-server-should-not-generate-detailed-server-header`): the MAY rule that flagged *absent* Server headers was removed (omitting Server is a legitimate security-positive choice); the SHOULD-NOT-detailed rule remains informational because "needlessly fine-grained" has no objective threshold.
- **Auth / From** (`server-should-not-use-from-for-authentication`): informational — using From for authz is a server-internal decision not observable from messages.

## Reclassification reasoning (non-trivial)

1. **Inverted MAYs (3 rules).** `origin-sever-may-send-allow-header`, `origin-server-may-generate-server-header-field` and `server-may-respond-with-417-...` shipped `.rule()` functions that treated the *absence* of an optional behaviour as a violation. That inverts the RFC's MAY and fires on conformant traffic. Per the brief (permissive MAY with no non-conformant condition → informational), all three are reclassified `informational` with the inverted checks removed and a rationale comment.

2. **Request-side `test` removal (4 rules).** `client-must-not-generate-100-continue-without-content`, `user-agent-should-not-send-from-without-configuration`, `sender-must-send-te-connection-option-with-te-header`, and `robotic-user-agent-should-send-valid-from-header` constrain what the *client* sends. Under `test`, Thymian generates the request and the user cannot exercise the non-conformant case, so `test` is dropped. The first three stay `static,analytics` (presence-level checks valid via the common projection); `robotic-...-from` drops `static` too (the From value cannot be validated at design time and "robotic" cannot be inferred from a spec) and runs `analytics`-only, validating that a present From is a valid mailbox.

3. **`appliesTo` widening.** Request-side analyze rules previously scoped to `'client'` (or `'origin server'` for response side) were widened to include `'user-agent'` where appropriate so they actually match HAR requests (whose default request role is `user-agent`).

This round addresses the #327 execution-context contract feedback for the message-context category.

### FIX-1 — TE requires the "te" Connection option (value check)
`request-context/te/sender-must-send-te-connection-option-with-te-header.rule.ts` previously ran a common `.rule()` that only checked the Connection header was PRESENT (`not(requestHeader('connection'))`), which cannot see values. The RFC requires the Connection VALUE to contain the `te` connection-option.
- Added `overrideAnalyticsRule` using `validateHttpTransactions(requestHeader('te'), fn)` that reads `getHeader(request.headers, 'connection')`, splits on commas, trims/lowercases, and flags when the tokens do not include `te`.
- Kept a coarse structural static check (Connection present when TE present) via `overrideStaticRule`, with a written rationale comment explaining that static can only assert presence, not the value.

### FIX-2 — Expect: 100-continue without content (value check)
`request-context/expect/client-must-not-generate-100-continue-without-content.rule.ts` previously ran a common `.rule()` with `requestHeader('expect', '100-continue')` whose value argument is IGNORED by the common filter visitor, so it flagged ANY `Expect` header on a body-less request.
- Added `overrideAnalyticsRule` using `validateHttpTransactions(and(requestHeader('expect'), not(hasRequestBody())), fn)` that reads `getHeader(request.headers, 'expect')`, splits/trims/lowercases, and flags ONLY when the value includes exactly `100-continue`.
- Kept the coarse common form as a static prefilter (`overrideStaticRule`) with a written rationale comment.

### FIX-3 — Rationale comments on informational rules (Theme C)
Added a concise 1-3 line rationale comment adjacent to `.type('informational')` for all 26 informational rules in `message-context/**` that lacked one, stating WHY each is informational (subjective/editorial judgment, permission (MAY) with no violating behaviour, client/server-internal state or timing with no wire signature, or cross-message/cross-hop correlation that is not currently available). Two rules whose real check would need per-hop captured-trace correlation (`intermediary-should-not-modify-referer-for-same-scheme-and-host`, `proxy-must-not-modify-allow-header`) note that they are analytics captured-trace candidates pending suitable recorded deployment traffic.

### Verification
- `npx tsc -b packages/rules-rfc-9110/tsconfig.lib.json` exits 0.
- Changed files formatted with prettier.


packages/rules-rfc-9110/src/rules/message-context/request-context/expect/client-must-not-generate-100-continue-without-content.rule.ts
```diff
@@ -5,12 +5,12 @@ export default httpRule(
   'rfc9110/client-must-not-generate-100-continue-without-content',
 )
   .severity('error')
-  .type('static', 'analytics', 'test')
+  .type('static', 'analytics')
```
BaggersIO: Request-side rule (constrains what the *client* generates). `test` is dropped per #327: during a test the request is generated by Thymian, so the user cannot drive the non-conformant case. `static` validates the request shape described in the spec and `analytics` validates real recorded client requests; both only need header-name and body presence (available via the common projection), so a single shared filter-only rule covers them. Request-side: match HAR requests (default request role `user-agent`) as well as generically labelled clients.

packages/rules-rfc-9110/src/rules/message-context/request-context/expect/server-may-respond-with-417-response-for-other-expect-than-100-continue.rule.ts
```diff
@@ -1,23 +1,13 @@
-import { and, not, requestHeader, statusCode } from '@thymian/core';
 import { httpRule } from '@thymian/core';
 
 export default httpRule(
   'rfc9110/server-may-respond-with-417-response-for-other-expect-than-100-continue',
 )
   .severity('hint')
-  .type('analytics')
+  .type('informational')
```
BaggersIO: Informational (#327): this is a permissive MAY with no non-conformant condition — a server *may* answer an unsupported Expect with 417, but it is equally free to handle it another way (e.g. ignore the expectation). The previous `analytics` rule treated every non-417 response to a non- 100-continue Expect as a violation, which inverts the MAY and fires on conformant traffic. There is no observable failure to flag, so the rule is reclassified informational and ships no rule function.

packages/rules-rfc-9110/src/rules/message-context/request-context/expect/server-must-ignore-100-continue-in-http-1.0.rule.ts
```diff
@@ -1,6 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// could be analytics at some point --> we should add the HTTP to the analytics context
 export default httpRule('rfc9110/server-must-ignore-100-continue-in-http-1.0')
   .severity('hint')
   .type('informational')
```
BaggersIO, line 8: Informational (#327): conformance turns on the request's HTTP *version* (the expectation must be ignored for HTTP/1.0). The protocol version is not exposed by the common projection nor by the live request/response model available here, and "ignore" is itself a non-action with no observable signature. No detectable non-conformant case in the current framework, so it stays informational. No rule function.

packages/rules-rfc-9110/src/rules/message-context/request-context/from/robotic-user-agent-should-send-valid-from-header.rule.ts
```diff
@@ -11,17 +10,19 @@ export default httpRule(
   'rfc9110/robotic-user-agent-should-send-valid-from-header',
 )
   .severity('warn')
-  .type('static', 'analytics')
+  .type('analytics')
```
BaggersIO, line 13: Request-side, analytics-only (#327). The previous `static` slot validated nothing meaningful: the common (lint) projection exposes header *names* only, so the From value cannot be validated at design time, and whether a user agent is "robotic" cannot be inferred from a spec. `test` is inapplicable (Thymian generates the request, so it cannot exercise an invalid From). The honest, observable subset is: on recorded client traffic, a present From header should carry a valid mailbox. Request-side: HAR requests default to the `user-agent` role.

packages/rules-rfc-9110/src/rules/message-context/request-context/from/user-agent-should-not-send-from-without-configuration.rule.ts
```diff
@@ -5,7 +5,7 @@ export default httpRule(
   'rfc9110/user-agent-should-not-send-from-without-configuration',
 )
   .severity('warn')
-  .type('static', 'analytics', 'test')
+  .type('static', 'analytics')
```
BaggersIO, line 8: Request-side rule (#327). `test` dropped: Thymian generates the request, so it cannot exercise the case of an unconfigured From. The "without explicit configuration" condition is itself unobservable, so the honest, identical check across `static` (described request) and `analytics` (recorded request) is header *presence*, which the common projection provides. A single shared filter-only rule covers both.

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-may-truncate-parts-other-than-referring-origin.rule.ts
```diff
@@ -5,7 +5,7 @@ export default httpRule(
 )
   .severity('hint')
   .type('informational')
```
BaggersIO, line 9: Informational (#327): a permissive MAY (the user agent is free to truncate the Referer beyond the origin). There is no non-conformant condition to detect. No rule function.

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-must-not-send-referer-in-unsecured-request-from-secure-resource.rule.ts
```diff
@@ -10,10 +16,39 @@ export default httpRule(
   .description(
     'A user agent MUST NOT send a Referer header field in an unsecured HTTP request if the referring resource was accessed with a secure protocol.',
   )
+  .summary(
+    'A user agent MUST NOT send a Referer referring to an https resource in an unsecured http request.',
+  )
   .appliesTo('user-agent')
```
BaggersIO, line 22: (Anchored to the `.appliesTo` line; the rule's `.type` line is not part of this PR's diff hunk.) Request-side, analytics-only (#327): the referring resource (the Referer value) must be inspected to see whether it was accessed over a secure protocol. That requires the header *value*, available only on recorded traffic via the live request model. Not `test` (Thymian generates the request) and not `static` (value not exposed by the common projection). Request-side: HAR requests default to the `user-agent` role.

packages/rules-rfc-9110/src/rules/message-context/request-context/te/sender-must-send-te-connection-option-with-te-header.rule.ts
```diff
@@ -5,12 +5,12 @@ export default httpRule(
   'rfc9110/sender-must-send-te-connection-option-with-te-header',
 )
   .severity('error')
-  .type('static', 'analytics', 'test')
+  .type('static', 'analytics')
```
BaggersIO, line 15: Request-side rule (#327). `test` dropped: Thymian generates the request, so it cannot exercise a sender that omits the Connection option. Validated identically across `static` (described request) and `analytics` (recorded request) using header-name presence only (the common projection). NOTE: the common projection exposes only header *names*, so this catches a missing Connection header entirely; it cannot verify a present Connection actually lists the "TE" option (that needs the value), which is an accepted under-approximation of the MUST. Request-side: HAR requests default to the `user-agent` role.

packages/rules-rfc-9110/src/rules/message-context/request-context/user-agent/sender-must-not-generate-advertising-in-product-identifier.rule.ts
```diff
@@ -16,11 +15,11 @@ export default httpRule(
   .description(
     'A sender MUST NOT generate advertising or other nonessential information within the product identifier.',
   )
-  .appliesTo('client')
+  .appliesTo('client', 'user-agent')
```
BaggersIO: (Anchored to the `.appliesTo` line; the rule's `.type` line is not part of this PR's diff hunk.) Request-side, analytics-only (#327): needs the User-Agent *value* (only on recorded traffic via the live request model). Not `test` (Thymian generates the request) nor `static` (value not in the common projection). Request-side: HAR requests default to the `user-agent` role.

packages/rules-rfc-9110/src/rules/message-context/request-context/user-agent/sender-should-not-generate-non-version-info-in-product-version.rule.ts
```diff
@@ -16,11 +15,11 @@ export default httpRule(
   .description(
     'A sender SHOULD NOT generate information in product-version that is not a version identifier (i.e., successive versions of the same product name ought to differ only in the product-version portion of the product identifier).',
   )
-  .appliesTo('client')
+  .appliesTo('client', 'user-agent')
```
BaggersIO: (Anchored to the `.appliesTo` line; the rule's `.type` line is not part of this PR's diff hunk.) Request-side, analytics-only (#327): needs the User-Agent *value* (only on recorded traffic via the live request model). Not `test` (Thymian generates the request) nor `static` (value not in the common projection). Request-side: HAR requests default to the `user-agent` role.

packages/rules-rfc-9110/src/rules/message-context/request-context/user-agent/user-agent-should-send-user-agent-header.rule.ts
```diff
@@ -10,6 +10,18 @@ export default httpRule('rfc9110/user-agent-should-send-user-agent-header')
   )
   .appliesTo('user-agent')
```
BaggersIO, line 11: (Anchored to the `.appliesTo` line; the rule's `.type` line is not part of this PR's diff hunk.) Request-side, analytics-only (#327): a User-Agent header *presence* check on real recorded requests. Header presence is enough, so the common projection (header names) suffices. Not `test` (Thymian generates the request and controls the User-Agent it sends) nor `static` (the spec does not assert a client always populates User-Agent). Request-side: HAR requests default to the `user-agent` role.

packages/rules-rfc-9110/src/rules/message-context/response-context/allow/origin-server-may-send-allow-header.rule.ts
```diff
@@ -1,13 +1,9 @@
-import { not, responseHeader } from '@thymian/core';
 import { httpRule } from '@thymian/core';
 
 export default httpRule('rfc9110/origin-sever-may-send-allow-header')
   .severity('hint')
-  .type('static', 'analytics', 'test')
+  .type('informational')
```
BaggersIO, line 7: Informational (#327): a permissive MAY — the origin server is free to send an Allow header, or not. The previous rule flagged every response *without* an Allow header as a violation, which inverts the MAY and fires on virtually all conformant traffic. There is no non-conformant condition to detect, so the rule is reclassified informational and ships no rule function.

packages/rules-rfc-9110/src/rules/message-context/response-context/allow/proxy-must-not-modify-allow-header.rule.ts
```diff
@@ -1,6 +1,5 @@
 import { httpRule } from '@thymian/core';
 
-// also analytics rule in the future
 export default httpRule('rfc9110/proxy-must-not-modify-allow-header')
   .severity('hint')
   .type('informational')
```
BaggersIO, line 9: Informational (#327): detecting a *modification* requires comparing the Allow value the proxy received from upstream against the value it forwarded downstream — i.e. correlating the same response at two hops. That cross-hop view is not reconstructable from a standard captured transaction / HAR, so this is not turned into an analytics check. No rule function.

packages/rules-rfc-9110/src/rules/message-context/response-context/server/origin-server-may-generate-server-header-field.rule.ts
```diff
@@ -1,17 +1,13 @@
-import { not, responseHeader } from '@thymian/core';
 import { httpRule } from '@thymian/core';
 
 export default httpRule(
   'rfc9110/origin-server-may-generate-server-header-field',
 )
   .severity('hint')
-  .type('static', 'analytics', 'test')
+  .type('informational')
```
BaggersIO: Informational (#327): a permissive MAY — the origin server is free to emit a Server header, or not. The previous rule flagged every response *without* a Server header as a violation, which inverts the MAY and fires on conformant traffic. (Omitting Server is in fact a common, security-positive choice as it avoids advertising software/version details.) No non-conformant condition exists, so the rule is reclassified informational with no rule function.

BaggersIO (commented):
Recording the execution-context triage rationale here (moved out of inline code comments per #327).

14 rationale(s) are attached as line-anchored review comments below. The remaining rules' `.type`/`.appliesTo` lines are not part of this PR's current diff (several reverted to match `main` once the inline comment was removed), so their rationale is recorded here instead:

- **request-context/expect/client-may-proceed-to-send-content-without-receiving-100-response.rule.ts** (`rfc9110/client-may-proceed-to-send-content-without-receiving-100-response`): Informational (#327): a permissive MAY about client timing (whether/when it proceeds to send content). There is no non-conformant condition to detect, and the timing of content transmission is not visible in a captured request/response transaction. No rule function.
- **request-context/expect/client-must-send-expect-header-for-100-response.rule.ts** (`rfc9110/client-must-send-expect-header-for-100-response`): Informational (#327): the requirement is conditioned on internal client intent ("a client that *will wait* for 100 Continue"). Whether a client intends to wait is not observable from a request, so there is no detectable non-conformant case. No rule function.
- **request-context/expect/client-should-not-wait-for-an-indefinite-period-for-100-response.rule.ts** (`rfc9110/client-should-not-wait-for-an-indefinite-period-for-100-response`): Informational (#327): a SHOULD NOT about client wait *timing*. The duration a client waits before sending content is not represented in a captured transaction, so the non-conformant case is unobservable. No rule function.
- **request-context/expect/client-should-repeat-request-without-expect-for-417.rule.ts** (`rfc9110/client-should-repeat-request-without-expect-for-417`): Informational (#327): describes a client retry strategy spanning two separate requests (a 417, then a repeat without the expectation). Detecting conformance requires correlating a follow-up request to the original, which is not reliably reconstructable from a captured transaction. No rule function.
- **request-context/expect/origin-server-must-not-wait-for-content-before-100-continue.rule.ts** (`rfc9110/origin-server-must-not-wait-for-content-before-100-continue`): Informational (#327): a MUST NOT about server *timing* (it must not delay the 100 Continue until content arrives). The relative timing of the interim response versus content receipt is not captured in a transaction record, so the non-conformant case is unobservable. No rule function.
- **request-context/expect/origin-server-must-respond-immediately-to-100-continue-request.rule.ts** (`rfc9110/origin-server-must-respond-immediately-to-100-continue-request`): Informational (#327): a MUST about *immediacy* of the response to a 100-continue request. Both the satisfying behaviours (immediate final status OR immediate 100 Continue) are conformant, and the interim 100 Continue + its timing are not represented in a captured request/response transaction. No detectable non-conformant case. No rule function.
- **request-context/expect/proxy-may-generate-immediate-100-response.rule.ts** (`rfc9110/proxy-may-generate-immediate-100-response`): Informational (#327): a permissive MAY governing an optional proxy optimisation, predicated on the proxy's private belief about a downstream server's HTTP version. No non-conformant condition exists and the proxy's internal state is unobservable. No rule function.
- **request-context/expect/proxy-must-handle-100-continue-expectation.rule.ts** (`rfc9110/proxy-must-handle-100-continue-expectation`): Informational (#327): a MUST whose conformant outcomes (send an immediate final status OR forward the request) cannot be distinguished from non-conformant behaviour without observing the proxy's ingress AND egress for the same message. That cross-hop correlation at a proxy is not reconstructable from a standard captured transaction / HAR, so this is not turned into an analytics check. No rule function.
- **request-context/expect/server-may-omit-sending-100-response-if-already-received-content.rule.ts** (`rfc9110/server-may-omit-sending-100-response-if-already-received-content`): Informational (#327): a permissive MAY (server may skip the 100 Continue). No non-conformant condition, and the interim response is not represented in a captured request/response transaction. No rule function.
- **request-context/expect/server-must-send-final-status-after-100-continue.rule.ts** (`rfc9110/server-must-send-final-status-after-100-continue`): Informational (#327): a MUST about eventual delivery of a final status after a 100 Continue, with an explicit exception for prematurely closed connections. Distinguishing a missing final status from a legitimate premature close requires connection-lifecycle observation that a captured request/response transaction does not provide. No rule function.
- **request-context/expect/server-should-indicate-if-closing-the-connection-when-sending-final-status-code-without-full-request.rule.ts** (`rfc9110/server-should-indicate-if-closing-the-connection-when-sending-final-status-code-without-full-request`): Informational (#327): conformance depends on connection-management framing (whether the server signals it will close vs. keep reading the request content), which is HTTP/1.1 connection-level state not represented in a captured request/response transaction. No rule function.
- **request-context/from/server-should-not-use-from-for-authentication.rule.ts** (`rfc9110/server-should-not-use-from-for-authentication`): Informational (#327): a SHOULD NOT governing the server's *internal* use of the From header for access control / authentication. Whether a server uses a header for an authorization decision is not observable from the request/response messages. No rule function.
- **request-context/referer/intermediary-should-not-modify-referer-for-same-scheme-and-host.rule.ts** (`rfc9110/intermediary-should-not-modify-referer-for-same-scheme-and-host`): Informational (#327): detecting a *modification* requires comparing the Referer value as it entered the intermediary against the value it forwarded — i.e. correlating the same message at two hops. That cross-hop view is not reconstructable from a standard captured transaction / HAR, so this is not turned into an analytics check. No rule function.
- **request-context/referer/user-agent-must-exclude-referer-or-send-about-blank-for-no-source.rule.ts** (`rfc9110/user-agent-must-exclude-referer-or-send-about-blank-for-no-source`): Informational (#327): conformance is conditioned on the *origin* of the target URI (typed by hand, chosen from bookmarks, …) — knowledge held only by the user agent. That source is not present in the request, so the triggering condition cannot be observed and there is no detectable non-conformant case. No rule function.
- **request-context/referer/user-agent-must-not-include-fragment-or-userinfo-in-referer.rule.ts** (`rfc9110/user-agent-must-not-include-fragment-or-userinfo-in-referer`): Request-side, analytics-only (#327): requires the Referer *value* (only on recorded traffic via the live request model). Not `test` (Thymian generates the request) nor `static` (value not in the common projection). Security-relevant: leaking userinfo (credentials) or a fragment in Referer can expose secrets or sensitive in-page state to a third-party origin. Request-side: HAR requests default to the `user-agent` role.
- **request-context/referer/user-agent-should-not-send-referer-for-secure-to-insecure.rule.ts** (`rfc9110/user-agent-should-not-send-referer-for-secure-to-insecure`): Request-side, analytics-only (#327): needs the Referer *value* and the request origin to compare origins — available only on recorded traffic via the live request model. Not `test` (Thymian generates the request) nor `static` (value not in the common projection). Security-relevant: a cross-origin Referer originating from a secure resource can leak the referring URL to a different origin. Request-side: HAR requests default to the `user-agent` role.
- **request-context/user-agent/sender-should-limit-generated-product-identifiers-to-necessity.rule.ts** (`rfc9110/sender-should-limit-generated-product-identifiers-to-necessity`): Informational (#327): a subjective SHOULD ("limit to what is necessary"). There is no objective, non-arbitrary threshold that distinguishes a conformant from a non-conformant product identifier, so a programmatic check would be unreliable (false positives). No rule function.
- **request-context/user-agent/user-agent-should-limit-addition-of-subproducts-by-third-parties.rule.ts** (`rfc9110/user-agent-should-limit-addition-of-subproducts-by-third-parties`): Informational (#327): conformance turns on *who* added a subproduct (a third party) and a subjective notion of "limit", neither of which is observable from the User-Agent value alone. No reliable non-conformant condition to detect. No rule function.
- **request-context/user-agent/user-agent-should-not-generate-overly-detailed-user-agent.rule.ts** (`rfc9110/user-agent-should-not-generate-needlessly-fine-grained-detailed-user-agent-field`): Informational (#327): a subjective SHOULD NOT ("needlessly fine-grained detail"). There is no objective threshold separating conformant from non-conformant detail, so a programmatic check would be arbitrary and false-positive prone. No rule function.
- **response-context/location/user-agent-must-inherit-fragment-for-3xx-without-fragment.rule.ts** (`rfc9110/user-agent-must-inherit-fragment-for-3xx-without-fragment`): Informational (#327): a MUST governing the user agent's *internal* processing of a redirect (how it reconstructs the target's fragment). This is recipient-side behaviour that produces no observable signal in the request/response messages, so there is no detectable non-conformant case. No rule function.
- **response-context/server/origin-server-should-limit-addition-of-subproducts-by-third-parties.rule.ts** (`rfc9110/origin-server-should-limit-addition-of-subproducts-by-third-parties`): Informational (#327): conformance turns on *who* added a subproduct (a third party) to the Server header and a subjective notion of "limit", neither of which is observable from the Server value alone. No reliable non-conformant condition to detect. No rule function.
- **response-context/server/origin-server-should-not-generate-detailed-server-header.rule.ts** (`rfc9110/origin-server-should-not-generate-needlessly-fine-grained-detail-server-header`): Informational (#327): a subjective SHOULD NOT ("needlessly fine-grained detail" in the Server header). There is no objective threshold separating conformant from non-conformant detail, so a programmatic check would be arbitrary and false-positive prone. (The related security concern — leaking precise software/versions — is best left to the operator's judgement.) No rule function.

packages/rules-rfc-9110/src/rules/message-context/request-context/expect/client-must-not-generate-100-continue-without-content.rule.ts
```diff
@@ -1,19 +1,57 @@
...
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#name-expect')
   .description(
     'A client MUST NOT generate a 100-continue expectation in a request that does not include content.',
   )
-  .appliesTo('client')
-  .rule((ctx) =>
+  .appliesTo('client', 'user-agent')
```
matthyk:
```suggestion
  .appliesTo('client')
```

packages/rules-rfc-9110/src/rules/message-context/request-context/expect/client-must-not-generate-100-continue-without-content.rule.ts
```diff
@@ -1,19 +1,57 @@
...
+  type RuleViolationLocation,
+} from '@thymian/core';
 import { httpRule } from '@thymian/core';
 
 export default httpRule(
   'rfc9110/client-must-not-generate-100-continue-without-content',
 )
   .severity('error')
-  .type('static', 'analytics', 'test')
+  .type('static', 'analytics')
```
matthyk:
```suggestion
  .type('analytics')
```

packages/rules-rfc-9110/src/rules/message-context/request-context/expect/client-must-not-generate-100-continue-without-content.rule.ts
```diff
@@ -1,19 +1,57 @@
...
+                location,
+                violation: {
+                  message:
+                    'A client MUST NOT generate a 100-continue expectation in a request that does not include content, but this request sends "Expect: 100-continue" with no content.',
+                },
+                findings: [],
+              },
+            ]
+          : [];
+      },
```
matthyk, line 47:
```suggestion
    ctx.validateHttpTransactions(
      and(requestHeader('expect', '100-continue'), not(hasRequestBody()))
```

packages/rules-rfc-9110/src/rules/message-context/request-context/from/robotic-user-agent-should-send-valid-from-header.rule.ts
```diff
@@ -11,17 +10,19 @@ export default httpRule(
...
   .description(
     'A robotic user agent SHOULD send a valid From header field so that the person responsible for running the robot can be contacted if problems occur on servers.',
   )
-  .appliesTo('client')
-  .rule((ctx) => ctx.validateCommonHttpTransactions(and(requestHeader('from'))))
-  .overrideAnalyticsRule((ctx) =>
+  .summary(
+    'A From header field sent by a (robotic) user agent SHOULD contain a valid mailbox.',
+  )
+  .appliesTo('client', 'user-agent')
```
matthyk:
```suggestion
  .appliesTo('user-agent')
```

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-must-not-include-fragment-or-userinfo-in-referer.rule.ts
```diff
@@ -19,15 +18,23 @@ export default httpRule(
...
         }
 
         // Check for fragment (#)
         if (referer.includes('#')) {
-          return [{ location, violation: {}, findings: [] }];
+          return [
+            {
+              location,
+              violation: {
+                message: `The Referer header value "${referer}" contains a fragment component. A user agent MUST NOT include the fragment when generating the Referer field value.`,
```
matthyk:
```suggestion
                message: `The Referer header value "${referer}" contains a fragment component.`,
```

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-must-not-include-fragment-or-userinfo-in-referer.rule.ts
```diff
@@ -36,13 +43,29 @@ export default httpRule(
           const url = new URL(referer);
           // URL.username or URL.password being non-empty indicates userinfo presence
           if (url.username || url.password) {
-            return [{ location, violation: {}, findings: [] }];
+            return [
+              {
+                location,
+                violation: {
+                  message: `The Referer header value "${referer}" contains a userinfo component (credentials). A user agent MUST NOT include userinfo when generating the Referer field value.`,
```
matthyk:
```suggestion
                  message: `The Referer header value "${referer}" contains a userinfo component (credentials).`,
```

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-must-not-include-fragment-or-userinfo-in-referer.rule.ts
```diff
@@ -36,13 +43,29 @@ export default httpRule(
...
         } catch {
           // If URL parsing fails, check for @ before the first / after ://
           const match = referer.match(/^[^:]+:\/\/([^/]+)/);
           if (match && match[1]?.includes('@')) {
-            return [{ location, violation: {}, findings: [] }];
+            return [
+              {
+                location,
+                violation: {
+                  message: `The Referer header value "${referer}" appears to contain a userinfo component (credentials). A user agent MUST NOT include userinfo when generating the Referer field value.`,
```
matthyk:
```suggestion
                  message: `The Referer header value "${referer}" appears to contain a userinfo component (credentials).`,
```

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-must-not-send-referer-in-unsecured-request-from-secure-resource.rule.ts
```diff
@@ -10,10 +16,39 @@ export default httpRule(
...
+        // conformant and must not be flagged.
+        if (!referer.trim().toLowerCase().startsWith('https://')) {
+          return [];
+        }
+
+        return [
+          {
+            location,
+            violation: {
+              message: `An unsecured (http) request carries a Referer header referring to a secure resource (${referer.trim()}). A user agent MUST NOT leak a Referer that was accessed with a secure protocol into an unsecured request.`,
```
matthyk:
```suggestion
              message: `An unsecured (http) request carries a Referer header referring to a secure resource (${referer.trim()}).`,
```

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-should-not-send-referer-for-secure-to-insecure.rule.ts
```diff
@@ -35,10 +34,25 @@ export default httpRule(
...
+          refererOrigin === requestOrigin
+        ) {
+          return [];
+        }
+
+        return [
+          {
+            location,
+            violation: {
+              message: `A Referer header referring to a secure resource (${referer.trim()}) was sent to a different origin (${requestOrigin}). A user agent SHOULD NOT send the Referer cross-origin from a secure referring resource unless that resource explicitly allows it.`,
```
matthyk:
```suggestion
              message: `A Referer header referring to a secure resource (${referer.trim()}) was sent to a different origin (${requestOrigin}).`,
```

packages/rules-rfc-9110/src/rules/message-context/request-context/te/sender-must-send-te-connection-option-with-te-header.rule.ts
```diff
@@ -1,19 +1,60 @@
...
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#name-te')
   .description(
     'A sender of TE MUST also send a "TE" connection option within the Connection header field to inform intermediaries not to forward this field.',
   )
-  .appliesTo('client')
-  .rule((ctx) =>
+  .appliesTo('client', 'user-agent')
```
matthyk:
```suggestion
```

packages/rules-rfc-9110/src/rules/message-context/request-context/user-agent/sender-must-not-generate-advertising-in-product-identifier.rule.ts
matthyk:
```suggestion
```

packages/rules-rfc-9110/src/rules/message-context/request-context/user-agent/sender-should-not-generate-non-version-info-in-product-version.rule.ts
matthyk:
```suggestion
```

matthyk (changes requested):
The rule violation messages should not be too long, as they are printed in the CLI and will result in line breaks

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-must-not-include-fragment-or-userinfo-in-referer.rule.ts
```diff
@@ -19,15 +18,23 @@ export default httpRule(
...
         // Check for fragment (#)
         if (referer.includes('#')) {
-          return [{ location, violation: {}, findings: [] }];
+          return [
+            {
+              location,
+              violation: {
+                message: `The Referer header value "${referer}" contains a fragment component. A user agent MUST NOT include the fragment when generating the Referer field value.`,
+              },
+              findings: [],
```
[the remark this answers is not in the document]
BaggersIO: Keeping the Referer value in the message text was an explicit human-review decision - Matthias' applied suggestions include the value. Findings are reported to the API owner analyzing their own recorded traffic, not to a shared log, and the value is what makes the finding actionable.

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-must-not-include-fragment-or-userinfo-in-referer.rule.ts
```diff
@@ -36,13 +43,29 @@ export default httpRule(
...
           // URL.username or URL.password being non-empty indicates userinfo presence
           if (url.username || url.password) {
-            return [{ location, violation: {}, findings: [] }];
+            return [
+              {
+                location,
+                violation: {
+                  message: `The Referer header value "${referer}" contains a userinfo component (credentials). A user agent MUST NOT include userinfo when generating the Referer field value.`,
+                },
+                findings: [],
```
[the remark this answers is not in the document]
BaggersIO: Keeping the Referer value in the message text was an explicit human-review decision - Matthias' applied suggestions include the value. Findings are reported to the API owner analyzing their own recorded traffic, not to a shared log, and the value is what makes the finding actionable.

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-must-not-include-fragment-or-userinfo-in-referer.rule.ts
```diff
@@ -36,13 +43,29 @@ export default httpRule(
...
           const match = referer.match(/^[^:]+:\/\/([^/]+)/);
           if (match && match[1]?.includes('@')) {
-            return [{ location, violation: {}, findings: [] }];
+            return [
+              {
+                location,
+                violation: {
+                  message: `The Referer header value "${referer}" appears to contain a userinfo component (credentials). A user agent MUST NOT include userinfo when generating the Referer field value.`,
+                },
+                findings: [],
```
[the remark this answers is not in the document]
BaggersIO: Keeping the Referer value in the message text was an explicit human-review decision - Matthias' applied suggestions include the value. Findings are reported to the API owner analyzing their own recorded traffic, not to a shared log, and the value is what makes the finding actionable.

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-must-not-include-fragment-or-userinfo-in-referer.rule.ts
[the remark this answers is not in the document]
BaggersIO: Keeping the Referer value in the message text was an explicit human-review decision - Matthias' applied suggestions include the value. Findings are reported to the API owner analyzing their own recorded traffic, not to a shared log, and the value is what makes the finding actionable.

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-must-not-include-fragment-or-userinfo-in-referer.rule.ts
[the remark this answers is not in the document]
BaggersIO: Keeping the Referer value in the message text was an explicit human-review decision - Matthias' applied suggestions include the value. Findings are reported to the API owner analyzing their own recorded traffic, not to a shared log, and the value is what makes the finding actionable.

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-must-not-include-fragment-or-userinfo-in-referer.rule.ts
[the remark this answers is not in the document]
BaggersIO: Keeping the Referer value in the message text was an explicit human-review decision - Matthias' applied suggestions include the value. Findings are reported to the API owner analyzing their own recorded traffic, not to a shared log, and the value is what makes the finding actionable.

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-must-not-send-referer-in-unsecured-request-from-secure-resource.rule.ts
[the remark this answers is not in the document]
BaggersIO: Keeping the Referer value in the message text was an explicit human-review decision - Matthias' applied suggestions include the value. Findings are reported to the API owner analyzing their own recorded traffic, not to a shared log, and the value is what makes the finding actionable.

packages/rules-rfc-9110/src/rules/message-context/request-context/referer/user-agent-should-not-send-referer-for-secure-to-insecure.rule.ts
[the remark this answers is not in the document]
BaggersIO: Keeping the Referer value in the message text was an explicit human-review decision - Matthias' applied suggestions include the value. Findings are reported to the API owner analyzing their own recorded traffic, not to a shared log, and the value is what makes the finding actionable.

packages/rules-rfc-9110/src/rules/message-context/request-context/expect/client-must-not-generate-100-continue-without-content.rule.ts
```diff
@@ -1,19 +1,57 @@
...
+  // Static context sees only header NAMES / body presence, so it can only
+  // coarsely prefilter to requests that carry Expect without content. It
+  // cannot tell whether the Expect VALUE is exactly "100-continue"; that
+  // value check is only possible against real recorded traffic in analytics.
+  .overrideStaticRule((ctx) =>
     ctx.validateCommonHttpTransactions(
-      and(requestHeader('expect', '100-continue'), not(hasRequestBody())),
+      and(requestHeader('expect'), not(hasRequestBody())),
+    ),
+  )
```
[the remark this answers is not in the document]
BaggersIO: Resolved by removing the static slot entirely (per Matthias' review): the rule is analytics-only now, and since 04ff471e the analyze check is value-aware (case-insensitive, list-aware) rather than name-based.

packages/rules-rfc-9110/src/rules/message-context/request-context/expect/client-must-not-generate-100-continue-without-content.rule.ts
```diff
@@ -5,14 +5,14 @@ export default httpRule(
...
   .description(
     'A client MUST NOT generate a 100-continue expectation in a request that does not include content.',
   )
   .appliesTo('client')
-  .rule((ctx) =>
-    ctx.validateCommonHttpTransactions(
+  .overrideAnalyticsRule((ctx) =>
+    ctx.validateHttpTransactions(
       and(requestHeader('expect', '100-continue'), not(hasRequestBody())),
     ),
```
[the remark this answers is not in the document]
BaggersIO: Two corrections: `requestHeader(name, value)` DOES match the header value in the analyzer (compiled to `reqHeader.value = ?` in the SQL filter - see `plugin-http-analyzer/src/db/http-filter-to-where-clause.ts`), and `appliesTo('client')` is an umbrella role that expands to `client` + `user-agent` (`expandHttpParticipantRoles`), so HAR requests are matched. The real limitation - the exact, case-sensitive value compare - is fixed in 04ff471e: the filter is now name-presence and the 100-continue check happens case-insensitively and list-aware in the validation callback.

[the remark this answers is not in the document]
BaggersIO: Right about the exact/case-sensitive value match - fixed in 04ff471e: the SQL filter is now header-presence only and the callback checks the Expect list case-insensitively, so `100-Continue` and `100-continue, foo` are caught. The role concern does not apply: `appliesTo('client')` expands to `client` + `user-agent` via `expandHttpParticipantRoles`, so HAR requests (default role `user-agent`) are matched.

packages/rules-rfc-9110/src/rules/message-context/request-context/te/sender-must-send-te-connection-option-with-te-header.rule.ts
```diff
@@ -1,19 +1,59 @@
...
+  // Static context sees only header NAMES, so it can only assert structurally
+  // that a request carrying TE also carries a Connection header. The RFC
+  // requirement is on the Connection VALUE (it must list the "te" option),
+  // which is only checkable against real recorded traffic in analytics below.
+  .overrideStaticRule((ctx) =>
     ctx.validateCommonHttpTransactions(
       and(requestHeader('te'), not(requestHeader('connection'))),
     ),
   )
+  .overrideAnalyticsRule((ctx) =>
```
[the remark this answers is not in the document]
BaggersIO: Dropping `appliesTo` here was Matthias' explicit suggestion - the requirement binds any *sender* of the field, which is not role-specific, so the rule intentionally runs unscoped on captured traffic.

packages/rules-rfc-9110/src/rules/message-context/request-context/user-agent/sender-must-not-generate-advertising-in-product-identifier.rule.ts
```diff
@@ -16,11 +15,10 @@ export default httpRule(
   .description(
     'A sender MUST NOT generate advertising or other nonessential information within the product identifier.',
   )
-  .appliesTo('client')
   .overrideAnalyticsRule((ctx) =>
     ctx.validateHttpTransactions(
       and(requestHeader('user-agent')),
-      (request, _res: HttpResponse, location: RuleViolationLocation) => {
+      (request, _res, location: RuleViolationLocation) => {
```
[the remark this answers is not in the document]
BaggersIO: Dropping `appliesTo` here was Matthias' explicit suggestion - the requirement binds any *sender* of the field, which is not role-specific, so the rule intentionally runs unscoped on captured traffic.

packages/rules-rfc-9110/src/rules/message-context/request-context/user-agent/sender-should-not-generate-non-version-info-in-product-version.rule.ts
```diff
@@ -16,11 +15,10 @@ export default httpRule(
   .description(
     'A sender SHOULD NOT generate information in product-version that is not a version identifier (i.e., successive versions of the same product name ought to differ only in the product-version portion of the product identifier).',
   )
-  .appliesTo('client')
   .overrideAnalyticsRule((ctx) =>
     ctx.validateHttpTransactions(
       and(requestHeader('user-agent')),
-      (request, _res: HttpResponse, location: RuleViolationLocation) => {
+      (request, _res, location: RuleViolationLocation) => {
```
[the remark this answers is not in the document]
BaggersIO: Dropping `appliesTo` here was Matthias' explicit suggestion - the requirement binds any *sender* of the field, which is not role-specific, so the rule intentionally runs unscoped on captured traffic.

packages/rules-rfc-9110/src/rules/message-context/response-context/allow/origin-sever-may-send-allow-header.rule.ts
```diff
@@ -1,13 +1,11 @@
-import { not, responseHeader } from '@thymian/core';
 import { httpRule } from '@thymian/core';
 
 export default httpRule('rfc9110/origin-sever-may-send-allow-header')
```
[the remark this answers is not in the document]
BaggersIO: Renamed (file + rule id) in 04ff471e / 5ed376da. The id was not introduced by this PR - the same typo also pre-exists in the 405 and CONNECT rules on main; the 405 one is fixed in #306 and the CONNECT one is left for a follow-up to avoid churning #309.

packages/rules-rfc-9110/src/rules/message-context/request-context/user-agent/user-agent-should-send-user-agent-header.rule.ts
```diff
@@ -10,6 +10,17 @@ export default httpRule('rfc9110/user-agent-should-send-user-agent-header')
...
+      (_req, _res, location: RuleViolationLocation) => [
+        {
+          location,
+          violation: {
+            message: 'The request does not carry a User-Agent header field.',
+          },
+          findings: [],
+        },
+      ],
+    ),
```
matthyk:
```suggestion
    ctx.validateCommonHttpTransactions(not(requestHeader('user-agent'))),
```

packages/rules-rfc-9110/src/rules/message-context/response-context/server/origin-server-may-generate-server-header-field.rule.ts
```diff
@@ -1,17 +1,15 @@
...
 import { httpRule } from '@thymian/core';
 
 export default httpRule(
   'rfc9110/origin-server-may-generate-server-header-field',
 )
   .severity('hint')
-  .type('static', 'analytics', 'test')
+  // This is a permission (MAY generate Server), not a testable constraint;
+  // there is no violating behaviour to detect.
+  .type('informational')
```
matthyk:
```suggestion
  .type('static', 'analytics', 'test')
```

packages/rules-rfc-9110/src/rules/message-context/response-context/server/origin-server-may-generate-server-header-field.rule.ts
```diff
@@ -1,17 +1,15 @@
...
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#name-server')
   .description(
     'An origin server MAY generate a Server header field in its responses.',
   )
   .appliesTo('origin server')
```
matthyk, line 13:
```suggestion
  .appliesTo('origin server')
  .rule((ctx) =>
    ctx.validateCommonHttpTransactions(not(responseHeader('server'))),
  )
```

packages/rules-rfc-9110/src/rules/message-context/request-context/expect/server-may-respond-with-417-response-for-other-expect-than-100-continue.rule.ts
```diff
@@ -1,23 +1,15 @@
...
 import { httpRule } from '@thymian/core';
 
 export default httpRule(
   'rfc9110/server-may-respond-with-417-response-for-other-expect-than-100-continue',
 )
   .severity('hint')
-  .type('analytics')
+  // This is a permission (MAY respond with 417), not a testable constraint;
+  // there is no violating behaviour to detect from traffic.
+  .type('informational')
```
matthyk:
```suggestion
  .type('analytics')
```

packages/rules-rfc-9110/src/rules/message-context/request-context/expect/server-may-respond-with-417-response-for-other-expect-than-100-continue.rule.ts
```diff
@@ -1,23 +1,15 @@
...
   .url('https://www.rfc-editor.org/rfc/rfc9110.html#name-expect')
   .description(
     'A server that receives an Expect field value containing a member other than 100-continue MAY respond with a 417 (Expectation Failed) status code to indicate that the unexpected expectation cannot be met.',
   )
   .appliesTo('server')
```
matthyk, line 13:
```suggestion
  .appliesTo('server')
  .rule((ctx) =>
    ctx.validateHttpTransactions(
      and(
        requestHeader('expect'),
        not(requestHeader('expect', '100-continue')),
        not(statusCode(417)),
      ),
    ),
  )
```

matthyk (changes requested):

matthyk (approved):

PR #316, opened 2026-07-09
fix(plugin-http-analyzer): expand umbrella appliesTo roles to concrete participant roles

BaggersIO:
## What & why

`appliesTo` role filtering matched stored role names exactly (`reqRole.name IN (...) OR resRole.name IN (...)`), with `intermediary` as the only expanded alias. Rules tagged `server` or `client` therefore silently skipped traffic labeled with concrete roles such as `origin server` or `user-agent` — notably **all HAR imports**, which default to `user-agent` (request) and `origin server` (response).

Fixes thymianofficial/thymian-internal#355.

## The fix

- **`@thymian/core`**: new shared `expandHttpParticipantRoles()` next to the role list in `rule-meta.ts`, encoding the role hierarchy:
  - `server` → `server`, `origin server`, `cache`, `proxy`, `gateway`, `tunnel`, `intermediary` (every participant that can receive and respond to requests, per the issue)
  - `client` → `client`, `user-agent` (request originators)
  - `intermediary` → `intermediary`, `proxy`, `gateway`, `tunnel` (unchanged behavior, moved out of the analyzer)
  - concrete roles pass through; overlapping expansions are deduplicated
- **`@thymian/plugin-http-analyzer`**: `AnalyticsApiContext` now calls the shared helper instead of its local `intermediary` special case. Repository SQL filtering stays exact-match against the already-expanded list, as proposed in the issue.

Note beyond the issue's write-up: the `client` → `user-agent` expansion is included for the same reason as `server` → `origin server` — without it, request-side rules tagged `client` never fire on HAR imports. (Left `client` NOT expanding to intermediary roles: a proxy does act as a client on its outbound leg, but captured roles describe the participant, not the leg, and widening `client` to intermediaries would make user-agent-specific rules fire on proxies.)

## Tests

- `packages/core/test/rules/expand-http-participant-roles.test.ts`: expansion table per role, server excludes client-side roles, dedup, passthrough.
- `packages/plugin-http-analyzer/test/analytics-api-context.test.ts`: end-to-end through the SQLite repository — `['server']` matches transactions labeled `origin server`/`cache`/`gateway` but not `client`/`user-agent`; `['client']` matches the HAR-shaped `user-agent` request leg. Existing `intermediary` test unchanged and green.

## Impact

Unblocks simplifying rule metadata to single umbrella roles (e.g. `.appliesTo('server')` instead of `.appliesTo('origin server', 'server')`) — wanted by the review on #300.

## Verification

- `nx build core`, `nx build plugin-http-analyzer`: pass
- Full vitest suites: core 186/186, plugin-http-analyzer 91/91
- Prettier/eslint via pre-commit hooks: clean

### DIFF
--- a/packages/core/src/rules/rule-meta.ts
+++ b/packages/core/src/rules/rule-meta.ts
@@ -15,6 +15,37 @@ export const httpParticipantRoles = [
 
 export type HttpParticipantRole = (typeof httpParticipantRoles)[number];
 
+// Captured traffic stores one concrete role per message and repositories match
+// role names exactly, so umbrella roles must be expanded to every concrete
+// role they subsume before filtering. 'server' covers every participant that
+// can receive and respond to requests; 'client' covers every participant that
+// originates requests.
+const httpParticipantRoleHierarchy: Partial<
+  Record<HttpParticipantRole, readonly HttpParticipantRole[]>
+> = {
+  server: [
+    'server',
+    'origin server',
+    'cache',
+    'proxy',
+    'gateway',
+    'tunnel',
+    'intermediary',
+  ],
+  client: ['client', 'user-agent'],
+  intermediary: ['intermediary', 'proxy', 'gateway', 'tunnel'],
+};
+
+export function expandHttpParticipantRoles(
+  roles: readonly HttpParticipantRole[],
+): HttpParticipantRole[] {
+  return [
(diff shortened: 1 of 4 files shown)


matthyk (approved):
If you work in the copilot changes, it looks good.

PR #318, opened 2026-07-15
Adapt CLI rendering to new report format

matthyk:
This PR updates the CLI output to the new report format. It now renders also multi-step test cases while sticking to the same overall layout for the `lint`, `test` and `analyze` commands. See screenshot below:

## Lint

Screenshot 2026-07-15 at 13 26 26

## Test

Screenshot 2026-07-15 at 13 25 01

I still see a problem in the length of the rule message. Obviously, it would be an easy solution to just truncate the messages, but for such technical and precise message, I don't think this is a good solution. I see this as a follow-up. But also happy to discuss.

Closes https://github.com/thymianofficial/thymian-internal/issues/335


packages/common-cli/src/render/create-execution-renderer.ts
```diff
@@ -0,0 +1,105 @@
...
+          rule?.summary?.text ?? rule?.description?.text ?? rule?.name,
+        );
+
+        lines.push(indent(indentationLevel + 1) + statusLine);
+
+        if (rule) {
+          lines.push(
+            indent(indentationLevel + 6) + ux.colorize('dim', '› ' + rule.id),
+          );
+        }
```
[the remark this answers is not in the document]
matthyk: Keeping as-is: `rule` is looked up *from* `execution.ruleId` via the rule index, so when the descriptor is missing `rule.id` would equal `execution.ruleId` — guarding on `rule` hides no additional identifying information. Added a code comment explaining this.

packages/common-cli/src/render/cli-report.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export function renderReport(
+  report: Report,
+  options: { format?: ThymianFormat } = {},
+): string {
+  if (report.runs.length === 0) {
+    return 'No tool runs were reported.';
+  }
+
+  const lines: string[] = [''];
```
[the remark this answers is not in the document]
matthyk: Intended: the leading blank line separates the command prompt from the report output.

packages/common-cli/src/render/findings.ts
```diff
@@ -0,0 +1,60 @@
...
+          `${indent(indentationLevel + 2)}actual: ${truncate(JSON.stringify(actual))}`,
+        );
+      }
+
+      return lines;
+    }
+    case 'rule-violation':
+      // Rule identity and outcome render at the execution level, so the
+      // rule-violation finding itself has no additional body to render.
+      return [];
```
BaggersIO: **[Medium] Per-step violation messages are dropped for multi-step test cases.**

The comment says rule identity/outcome render at the execution level. That holds for **lint/analyze** — their message goes into `status.reason` (`report-builder.ts:154,187`) which `renderStatus` prints. It does **not** hold for **tests**:

- `plugin-http-tester` stores each *step's* specific message as the `title` of a step-level `rule-violation` finding (`plugin-http-tester/src/index.ts:51-66`, pushed in `buildTestStep` at line 81).
- The execution's `status.reason` only receives the **first** violation message — `signals.violationReason` is first-wins (`index.ts:222-224`).

So in a multi-step / multi-violation test case, only the first reason survives (on the status line); every other step's reason is silently lost. The removed renderer printed `finding.title` for every kind, so this is a regression — in exactly the multi-step feature this PR adds.

**Suggestion:** render the `rule-violation` finding's `title` (and `message.text` if present) instead of `[]`, and add a multi-step test whose steps carry `rule-violation` findings (the current multi-step test only uses `assertion-*` findings, so it doesn't catch this).

packages/common-cli/src/render/findings.ts
```diff
@@ -0,0 +1,60 @@
...
+    default:
+      // Superseded/unknown finding kinds are intentionally not rendered.
+      return [];
+  }
+}
+
+export function truncate(str: string, maxLength = 30): string {
```
BaggersIO: **[Low-Medium] `expected`/`actual` truncated to ~26 chars.**

`truncate` (default `maxLength = 30`) is applied to `JSON.stringify(expected/actual)` at lines 37/40. The removed renderer emitted these in full via core's `findingDetails`. Any object or longish string value is clipped to 26 chars + `" ..."` (e.g. `{"name":"Fluffy","specie ...`), which hides the very diagnostic the expected/actual lines exist to show.

The PR description flags rule-*message* length as a follow-up — this is the same problem in assertion detail. Suggest not truncating these values (or a much larger cap).

packages/common-cli/src/render/findings.ts
```diff
@@ -0,0 +1,60 @@
...
+export function renderFindings(
+  findings: FindingRecord[],
+  indentationLevel: number,
+): string[] {
+  return findings.flatMap((finding) =>
+    renderFinding(finding, indentationLevel),
+  );
+}
+
+function renderFinding(
```
BaggersIO, line 21: **[Low] `finding.message.text` is never rendered.**

`renderFinding` shows only `title` (+ expected/actual). The removed renderer also printed `message.text` when it differed from the title. For `assertion-failure`, `title` is the short assertion label and `message` is the human explanation (`report-builder.ts:324-325`), so some context is lost. Lower impact (informational titles already equal their message) — flagging so the drop is a conscious decision. Related to the `rule-violation` finding above.

BaggersIO (commented):
Reviewed the CLI renderer rewrite — clean refactor with good happy-path coverage. Four inline findings, all about behavior parity with the removed `cli-report-renderer.ts`, none overlapping the existing bot comments:

1. **[Medium]** per-step `rule-violation` messages are dropped for multi-step test cases (regression in the feature this PR adds).
2. **[Low-Medium]** `expected`/`actual` are truncated to ~26 chars, hiding assertion diagnostics.
3. **[Low]** `finding.message.text` is no longer rendered.
4. **[Low]** minor simplification + a lost explanatory comment.

Details inline.

_Note: I verified separately that the `filter((e) => isLintExecution(e) || isAnalyzeExecution(e))` call type-checks — TS 5.9 infers a type predicate through the `||`, so it narrows to `(LintExecution | AnalyzeExecution)[]`._

packages/common-cli/src/render/utils.ts
```diff
@@ -0,0 +1,19 @@
+const SINGLE_INDENTATION = '  ';
+
+export function indent(times: number): string {
+  return Array.from({ length: times })
```
BaggersIO: **[Low / nits]**

- `indent()` here and the header rule in `cli-report.ts` build strings via `Array.from({ length }).map(...).join('')`; `SINGLE_INDENTATION.repeat(times)` and `'─'.repeat(...)` are simpler and cheaper.
- Separately: `collectSeverityCounts` in `cli-report.ts` dropped the old renderer's note explaining that the summary counts *failed executions by resolved severity*, not `informational` findings (so the info count can read 0 while info findings render). Worth keeping the technical part of that note (minus any PR/triage references).

packages/common-cli/src/render/findings.ts
```diff
@@ -0,0 +1,61 @@
...
+      const { expected, actual } = finding as ReportAssertionFailure;
+
+      if (actual !== undefined && expected !== undefined) {
+        lines.push(
+          `${indent(indentationLevel + 2)}expected: ${JSON.stringify(expected)}`,
+        );
+        lines.push(
+          `${indent(indentationLevel + 2)}actual: ${JSON.stringify(actual)}`,
+        );
+      }
```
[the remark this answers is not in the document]
matthyk: intended

packages/core/src/report/finding-render.ts
```diff
@@ -44,7 +40,7 @@ export function buildRuleIndex(
 export function resolveExecutionSeverity(
   execution: Execution,
   ruleIndex: ReadonlyMap<string, RuleDescriptor>,
-  logger?: SeverityWarnLogger,
+  logger?: Logger,
 ): Severity | undefined {
```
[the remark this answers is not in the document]
matthyk: Let's stick with a single `Logger` type

BaggersIO (commented):
Re-review of `b2df713f`. Confirmed resolved since my last round: per-step rule-violation messages for multi-step tests, untruncated `expected`/`actual`, and the `indent`/header `.repeat()` nits — plus Copilot's prototype-pollution grouping and `thymianFormat` clobber fixes. CI is green across lint/typecheck/e2e/CodeQL.

One residual edge and a test gap remain, both inline. Not re-filing items already triaged by you or Copilot (single `Logger` type, assertion `both-present`, hex-color false positive).

packages/common-cli/src/render/test-executions.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const groupTestCaseExecutions: GroupExecutionsFn<TestCaseExecution> = (
+  execution,
+) => execution.name;
+
+export const renderTestCaseExecution: RenderExecutionDetailsFn<
+  TestCaseExecution
+> = (execution, indentationLevel, locationResolver, toolRun) => {
+  if (execution.steps[0] && execution.steps.length < 2) {
+    return renderFindings(execution.steps[0].findings, indentationLevel);
```
BaggersIO, line 18: **[Low-Med] Single-step test cases with >1 violation still surface only the first.**

This single-step branch calls `renderFindings(...)` **without** `renderRuleViolationTitle: true`, so every `rule-violation` finding in the step renders as `[]`. The rationale (per the `findings.ts` comment) is that the status line already shows the reason — but the tester sets `status.reason` to `signals.violationReason`, the **first** violation message only (`plugin-http-tester/src/index.ts:222-224`).

**Failure case:** a one-step test whose rule reports two violations at that step — the status line shows the first, the second renders nowhere. Same info-loss class as the multi-step issue this PR just fixed, narrowed to the single-step path.

Trade-off: passing `renderRuleViolationTitle: true` here duplicates the first message (status line + body) for the common single-violation case; the alternative is to render violations only in the body for tests and drop them from the status line. Either way worth a conscious decision — Low-Med since it needs one step + multiple violations from one rule.

packages/common-cli/src/render/test-executions.ts
```diff
@@ -0,0 +1,37 @@
...
+  const lines: string[] = [];
+
+  const lastStepIdx = execution.steps.length - 1;
+  for (const [idx, step] of execution.steps.entries()) {
+    lines.push(indent(indentationLevel) + '│');
+    lines.push(
+      `${indent(indentationLevel)}${idx === lastStepIdx ? '└──' : '├──'} ${step.name}: ${locationResolver(step.location, toolRun.thymianFormatVersion)}`,
+    );
+    lines.push(
+      ...renderFindings(step.findings, indentationLevel + 3, {
```
BaggersIO, line 30: **[Low] This `renderRuleViolationTitle` path has no test.**

The fix for dropped multi-step violations is exercised by no test — the `renders multi-step test executions as a step tree` case uses only `assertion-*` findings, never a `rule-violation`. A multi-step test with `rule-violation` step findings would lock this behavior in (and would also cover the single-step edge noted above).

BaggersIO (approved):

packages/common-cli/src/render/test-executions.ts
matthyk: An execution should contain only contain rule violations from a single rule by design. If we now have a single step test case which includes 2 rule violations, I do not see any benefit to display the same message/rule violation twice. It would be impossible to resolve one rule violation but not the other. So this behavior is intended.

PR #319, opened 2026-07-20
feat(plugin-sampler): add validate command for generated samples

matthyk:
## Summary

Adds `thymian sampler validate` — a read-only command that verifies the
generated sampler artifacts committed in the repo still match what the
sampler would produce for the current API specification. It never writes
files, exits non-zero on any drift, and gives file-specific diagnostics,
so it works for local pre-commit checks and CI enforcement.

## How it works

- Computes the *expected* output in-memory by reusing the real generation
  path (`generateSamplesForThymianFormat` + type generation +
  `writeSamplesToDir`), then diffs it against the files on disk. To keep
  this side-effect-free, `writeSamplesToDir` gained injectable
  `mkdir`/`writeToFile` seams — the validator captures output into a map
  instead of touching the filesystem.
- Classifies each mismatch: `missing-artifact`, `changed-artifact`,
  `stale-root-metadata`, `metadata-out-of-sync`, `unexpected-artifact`,
  `invalid-json`. JSON artifacts get pointer-level diffs via
  `@scalar/json-magic`; timestamp-only root-metadata differences are
  ignored to avoid false positives.
- Diagnostics reuse the core report styling (`SEVERITY_COLORS` /
  `SEVERITY_SYMBOLS`) so output is consistent with the rest of the CLI —
  **without** constructing a `Report`.

## Command surface

| Invocation | Behavior |
|---|---|
| `sampler validate` | Summary — lists each out-of-sync artifact (`<type>: <path>`) and a checked/failed count. |
| `sampler validate --full-diffs` | Expands every out-of-sync artifact to its full diff. |
| `sampler validate --for-path <path>` | Scopes the whole report (output + exit code) to one artifact; errors if the path is unknown. |
| `sampler validate --json` | Machine-readable report on stdout for CI/agents; exit code preserved. |

Exit codes: `0` when in sync, `1` on drift, `2` on usage errors (e.g.
unknown `--for-path`).

## Notes / decisions

- **No fix-command hint** (e.g. "run `sampler generate`") in the output —
  intentional; that guidance is handled as a follow-up, not by this command.
- **Human vs. machine split:** `--json` covers agents/CI; `--full-diffs`
  and `--for-path` cover human inspection. The default stays a compact
  summary rather than dumping diffs.
- Hook-file detection is shared with the sampler's actual hook loader
  (single `isHookFileName` predicate) so validate can't flag a file the
  sampler loads as a hook.

## Testing

- Unit tests for the validation logic (in-sync, missing/changed/unexpected,
  stale/nested metadata, invalid JSON, `--for-path` scoping, full content).
- E2E tests covering the CLI surface: default summary, `--full-diffs`,
  `--for-path` (scoped + unknown), and both `--json` cases.
- Verified manually against the built CLI.

### DIFF
--- a/e2e-tests/src/cli-sampler-validate.test.ts
+++ b/e2e-tests/src/cli-sampler-validate.test.ts
@@ -45,7 +45,7 @@ describe('thymian sampler validate', () => {
     expect(result.exitCode).toBe(1);
     expect(result.output).toContain('changed-artifact: types.d.ts');
     expect(result.output).toMatch(
-      /Checked \d+ generated artifacts in .+\. \d+ failed\./,
+      /Checked \d+ generated artifacts in .+\. \d+ out of sync\./,
     );
     expect(result.output).not.toContain('sampler init');
     expect(result.output).not.toContain('Re-run:');
@@ -169,6 +169,22 @@ describe('thymian sampler validate', () => {
     expect(result.output).toContain('does-not-exist.json');
   }, 180_000);
 
+  it('emits a JSON report and exit 2 when --json --for-path targets an unknown artifact', () => {
+    copyFixturesToTempDir(join(fixturesDir, 'dynamic-test'), getTempDir());
+    execThymian(['sampler', 'init'], { cwd: getTempDir() });
+
+    const result = execThymianRaw(
+      ['sampler', 'validate', '--json', '--for-path', 'does-not-exist.json'],
+      { cwd: getTempDir(), allowFailure: true },
+    );
+
+    expect(result.exitCode).toBe(2);
+
+    const report = JSON.parse(result.stdout);
+    expect(report.checkedArtifacts).toBe(0);
+    expect(report.failures).toEqual([]);
+  }, 180_000);
+
   it('emits a machine-readable report and non-zero exit with --json', () => {
     copyFixturesToTempDir(join(fixturesDir, 'dynamic-test'), getTempDir());
     execThymian(['sampler', 'init'], { cwd: getTempDir() });
@@ -262,8 +278,8 @@ describe('thymian sampler validate', () => {
     });
 
     expect(result.exitCode).toBe(0);
-    expect(result.output).toContain(
-      'Sampler validation passed: 5 generated artifacts are in sync.',
+    expect(result.output).toMatch(
+      /Sampler validation passed: \d+ generated artifacts are in sync\./,
     );
   }, 180_000);
 });
--- a/e2e-tests/src/cli-sampler-validate.test.ts
+++ b/e2e-tests/src/cli-sampler-validate.test.ts
@@ -1,0 +1,269 @@
+import { mkdirSync, readFileSync, unlinkSync, writeFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+import { describe, expect, it } from 'vitest';
+
+import {
+  copyFixturesToTempDir,
+  execThymian,
+  execThymianRaw,
+  fixturesDir,
+  useTempDir,
+} from './helpers.js';
+
+describe('thymian sampler validate', () => {
+  const getTempDir = useTempDir();
+
+  it('passes when generated sampler artifacts are in sync', () => {
+    copyFixturesToTempDir(join(fixturesDir, 'dynamic-test'), getTempDir());
+    execThymian(['sampler', 'init'], { cwd: getTempDir() });
+
+    const result = execThymianRaw(['sampler', 'validate'], {
+      cwd: getTempDir(),
+    });
+
+    expect(result.exitCode).toBe(0);
+    expect(result.output).toMatch(
+      /Sampler validation passed: \d+ generated artifacts are in sync\./,
+    );
+  }, 180_000);
+
+  it('fails with file-specific diagnostics when generated artifacts drift', () => {
+    copyFixturesToTempDir(join(fixturesDir, 'dynamic-test'), getTempDir());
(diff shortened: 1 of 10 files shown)


BaggersIO (commented):
## Review — `sampler validate`

Solid, well-tested, and correctly read-only. I traced the findings against the actual generation path rather than the diff alone. **No blocking correctness bugs** — everything below is maintainability/cleanup-level. A few concerns I initially suspected (compact-`meta.json` re-serialization, nested-metadata timestamp false positives, `types.d.ts`/`tsconfig.json` fidelity) turned out to be non-issues once traced; noted under "Verified" so the reasoning is on record.

### 🟡 Medium

**1. Incomplete write-reuse: `types.d.ts` and `tsconfig.json` are hand-written in two places and can silently drift**

`src/index.ts:254-262` (real `sampler.init`) and `src/validation/validate-sampler-output.ts:830-838` (`createExpectedSamplerFiles`) each independently emit these two artifacts — currently identically (`generatedTypesToString(...)` and `JSON.stringify(tsConfig, null, 2)`). The design goal is "reuse the real generation path so expected can't drift," yet these two files sit *outside* `writeSamplesToDir`. If `init` later changes how either is written (e.g. a trailing newline), `validate` will emit false `changed-artifact` failures with no compile-time signal.

Suggestion: move both writes into `writeSamplesToDir` (or a shared helper both callers use) so the reuse guarantee is complete. This is the one item I'd genuinely like addressed.

### 🟢 Low

**2. Dead import** — `src/cli/commands/sampler/validate.ts:3`: `errorSymbol` is imported but never used (`successSymbol`, `SEVERITY_COLORS`, `SEVERITY_SYMBOLS`, `Severity` are). Lint passed CI, but it's dead — remove it.

**3. Inert eslint config** — `eslint.config.mjs`: `'ajv'` is added to `ignoredDependencies`, but `ajv` is not in `package.json` and is not imported anywhere in the package — it appears only inside a comment (`index.ts:166-168`). Looks like a leftover from an earlier `JSONSchemaType`-from-`ajv` approach replaced by `as never`. Remove the `'ajv'` entry.

**4. `checkedArtifacts` omits unexpected files, so the summary can mislead** — `validate-sampler-output.ts:906-909` sets `checkedArtifacts = expectedFiles.size`, but `failures` includes `unexpected-artifact` entries. With 5 in-sync files + 2 orphans, the summary (`validate.ts:451-453`) reads "Checked 5 generated artifacts… 2 failed," which reads as "2 of 5 failed" when 0 of 5 failed and 2 are extra. Also `unexpected-artifact` is styled `warn` (yellow) yet still forces exit 1. Consider counting extras into the total or rewording. (Exit-1-on-extra is correct for drift detection — only the phrasing/count is off.)

**5. `as never` schema cast is blunter than the codebase norm** — `index.ts:170`: `samplerValidateActionSchema as never` disables *all* structural checking of the schema literal, so a wrong response-schema shape wouldn't be caught. The sibling `sampler.init` uses `@ts-expect-error` (`index.ts:145-146`) and core uses `{} as JSONSchemaType<...>`. Prefer `as unknown as JSONSchemaType<...>` for parity and to keep some checking. (Minor: schema uses `nullable: true` for `forPath`/`expected`/`actual`, but the TS types are optional `?`, not `| null`, and the code omits rather than nulls them — harmless mismatch.)

### ⚪ Nits

**6.** `--for-path` scopes output but not work — the full expected set is always regenerated even when scoping to one artifact. Inherent to holistic generation and fine, but worth a doc note if users expect it to be faster for pre-commit use.

**7.** `--json` + unknown `--for-path` — `validate.ts:408-415` calls `this.error(..., {exit: 2})` before the `jsonEnabled()` branch, so this usage error isn't emitted as a JSON envelope (unlike drift, which is JSON + exit 1). Defensible for a usage error, but a machine consumer gets non-JSON here, and it's untested.

**8.** Cross-platform `--for-path` — keys are normalized to `/` (`normalizeRelativePath`), so a Windows-style `a\b.json` argument won't match. Doc examples use `/` and Windows CI is green, so likely fine.

**9.** E2E brittleness — the custom-path test asserts exactly `"5 generated artifacts"`; will break if the fixture's artifact count changes.

### ✅ Verified (things that look risky but are correct)

- **Faithful expected output** — reuses the exact generation functions; `types.d.ts`/`tsconfig.json`/samples/root-meta all match `sampler.init` byte-for-byte.
- **Timestamp handling is correct** — root `meta.json` is written *compact* (`write-samples-to-dir.ts:255`), so `preserveRootMetadataTimestamp`'s compact re-serialization is consistent; nested `SamplesNodeMeta` carries *no* timestamp (`samples-tree-structure.ts:96-99`), so no nested false positives. Copying the whole `version` object equals timestamp-only, since the type is exactly `{version, timestamp}`.
- **Genuinely read-only** — no-op `mkdir` + in-memory capture; a test asserts the samples dir isn't created when missing.
- **Shared `isHookFileName`** keeps validate in lockstep with the hook loader (`read-samples-from-dir.ts:198-211`); the ignore set is a safe superset of what's loaded.
- **Strong coverage** — unit + e2e across in-sync, missing/changed/unexpected, stale/nested metadata, invalid JSON, `--for-path` (scoped + unknown), and both `--json` cases.
- **No security concerns** — read-only, JSON-pointer escaping is correct, `forPath` used only for map lookup (no path traversal).

_Note: the `CodeQL — fail` status is the aggregate check context; the actual CodeQL analysis jobs passed and code-scanning alerts are empty._

BaggersIO (approved):

PR #328, opened 2026-07-22
feat(thymian): Stabilize, widen, and harden the new report model

matthyk:
# Stabilize, widen, and harden the new report model

## Summary

This finalizes the new report model: a flat, per-runType execution model with an explicit status, richer schema-validation detail, and formatters adapted to it. On top of the feature work, it includes a round of review-driven correctness fixes to the http-tester → report mapping path.

### Report model (@thymian/core)
- Flat, non-recursive Execution list per run with an explicit status (passed / failed / skipped) instead of nested visit wrappers.
- walkExecutions now yields Execution directly; the ExecutionVisit wrapper type was removed. ⚠️ Breaking for external consumers of the traversal API — all in-repo consumers (reporter, common-cli) are migrated in this PR.
- Wider, clearer schema-error reporting: describeSchemaError names the offending property path (incl. additionalProperties keys, e.g. property "user.extra" is not allowed), and schemaErrorDetail derives expected/actual pairs for renderers.
- validate-* helpers consolidated onto the shared schema-error detail path.

### Formatters (@thymian/plugin-reporter, @thymian/common-cli)
- Markdown / CSV / CLI adapted to the flat execution model and the new finding shape.

## Review hardening (correctness fixes)

Findings from a review of the report-model rework, each with tests:

- Schema-validation failures were dropped from the report — checkBody/checkHeaders/checkResponse failures were pushed without a location, so buildTestStep couldn't tie them to a step and their per-property detail vanished. Now tagged with { stepIdx, transactionIdx } like expectForTransactions. (the one genuine regression in the rework)
- Markdown table corruption — the lint/analyze rule cell interpolated helpUri into a raw [text](url) link; a )/whitespace in the URL broke the row. Unified both report sections onto a shared, escaped HTML-anchor helper.
- Schema boundary mislabeled as expected — maxLength: 3 on "abcd" rendered as expected: 3, actual: "abcd". Boundary keywords now emit no misleading pair (the message already states the bound).
- Failed top-level expect reported as skipped — an assertion failure with no transaction produced no violation signal; now surfaces as failed while genuine execution failures still map to skipped.
- Distinct violation.message dropped — the generic step marker was suppressed whenever failure detail existed, discarding a rule's own rationale; now retained when the violation carries its own message.
- Perf/readability — step placements are grouped once per case instead of re-filtered per step.

## Known limitation / follow-up

- When the same transactionId runs in multiple steps (e.g. a replayed step), an unplaced rule result can only be attached to the first occurrence — the result carries no step index to disambiguate. Documented in-code; tracked in #423.

## Testing

- All touched packages build clean; 588 tests pass across core, plugin-http-tester, plugin-reporter, common-cli.
- New regression tests: schema-validation detail routing, top-level-expect status, violation.message retention, helpUri escaping, schema-boundary labeling.

### DIFF
--- a/packages/core/src/report/report.ts
+++ b/packages/core/src/report/report.ts
@@ -19,6 +19,9 @@ export interface Message {
 /**
  * External or generated file that belongs to a tool run, such as HAR files,
  * screenshots, generated requests, raw logs, or exported diagnostics.
+ *
+ * Reserved public surface: referenced by {@link ToolRun.artifacts} but not yet
+ * populated by any producer. Kept as an intentional extension point.
  */
 export interface Artifact {
   /** Stable identifier within the report. */
@@ -35,21 +38,6 @@ export interface Artifact {
   mimeType?: string;
 }
 
-/** HTTP header bag as represented in report payloads. */
-export type Headers = Record<string, string | string[]>;
-
-/** Generic body representation for future report-owned HTTP payloads/artifacts. */
-export interface BodyContent {
-  /** Textual body content, if it is safe and useful to inline. */
-  text?: string;
-  /** Base64-encoded body content for binary payloads. */
-  binaryBase64?: string;
-  /** Media type of the body content. */
-  mimeType?: string;
-  /** Encoding marker for consumers that need to decode the body. */
-  encoding?: 'plain' | 'base64' | (string & {});
-}
-
 /** HTTP request shape reused from the core HTTP model. */
 export type ReportHttpRequest = HttpRequest;
 
@@ -57,10 +45,12 @@ export type ReportHttpRequest = HttpRequest;
 export type ReportHttpResponse = HttpResponse;
 
 /**
- * Captured HTTP transaction associated with an execution. It is intentionally
- * attached to executions, not findings, because multiple findings can originate
- * from the same transaction and renderers may decide how much transaction detail
- * to show.
+ * Captured HTTP transaction collected while running a test. Transactions are
+ * attached at {@link TestStep} granularity (see {@link TestStep.httpTransactions}),
+ * not to individual findings, because multiple findings can originate from the
+ * same transaction and renderers may decide how much transaction detail to show.
+ * A finding may point at a specific transaction within its step via
+ * {@link BaseFinding.transactionIndex}.
  */
 export interface ReportHttpTransaction {
   /** Captured request. */
@@ -94,7 +84,14 @@ export interface RuleDescriptor {
   }>;
 }
 
-/** Invocation describes one process/command that contributed to a tool run. */
+/**
+ * Invocation describes one process/command that contributed to a tool run.
+ *
+ * Reserved public surface: the model carries it and {@link ToolRun.invocations}
+ * references it, but no producer populates it yet. Kept (not removed) as an
+ * intentional SARIF-parity extension point for capturing CLI/process invocation
+ * detail in a future story.
+ */
 export interface Invocation {
   /** Optional invocation identifier. */
   id?: string;
@@ -127,7 +124,11 @@ export interface Invocation {
 export interface BaseFinding {
   /** Stable identifier within the report. */
   id: string;
-  /** Discriminator describing what kind of fact this finding represents. */
+  /**
+   * Discriminator describing what kind of fact this finding represents. Kept
+   * open (`string & {}`) so producers can emit custom finding kinds; consumers
+   * should handle unknown kinds gracefully.
+   */
   kind:
     | 'rule-violation'
     | 'informational'
@@ -138,6 +139,14 @@ export interface BaseFinding {
   title: string;
   /** Optional longer message. */
   message?: Message;
+  /**
+   * Zero-based index into the owning {@link TestStep.httpTransactions} of the
+   * transaction this finding is about, when it is tied to a specific transaction.
+   * Lets a consumer trace a finding (e.g. an assertion failure or an invalid
+   * transaction) back to the exact HTTP exchange. Undefined when no single
+   * transaction applies (e.g. lint/analyze findings, or step-wide findings).
+   */
+  transactionIndex?: number;
 }
 
 /**
@@ -316,14 +325,16 @@ export interface TestCaseExecution extends ExecutionBase {
  */
 export type Execution = LintExecution | TestCaseExecution | AnalyzeExecution;
 
-/** One run of one tool/plugin for a workflow mode such as lint/test/analyze. */
-export interface ToolRun {
+/**
+ * Fields shared by every {@link ToolRun}, independent of its run type. The
+ * discriminating `runType` and the type-correlated `executions` are added per
+ * arm of the {@link ToolRun} union.
+ */
+export interface ToolRunBase {
   /** Stable identifier for this run within the report. */
   runId: string;
   /** Tool/plugin that produced the data. */
   tool: Tool;
-  /** Workflow/run kind. Closed union — every execution's `kind` matches this. */
-  runType: 'lint' | 'test' | 'analyze';
   /** ISO timestamp when the run was created/started. */
   runAt: string;
   /** Duration in milliseconds, if measured. */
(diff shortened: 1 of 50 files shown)


packages/core/src/report/report.ts
```diff
@@ -35,32 +38,19 @@ export interface Artifact {
...
-  mimeType?: string;
-  /** Encoding marker for consumers that need to decode the body. */
-  encoding?: 'plain' | 'base64' | (string & {});
-}
-
 /** HTTP request shape reused from the core HTTP model. */
 export type ReportHttpRequest = HttpRequest;
 
 /** HTTP response shape reused from the core HTTP model. */
 export type ReportHttpResponse = HttpResponse;
```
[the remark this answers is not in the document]
matthyk: That's fine because we're not yet at a stable version 1.x

packages/plugin-http-tester/src/index.ts
```diff
@@ -184,13 +239,76 @@ function computeTestCaseStatus(
...
+      for (const transaction of step.transactions) {
+        const id = transaction.source?.transactionId;
+        if (id !== undefined && !positionByEdgeId.has(id)) {
+          positionByEdgeId.set(id, { testCaseIndex, stepIndex });
+        }
+      }
+    });
+  });
+
+  const placed = new Set(
```
BaggersIO: **Blocking — cross-entry placement leak: a failure from one `httpTest` call can double-report and flip a passing case from another call to failed.**

`placed` is built from the *current* diagnostics entry's placements only, but `ruleFnResult` contains the rule's results from *all* context calls — and each `ctx.httpTest()` / `ctx.validate*()` call pushes its own diagnostics entry.

Failure scenario: a rule first calls `validateCommonHttpTransactions(...)` (entry A runs `tx-1`, case passes, no violations), then `httpTest(...)` (entry B also runs `tx-1`, an assertion fails → the context records a placement for result R in entry B). When `createRuns` processes entry A first, R is not in A's `placed` set and not yet in `consumedResults`, and its edge id matches `tx-1` in A's test result — so R is synthesized into entry A: the *passing* case is reported `failed` with a violation marker, and entry B then renders R again from its own placement. One failure → two failed cases, one of them false.

The fix is small: build the placed set once across **all** entries (e.g. `diagnostics.flatMap((d) => d.placements)` computed before the entry loop) and pass it in, keeping `consumedResults` as-is. The existing tests don't catch this because they all use single-entry diagnostics — worth a regression test with two entries sharing a transaction id.

packages/plugin-http-tester/src/index.ts
```diff
@@ -66,36 +70,87 @@ function ruleViolationToStepFinding(
...
+    }
+  }
+
+  // A violation with no failure detail still needs a visible marker; when detail
+  // findings already convey the failure, don't add a duplicate generic marker —
+  // UNLESS the violation carries its own message (a distinct rationale that the
+  // detail records don't convey), in which case keep it so it isn't dropped.
+  const violation = stepPlacements.find(
+    ({ result }) => result.violation !== undefined,
+  )?.result.violation;
```
BaggersIO: **Non-blocking — only the first violation per step renders a marker (rendering regression vs. pre-PR behavior).**

Before this PR, every violation-bearing placement produced a marker finding. This `.find(...)` selects a single violation, so when two placements with violations land on the same step — e.g. a `.transactions()` callback returning two violations with distinct messages for the same transaction, or two failing transactions in one grouped step — the second violation's `message` is silently dropped. That's the same bug class as the "distinct violation.message dropped" issue this PR fixes. Status derivation is unaffected (the signals loop sees all placements); the loss is rendering-only.

Suggest iterating all violation-bearing placements and applying the keep/suppress rule per violation (deduping identical messages).

BaggersIO (changes requested):
**Verdict:** Solid PR overall — the flat execution model, the discriminated `ToolRun` union, and the schema-error detail work are real improvements with genuinely good regression-test coverage. Requesting changes for **one correctness bug in new code** (`resolvePlacements` cross-entry placement leak — see the inline comment in `packages/plugin-http-tester/src/index.ts`); everything else below is non-blocking.

Verified while reviewing (rather than assumed):

- The `stepIdx` tagging in both operators is correct — `expectForTransactions` and `runRequests` genuinely operate on `steps.at(-1)` / `steps.at(stepIdx)`, and since steps are append-only the recorded indices stay stable even when later steps are added.
- `report.event.ts`'s `findingSchema` has `additionalProperties: true`, so the new `transactionIndex`/`expected`/`actual` finding fields pass report-event validation.
- The runtime-filter → type-narrowing replacement in `cli-report.ts` is safe: `!run.executions?.length` guards before the assignments.
- All 7 `walkExecutions`/`ExecutionVisit` consumers on `main` are files this PR touches — the breaking traversal migration is complete in-repo.
- The `transactionIndex` rebasing onto the filtered `httpTransactions` array (`rawToExposedTxIndex`) is careful and correct, including dropping the index when the transaction was filtered out.
- JSON-pointer unescape order in `instancePathToProperty` (`~1` before `~0`) is correct per RFC 6901.

## Non-blocking findings

**Unbounded findings with `allErrors: true`** — `packages/core/src/http-testing/validate/ajv.ts` + the `validate-*.ts` mappers now emit one assertion-failure per ajv error with no cap. A pathologically invalid body (say a 50k-element array where every item fails the schema) previously produced one message and now produces tens of thousands of findings, all flowing into the report, markdown, and CSV. Ajv's own docs flag `allErrors` as a DoS consideration with untrusted data. A cap (first N errors + "and X more" summary finding) would keep the improvement without the tail risk.

**`multipleOf` still mislabeled as an expected value** — in `schema-error.ts`, the fallback `isPrimitive(error.schema) ? error.schema : undefined` means `multipleOf: 5` on `7` emits `expected: 5, actual: 7` — exactly the mislabeling `BOUNDARY_KEYWORDS` was introduced to prevent for `maxLength` etc. Suggest adding `multipleOf` to the set.

**`helpUri` scheme not validated (hardening)** — `renderRuleCell` in `markdown.ts`: the quote-escaping fix for attribute breakout is correct, but a `helpUri` of `javascript:alert(1)` still renders as a live link when the markdown is rendered to HTML. Since rule descriptors come from rule packages that already execute as code, this is defense-in-depth only — an `http(s):`-scheme allowlist would close it.

**Unannounced public-API breaks** — the description flags the `walkExecutions`/`ExecutionVisit` break, but two other public-surface removals aren't mentioned: `RuleFinding.severity` + the `RuleFindingSeverity` export (`packages/core/src/rules/rule-violation.ts`) — rule-author-facing; external rule packages that set `severity` on findings will break — and the `Headers`/`BodyContent` types (`packages/core/src/report/index.ts`). Worth listing in the description/changelog alongside the traversal break.

**Small nits**

- `report-builder.ts` — the comment on `httpTestResultToRuleFindings` still says `timeout`/`skip`/`invalid-transaction` are "produced only by the test runner, never by the lint/analyze rule path that consumes this helper"; the helper is now also consumed by the tester path (`buildTestStep`), so the defensive-dead-code framing is stale.
- `invalid-transaction` results produced in `run-requests.operator.ts` (status-code mismatch paths) still carry no `location`, so the new invalid-transaction → step-finding mapping is only reachable from tests/hooks; production surfaces them via the case's skip reason. Fine, but the asymmetry with the newly-tagged assertion failures is worth knowing.

## Quality notes

- The discriminated `ToolRun` union is a real improvement — it deletes runtime filtering and makes kind/runType mismatches unrepresentable; the single documented `as ToolRun` cast in `createToolRun` is a reasonable, well-explained tradeoff.
- The #423 known limitation (first-occurrence attachment for replayed transactions) is honestly documented in-code with a tracking issue.
- Reusing `format.attributes.hash` instead of recomputing `toHash()` is arguably *more* correct, not just faster — the report's `thymianFormat` map is keyed by the sender's hash, so reusing it keeps location-resolution keys consistent even if serialization ever drifts.
- Test quality is high: regression tests exist for every fix claimed in the description, including the `helpUri` escaping and the schema-boundary mislabeling.

packages/plugin-http-tester/src/index.ts
matthyk: quick fixed it. However, to solve the underlying problem is out of scope and is tracked with https://github.com/thymianofficial/thymian-internal/issues/451

BaggersIO (approved):
**Verdict:** Both findings from my previous review are fixed properly — `allPlacedResults` spanning all diagnostics entries kills the cross-entry leak, and the per-violation keep/suppress loop restores multi-violation rendering; both come with regression tests that reproduce the original scenarios. Nice work, approving.

**One ask:** please revert the `--build-from-source=false` change in `ci.yaml` — it's a no-op for this failure. better-sqlite3's install script is `prebuild-install || node-gyp rebuild --release`; since no prebuilt exists for Node 20.20.2 on win32-x64, prebuild-install exits non-zero and the `||` fallback runs node-gyp regardless of npm config (and `build-from-source=false` is prebuild-install's default anyway). The lane fails identically with the flag active, and it fails on `main` too — it's the rolled `windows-latest` image (VS "18" unparseable by node-gyp 10.1.0), worth tracking as its own issue rather than a flag in this PR.

**Optional, re Copilot's open thread on `http-test-api-context.ts`:** it's right — a failed top-level `expect` currently reports `failed` with no reason anywhere (the caseless arm synthesizes `violation: {}`, and `computeTestCaseStatus` never falls back to `testCase.reason`, which `ctx.fail` populates with the assertion message). Setting `violation: { message: assertionFailure.message }` in the caseless arm would carry it into `status.reason` with no duplicate-marker risk, since case-level placements never reach `buildTestStep`. Fine as a fast-follow too.

**Nit:** `multipleOf` isn't in `BOUNDARY_KEYWORDS`, so it still renders the misleading `expected: 2, actual: 7` pair the boundary fix addressed for `maxLength`/`maximum`.

CI: the two remaining failures (`e2e`, `main (22, windows-latest)`) are a Verdaccio proxy blip and a plugin-sampler cold-start timeout — both untouched by this PR; a re-run should clear them.

matthyk:
this PR includes a fix to run node v20 on windows. We should remove v20 anyways, but this is out of scope for this PR. See https://github.com/thymianofficial/thymian-internal/issues/406

PR #330, opened 2026-07-23
chore(repo): migrate to nx v23

matthyk:
Upgrades the workspace from Nx 22.6.4 → 23.1.0 and applies all associated nx migrate codemods.

### Dependency bumps (package.json)
- nx, @nx/eslint, @nx/eslint-plugin, @nx/js, @nx/node, @nx/vitest: 22.6.4 → 23.1.0
- Regenerated package-lock.json

### Release config (nx.json)
- Consolidated the legacy flat releaseTagPattern: "{version}" into the nested releaseTag: { pattern: "{version}" } object. The flat property was removed in Nx 23.

tslib added as an explicit dependency to packages that emit helper-dependent output: core, plugin-openapi, rules-api-description-validation, rules-rfc-9110.

### Ignore files
- .gitignore: added .nx/migrate-runs, .nx/polygraph, .nx/self-healing
- .prettierignore: added .nx/self-healing
- 
### Migration bookkeeping
- Refreshed migrations.json with the Nx 23 migration set (internal-subpath import rewrites for @nx/eslint|node|js, createNodesV2 → createNodes renames, ESLint flat-config / typescript-eslint v8 rule handling).
- Added AI-migration prompt docs under tools/ai-migrations/@nx/eslint/23.1.0/ (convert-to-flat-config.md, migrate-ban-types-rule.md).

### Notes

- No product code changed — this is tooling/config only.
- The ESLint flat-config and ban-types migrations are AI-prompt driven (options don't map 1:1); their prompt files are committed for reference/reproducibility.


BaggersIO (approved):
Upgrade looks good — mechanical, matches the description, and CI exercises Nx 23 end-to-end (target inference, lockfile, build/test on Node 20/22/24 × 3 OSes). A few things I verified beyond CI:

- Peer-dep tightening is satisfied: `@nx/eslint` 23 drops ESLint 8 (we're on ^9.23.0), `@nx/eslint-plugin` needs `@typescript-eslint/parser` ^8 (we're on 8.44.0), `@nx/vitest` drops vitest 1/2 (we're on ^4.1.10).
- The subpath-rewrite / `createNodesV2` codemods were correctly no-ops — nothing in the repo imports `@nx/*/src/*` or `createNodesV2`.
- The ESLint AI migrations were no-ops too (already flat config, no `.eslintrc*` left, no `ban-types` references); committing the prompt docs for reference is fine.
- `releaseTag: { pattern }` matches the Nx 23 schema (flat `releaseTagPattern` removed). Flagging because no workflow runs `nx release`, so this only gets exercised at the next manual release — worth a quick dry-run then.

One non-blocking suggestion:

**`plugin-request-dispatcher` is now the only package without a `tslib` dependency.** All 15 packages build with `importHelpers: true`; this PR brings the count to 14/15. Not a live issue — with target ES2023 no dist currently emits a tslib import (I grepped all build outputs) — but if that package ever starts emitting a helper, its published dist would import an undeclared dependency. A one-liner (`"tslib": "^2.3.0"`) would make it uniform; fine as a follow-up.

Nit: `.gitignore` / `.prettierignore` lost their trailing newlines (migration side effect).

PR #337, opened 2026-07-29
feat: reintroduce --sort-reports-by flag

matthyk:
# Reintroduce `--sort-reports-by` grouping flag

## Summary

Reintroduces the `--sort-reports-by=<rule|endpoint|severity>` `BASE` flag on `lint`, `test`, and `analyze`. It controls how validation findings are **grouped** in report output. The flag was during the report-model migration; this brings it back as a **render-time grouping strategy** rather than reviving the deleted core `report-sorter.ts`/`sortReports()` reshaping.

`endpoint` is the default and reproduces today's output **byte-for-byte** on every surface. `rule` and `severity` are new grouping modes.

## Behavior

| Mode | Groups findings by |
|------|--------------------|
| `endpoint` (default) | Resolved location (lint/analyze) / test-case name (test) — unchanged from today |
| `rule` | Rule id (`unnamed check` for rule-less executions) |
| `severity` | Resolved severity, ordered **error → warn → hint → info**, with a dedicated `skipped` group last |

Applies to the **CLI terminal report** and the **markdown** file formatter. **CSV is intentionally a no-op** — it's a flat data export whose `rule_id`/`location`/`severity` columns already let consumers sort themselves; grouping is a presentation concern.

**CLI terminal** grouped modes lead each group with a severity-symbol heading + violation count and list each violation by location + reason (+ rule under severity grouping). For example, `lint --sort-reports-by=severity`:

```
  ✖ ERRORS (2)

    • GET /a
      ➜ Date header missing
       › rfc9110/date-header

    • POST /b
      ➜ charset should be specified

  ⚠ WARNINGS (1)

    • PUT /c
      ➜ deprecated media type
       › rfc9110/media-type
```

**Markdown** grouped-mode tables gain a `Location` column (so the location isn't lost when it's no longer the heading) and drop the column the heading already carries; test cases regroup under a `###` group heading with `####` per-case sub-headings. Headings mirror the CLI (`### ✖ error: rfc9110/date-header (2)`, `### ✖ ERRORS (2)`).

## What changed

- **`@thymian/core`** — added the `SortReportsBy` type + `SEVERITY_GROUP_ORDER` constant (shared, single source of truth).
- **`@thymian/common-cli`**
  - New `--sort-reports-by` `BASE` flag on the run commands (oclif rejects invalid values, listing the options).
  - `create-execution-renderer` refactored to pluggable heading/entry renderers; `renderReport` selects grouping + ordering per mode; `sortRecordBySeverity` orders severity groups.
  - `applyOptionsToPlugins` injects the flag into the reporter plugin's options **only when the reporter is already configured** (never auto-registers it). Precedence: an explicit `--sort-reports-by` wins over `-o …sortReportsBy=` / config.
- **`thymian` (`lint`/`test`/`analyze`)** — pass the flag through `handleWorkflowOutcome`.
- **`@thymian/plugin-reporter`** — `sortReportsBy` plugin option (JSON-schema-typed); markdown formatter grows mode-aware columns + mirrored headings; CSV unchanged. No new dependency on `common-cli` (shared only via `@thymian/core`).

## Invariants preserved

- `endpoint` (default) output is byte-identical on CLI, markdown, and CSV.
- No revival of core report reshaping (ADR-0014); report model v4 untouched.
- Reporter is never auto-registered by the flag.
- Module boundaries intact (no `plugin-reporter → common-cli` edge).

## Edge cases handled (surfaced by adversarial code review)

- **Skipped executions** get their own `skipped` group under `severity`, never mislabeled as errors.
- **Passed executions** (and their informational findings) are excluded from `rule`/`severity` grouping, matching the CLI — so no `ERRORS (0)` heading with a mis-filed row.
- **Descriptor-less rules**: the markdown `rule` heading resolves severity from the group's executions (CLI parity) rather than hardcoding `error`.
- Rule ids are HTML-escaped in markdown headings, like table cells.

## Testing

- Unit tests for all three modes across `lint`/`analyze`/`test` in the CLI renderer and the markdown formatter, plus CSV flag-invariance and the edge cases above.
- e2e (`cli-lint`): flag acceptance + invalid-value rejection.
- `endpoint` regressions all pass unchanged. `build` + `test` green across `core`, `common-cli`, `plugin-reporter`, `thymian`; lint clean; Prettier clean.

## Docs

The flag's `description` documents it for `--help`; CI regenerates the oclif README / astro-docs references from it (no hand-edited generated docs).

### DIFF
--- a/e2e-tests/src/cli-lint.test.ts
+++ b/e2e-tests/src/cli-lint.test.ts
@@ -126,6 +126,37 @@ describe('thymian lint', () => {
     expect(normalize(first.stdout)).toBe(normalize(second.stdout));
   }, 180_000);
 
+  it('should accept --sort-reports-by and still render the report', () => {
+    copyFixturesToTempDir(join(fixturesDir, 'static-lint'), getTempDir());
+
+    const { stdout, exitCode } = execThymianResult(
+      ['lint', '--sort-reports-by', 'severity'],
+      { cwd: getTempDir() },
+    );
+
+    // Findings are present, so the exit code stays non-zero; the flag is
+    // accepted and the report is grouped/rendered as usual.
+    expect(exitCode).not.toBe(0);
+    expect(stdout).toContain(
+      'rfc9110/origin-server-with-clock-must-generate-date-for-2xx-3xx-4xx',
+    );
+    expect(stdout).toContain('Summary: 1 error, 0 warnings, 0 hints, 0 infos.');
+  }, 90_000);
+
+  it('should reject an invalid --sort-reports-by value listing the options', () => {
+    copyFixturesToTempDir(join(fixturesDir, 'static-lint'), getTempDir());
+
+    const { exitCode, stderr } = execThymianResult(
+      ['lint', '--sort-reports-by', 'nonsense'],
+      { cwd: getTempDir() },
+    );
+
+    expect(exitCode).toBe(2);
+    expect(stderr).toContain('rule');
+    expect(stderr).toContain('endpoint');
+    expect(stderr).toContain('severity');
+  }, 90_000);
+
   it('should separate report content (stdout) from operational messages (stderr)', () => {
     copyFixturesToTempDir(join(fixturesDir, 'static-lint'), getTempDir());
 
--- a/packages/common-cli/src/apply-plugin-options.ts
+++ b/packages/common-cli/src/apply-plugin-options.ts
@@ -1,0 +1,52 @@
+import type { SortReportsBy } from '@thymian/core';
+
+import type { ThymianConfig } from './thymian-config.js';
+
+/** The reporter plugin whose file formatters honour `--sort-reports-by`. */
+export const REPORTER_PLUGIN_NAME = '@thymian/plugin-reporter';
+
+/**
+ * Sets a single option on a plugin's config — but ONLY when that plugin is
+ * already configured, so wiring a flag never auto-registers a plugin the user
+ * did not ask for. Overwrites any existing value for `key`.
+ */
+export function setConfiguredPluginOption(
+  config: ThymianConfig,
+  pluginName: string,
+  key: string,
+  value: unknown,
+): void {
+  const plugin = config.plugins[pluginName];
+  if (!plugin) {
+    return;
+  }
+
+  plugin.options ??= {};
+  (plugin.options as Record<string, unknown>)[key] = value;
+}
+
+/**
+ * Forwards an explicit `--sort-reports-by` flag to the reporter plugin so its
+ * file formatters group findings to match the CLI report.
+ *
+ * PRECEDENCE: this is called AFTER the `-o` overrides are applied, so an
+ * explicit flag intentionally WINS over — and overwrites — any
+ * `-o …options.sortReportsBy=…` or config-file value. When the flag is absent
+ * (`undefined`) nothing is written, so the `-o`/config value survives. If the
+ * reporter is not configured, this is a no-op (never auto-registers it).
+ */
+export function applyReporterSortReportsBy(
+  config: ThymianConfig,
+  sortReportsBy: SortReportsBy | undefined,
+): void {
+  if (sortReportsBy === undefined) {
+    return;
+  }
+
+  setConfiguredPluginOption(
+    config,
+    REPORTER_PLUGIN_NAME,
+    'sortReportsBy',
+    sortReportsBy,
+  );
+}
--- a/packages/common-cli/src/base-cli-run-command.ts
+++ b/packages/common-cli/src/base-cli-run-command.ts
@@ -114,6 +114,12 @@ export abstract class BaseCliRunCommand<
       helpGroup: 'BASE',
       options: ['off', 'error', 'warn', 'hint'],
     }),
+    ['sort-reports-by']: Flags.string({
+      description:
+        'Group report findings by rule, endpoint, or severity (default: endpoint). Affects the CLI report and any configured reporter that supports grouping.',
+      helpGroup: 'BASE',
+      options: ['rule', 'endpoint', 'severity'],
+    }),
     timeout: Flags.integer({
       default: Thymian.DEFAULT_TIMEOUT,
       charAliases: ['t'],
@@ -302,6 +308,7 @@ export abstract class BaseCliRunCommand<
       this.debug('Autoloading Thymian plugins.');
       this.logger.info('Autoloading plugins from configuration...');
       await this.addPluginsToThymianConfig();
+      this.applyOptionsToPlugins();
       await this.registerPluginsFromConfig();
     }
 
@@ -438,6 +445,46 @@ export abstract class BaseCliRunCommand<
         override.value,
       );
     }
+  }
+
+  /**
+   * Maps CLI flags onto plugin options before plugins are registered — the one
+   * place where a flag is wired to a specific plugin's config. Add future
+   * flag→plugin mappings here.
+   *
+   * Currently: `--sort-reports-by` is forwarded to the reporter plugin so its
+   * file formatters group findings to match the CLI report. The terminal
+   * renderer receives the flag via a separate channel (`handleWorkflowOutcome`).
+   */
+  protected applyOptionsToPlugins(): void {
+    const sortReportsBy = this.flags['sort-reports-by'];
+    if (sortReportsBy !== undefined) {
+      this.setPluginOption(
+        '@thymian/plugin-reporter',
+        'sortReportsBy',
+        sortReportsBy,
+      );
+    }
+  }
+
+  /**
+   * Sets a single option on a plugin's config — but ONLY when that plugin is
+   * already configured, so wiring a flag never auto-registers a plugin the user
+   * did not ask for, and leaves any existing config/`-o` value untouched when
+   * the flag is absent (callers guard on that).
+   */
+  private setPluginOption(
+    pluginName: string,
+    key: string,
+    value: unknown,
+  ): void {
+    const plugin = this.thymianConfig.plugins[pluginName];
+    if (!plugin) {
+      return;
+    }
+
+    plugin.options ??= {};
+    (plugin.options as Record<string, unknown>)[key] = value;
   }
 
   /**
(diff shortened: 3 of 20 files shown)


BaggersIO (changes requested):
**Verdict: high-quality PR — no correctness bugs found.** Requesting changes for two small pre-merge items (findings 1 and 4); the rest are follow-up candidates.

## Verified claims

I checked the PR's core claims against the branch source (not just the diff):

- **Byte-identical `endpoint` output (CLI):** confirmed. The indentation offsets were rebased correctly in the refactor (status line old `+1` → entry base `+1`; rule id old `+6` → `level+5`; details old `+7` → `level+6` in `endpointEntryRenderer`). Markdown keeps insertion-order groups and the old table layout in the default mode.
- **Precedence doc-comment is accurate:** `overridePluginOptions()` (the `-o` path) runs in `init()` well before `applyOptionsToPlugins()` (`base-cli-run-command.ts`), so an explicit flag wins as documented.
- **Skipped invariant holds:** `resolveExecutionSeverity()` returns `undefined` for every non-`failed` execution (even when the rule descriptor carries a severity), so skipped executions always land in the `skipped` group.
- **No auto-registration:** `setPluginOption` only writes when the plugin entry already exists.
- **Module boundaries intact:** `markdown.ts` imports only from `@thymian/core`; no `plugin-reporter → common-cli` edge.
- **No external breakage:** `createExecutionsRenderer` / `renderReport` have no call sites outside the changed files, so the new signatures break nothing.

## Findings

### 1. CLI `severity` mode loses rule identity when the rule has no descriptor (minor, pre-merge)

`groupedEntryRenderer` derives the `› rule-id` ref from the resolved descriptor (`options.showRule && rule`), not from `execution.ruleId` (`packages/common-cli/src/render/create-execution-renderer.ts`). If a rule has no descriptor in the run, the `›` line is omitted — and under severity grouping nothing else carries rule identity (the heading is the severity). The markdown formatter shows the raw `ruleId` in the same case (`renderRuleCell(execution.ruleId, …)`), so the two surfaces diverge exactly where this PR aims for parity. The historical justification for hiding it ("descriptor missing ⇒ `rule.id` would equal `execution.ruleId`") worked for the endpoint layout but doesn't transfer: here the information is lost entirely. Suggestion: render the ref from `execution.ruleId` and use the descriptor only for `helpUri`/summary.

### 2. Markdown grouped tables drop the per-row kind label → count/rows mismatch (minor, UX)

The removed first column encoded the row type (`failed`/`skipped`/`info`/severity word), not just the severity. Under `rule`/`severity` grouping: a skipped row is indistinguishable from a failed one in `rule` mode, and informational findings of a failed execution appear as identical-looking extra rows under e.g. `### ✖ ERRORS (1)` — the heading counts 1, but 2+ rows sit below it. The CLI doesn't have this problem (findings nest visually under their entry). Suggestion: keep a Status column in grouped modes, or document the trade-off deliberately.

### 3. Mirror logic is duplicated rather than shared (maintainability)

`SortReportsBy` / `SEVERITY_GROUP_ORDER` live in core as the "single source of truth", but the *logic* is implemented twice: group key (`selectGroupBy` vs `executionGroupKey`, incl. the identical `?? 'skipped' : 'error'` fallback), severity ordering (`sortRecordBySeverity` vs `orderGroupKeys`), pluralization (`pluralizeSeverity` vs `pluralizeSeverityWord`), and the `'unnamed check'` literal (3 sites). Since both surfaces must stay in sync by design, drift risk is real. All four pieces are renderer-neutral and could live in core (e.g. `severityGroupKey()`, `compareSeverityGroupKeys()`, `pluralizeSeverity()`, `UNNAMED_CHECK_LABEL`) without violating the module boundary.

### 4. The documented flag→plugin precedence contract is untested (test gap, pre-merge)

`applyOptionsToPlugins`/`setPluginOption` have no unit test: neither "flag overrides `-o …sortReportsBy=`", nor "without the flag the `-o`/config value survives", nor "reporter not configured → no-op". The e2e tests only cover the terminal path. Since the contract is spelled out in the doc comment, a regression guard is cheap and valuable. Also missing: a markdown test for skipped *test cases* under `severity` (only lint is covered).

### 5. Value list repeated three times; casts instead of a typed flag (nit)

`['rule', 'endpoint', 'severity']` appears in the flag definition, the JSON-schema enum, and as the core type. A `SORT_REPORTS_BY_VALUES` const in core plus `Flags.custom<SortReportsBy>()` would remove the repetition and the three `as SortReportsBy | undefined` casts in `lint.ts`/`test.ts`/`analyze.ts`.

### 6. Polish (nit)

In `groupedEntryRenderer`: the `options.showRule && ruleRef` guard is redundant (`ruleRef` is already `''` otherwise); the leading space inside the `ruleRef` string acting as a half-indent is easy to miss; typo `detailIndentionLevel`. Also `buildLintAnalyzeSection`'s doc comment ("location-grouped `Severity | Rule | Message` tables") is now only accurate for the default mode.

## Tests / security / performance

- Tests are strong overall: all three modes across lint/analyze/test on both surfaces, skipped group, passed exclusion, descriptor-less fallback, HTML escaping, CSV invariance, e2e accept/reject. Gaps: see finding 4.
- Security: rule ids are HTML-escaped in headings (tested); the new Location column follows the existing `escapeCell` convention. No new attack surface.
- Performance: negligible (linear grouping, `indexOf` over ≤5 keys).
- Observation (no action needed): as a BASE flag, `--sort-reports-by` also shows up on `serve`/`request`/`validate` etc., where only the reporter file path honors it — consistent with existing BASE-flag conventions. Likewise, the asymmetry that a `-o`/config-only `sortReportsBy` affects file formatters but never the terminal report is intentional per the design; worth a sentence in the docs.

---

Requesting changes for findings 1 and 4 — both should be quick. Findings 2, 3, and 5 are fine as follow-ups if you'd rather keep this PR contained.

BaggersIO (approved):
**Re-review of f7b410c — all six findings addressed, no new issues found. Approving.**

Both pre-merge items (F1, F4) are fixed with regression tests, and the optional ones (F2, F3, F5, F6) were addressed as well. CI is fully green (unit matrix 3 OS × 3 Node, e2e, lint, CodeQL).

## Fix verification

- **F1 (CLI severity mode loses rule identity) — fixed.** `groupedEntryRenderer` now derives the `›` ref from `execution.ruleId`; the descriptor is only used for the message fallback (`packages/common-cli/src/render/create-execution-renderer.ts`). Regression test covers the descriptor-less case (`› ghost/rule`), restoring parity with markdown.
- **F2 (markdown count/rows mismatch) — resolved by dropping the count.** Instead of adding a Status column, both surfaces drop the violation count from `rule`/`severity` group headings. That removes the contradiction at its root (markdown counted executions but renders per-finding rows, so a failed execution with informational findings showed more rows than the count) and keeps CLI/markdown headings mirrored. Deliberate residual, fine as a follow-up: grouped markdown tables still carry no per-row status label, so in `rule` mode a skipped row looks like a failed one (the CLI marks skips with `⏭ skipped:`; in `severity` mode the group itself disambiguates).
- **F3 (duplicated mirror logic) — fixed.** New `packages/core/src/report/grouping.ts` centralizes `severityGroupKey`, `resolveGroupSeverity`, `compareSeverityGroupKeys`, `pluralizeSeverity`, `UNNAMED_CHECK_LABEL`; verified both surfaces now delegate to it (CLI via `selectGroupBy`/`sortRecordBySeverity`, markdown via `executionGroupKey`/`orderGroupKeys`).
- **F4 (untested flag→plugin contract) — fixed.** The injection is extracted to pure `applyReporterSortReportsBy`/`setConfiguredPluginOption` with unit tests for all four contract points: flag overrides `-o`/config, absent flag preserves the `-o` value, never-auto-registers, creates a missing `options` object. The markdown skipped-test-case-under-`severity` gap is also covered now.
- **F5 (value list ×3, casts) — fixed.** `SORT_REPORTS_BY_VALUES` in core drives the flag (`Flags.custom<SortReportsBy>()`), the type, and the schema enum; all three `as SortReportsBy` casts in the commands are gone.
- **F6 (polish) — fixed.** Guard is now meaningful (`options.showRule && execution.ruleId`), the leading-space half-indent is an explicit `indent(indentationLevel + 1)`, the `detailIndentionLevel` typo is gone, and `buildLintAnalyzeSection`'s doc comment matches the mode-aware behavior.

## Invariants re-verified after the refactor

- **Precedence still holds in the new placement:** `overridePluginOptions()` (the `-o` path, `init()` line 198) runs before `applyOptionsToPlugins()` (line 314) — now also locked in by the unit tests.
- **Injection placement is sound:** `applyOptionsToPlugins()` sits inside the `shouldAutoload()` branch, which is the only path where `registerPluginsFromConfig()` runs — there is no path where a configured reporter misses the flag.
- **Passed executions can't distort severity groups:** `severityGroupKey` would bucket a passed execution under `error`, but markdown excludes passed executions before grouping and the CLI filters them in the render loop (and suppresses all-passed groups); with counts gone, headings can't over-count either.
- **No half-applied surface:** `validate`/`request`/`serve` never call `handleWorkflowOutcome`/`renderReport`, so no command renders a terminal report that ignores the flag; on those commands the BASE flag affects only the reporter file, consistent with the documented design.
- **Endpoint byte-identity survives the test-section refactor:** the `<details>` summary rebuild via `summaryParts.join(' · ')` reproduces the old string exactly, including the trailing-separator quirk for empty messages.
- **Schema/plumbing:** the `sortReportsBy` option follows the sibling `nullable: true` convention and derives its enum from the shared const; `getFormatters` passes it to every formatter after validation, with CSV proven byte-invariant by test.

## Remaining nits (non-blocking)

1. **Stale count references:** the PR description still shows `✖ ERRORS (2)` / "violation count" examples, and two test comments still mention a count (`cli-report-renderer.test.ts` "rule id, and a violation count"; `markdown.test.ts` "severity symbol + severity + rule id + count"). Worth updating the PR body before merge; the comments can ride along with any later touch.
2. **Observation only:** a descriptor-less rule group containing only skipped executions gets an `✖ error:` rule heading (`resolveGroupSeverity` → `undefined` → `'error'` default). Both surfaces agree and the default is documented, so this is fine — just noting it's a known edge.

PR #338, opened 2026-07-30
chore(deps): shrink thymian CLI install size

matthyk:
Closes https://github.com/thymianofficial/thymian-internal/issues/302

Removes the packages `sharp` and `@oclif/plugin-plugins` to reduce install size.


BaggersIO (approved):
**Verdict: Approve.** Both removals match the decisions recorded in thymian-internal#302, CI is green across the full matrix (Node 20/22/24 × 3 OSes, e2e, lint, license, CodeQL), and I verified the replacement mechanism end-to-end down to the published-tarball path. Everything below is a non-blocking suggestion.

## What I verified (all sound)

- **`@oclif/plugin-plugins` was dead weight.** It was in `dependencies` but never listed in `oclif.plugins` — on `main` either — so `thymian plugins install` was never actually wired up. Removal is a true runtime no-op, exactly as the issue's team decision assumed. The eslint `@nx/dependency-checks` allowlist was kept in sync.
- **The assets reach users.** `tsc --build` (`rootDir: src` → `outDir: dist`) puts the compiled strategy at `dist/generation/content-type-strategies/`, `copy-assets` puts the assets in the sibling `assets/` dir, so the top-level `join(import.meta.dirname, 'assets')` resolves correctly both from `src` (vitest) and `dist` (runtime). `dependsOn: ["^build", "copy-assets"]` merges correctly into the Nx-inferred build target, `tsc --build` doesn't clean `dist/` so the pre-copied assets survive, `nx release`'s `preVersionCommand` runs a full build before publishing, and `files: ["dist"]` includes them in the tarball.
- **The committed binaries are what they claim.** I decoded both: 652 B JPEG + 1.4 KB PNG, genuinely 256×256 — ~2 KB replacing a ~19 MB native dependency subtree.
- **No public-API break.** `ImageContentTypeStrategyOptions` and the constructor options were internal-only — the package's `index.ts` doesn't export the strategy, and the sole call site (`generate-samples.ts`) passes no args.
- **Lockfile is genuinely clean.** No `sharp`/`@img/*`/`npm`/`yarn` entries are re-added anywhere; the unrelated-looking jest churn under `e2e-tests/` is normal npm re-resolution, and `npm ci` passed on all nine matrix jobs.
- **Bonus fixes worth calling out:** the old code returned `$ext: 'png'` even for JPEG bodies — now correct — and the old random-pixel generation made every `sampler init` produce different bytes; fixed assets make image samples deterministic and diff-friendly. The magic-byte assertions in the new test are a nice touch: they validate the committed binaries themselves.

## Suggestions (non-blocking)

1. **Nothing exercises the asset wiring from the built artifact.** The unit test runs via vitest against `src/` (where the assets also live), and `e2e-tests` has no `image/*` content-type fixture — so if `copy-assets` ever silently breaks (rename, path drift), CI stays green and users get `ENOENT` at runtime. A small e2e fixture with an image request body would close that gap cheaply.
2. **Two patterns now exist for the same problem.** `plugin-http-analyzer` ships `init.sql` via an inline build command (`tsc && cpSync ...`); this PR introduces a separate cached `copy-assets` target. The new approach is the better one (cacheable, declared inputs/outputs, keeps the inferred build target) — consider a follow-up migrating the analyzer to it so the repo has one convention.
3. **Node-floor nuance.** `ASSETS_DIR` is computed at module top level, so on Node < 20.11 (`import.meta.dirname` is `undefined`) the module throws at *load* time rather than at feature use. There's precedent (the analyzer reads `init.sql` the same way, though lazily inside a method) and CI's floor is latest Node 20, so nothing changes in practice — but no published package declares `engines.node`, so old-Node users get a cryptic `TypeError` instead of an npm warning. Worth an `engines` follow-up; computing the path lazily inside `generate()` would also soften it.
4. **Changelog visibility.** With `nx release` on conventional commits, a `chore(...)` squash title produces no changelog entry — but the `$ext` fix and the switch to deterministic bytes are user-visible sampler behavior changes. Consider `fix(plugin-sampler): ...` framing for the squash commit (or a release-notes note) so it doesn't land silently.
5. **Issue AC hygiene.** thymian-internal#302's ACs ask for a measured before/after — posting `npm pack` / cold-install numbers here would close that box. Also, the "no bundled C/C++ sources" AC can't be met while `better-sqlite3` stays (kept by explicit decision in the issue), so the checklist there should be amended rather than left dangling.
6. **Drive-by, pre-existing:** `@oclif/plugin-version` appears twice in the eslint `ignoredDependencies` list in `packages/thymian/eslint.config.mjs` — trivial to dedupe while you're in that file.

## Risks

Low. The one real failure mode — assets missing from the published `dist/` — is mitigated by the release pipeline's mandatory pre-publish build, but has no automated guard (suggestion 1). Everything else is either verified or pre-existing behavior unchanged.

PR #339, opened 2026-07-30
fix(plugin-openapi): declare @scalar/json-magic runtime dependency

matthyk:
## Problem

`@thymian/plugin-openapi` imports `@scalar/json-magic` at runtime (`src/load-openapi.ts`: `@scalar/json-magic/bundle` and `@scalar/json-magic/bundle/plugins/node`) but did not declare it in `dependencies`. Under npm/yarn hoisting the import resolved incidentally via the transitive dep from `@scalar/openapi-parser`. Under pnpm's strict per-package isolation it fails and **every** Thymian CLI command dies at plugin load:

```
Cannot find package '@scalar/json-magic'
Failed to load plugin module "@thymian/plugin-openapi"
PluginLoadError  (exit code 2)
```

`pnpm dlx thymian lint ...` is a complete repro; `--config.shamefully-hoist=true` does not help. Confirmed in `0.1.12` and `0.1.13-canary.20260727-02bb901`.

## Fix

Declare `@scalar/json-magic` `^0.9.0` in `packages/plugin-openapi/package.json` — compatible with the `0.9.0` that `@scalar/openapi-parser@^0.23.2` resolves (root already hoists `0.9.6`, which satisfies the range). Adds the matching direct-dependency edge to `package-lock.json`; no package versions change.

## Notes

- Purely additive dependency declaration for an already-installed, already-used package. No code or resolved-version changes under npm/yarn; the fix restores resolution under strict installers (pnpm).

Tracked in thymianofficial/thymian-internal#454.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34813,6 +34813,7 @@
       "version": "0.0.0-PLACEHOLDER",
       "license": "AGPL-3.0-only",
       "dependencies": {
+        "@scalar/json-magic": "^0.9.0",
         "@scalar/openapi-parser": "^0.23.2",
         "@scalar/openapi-upgrader": "^0.1.4",
         "@thymian/common-cli": "0.0.0-PLACEHOLDER",
--- a/packages/plugin-openapi/package.json
+++ b/packages/plugin-openapi/package.json
@@ -36,6 +36,7 @@
     "openapi-types": "^12.1.3"
   },
   "dependencies": {
+    "@scalar/json-magic": "^0.9.0",
     "@scalar/openapi-parser": "^0.23.2",
     "@scalar/openapi-upgrader": "^0.1.4",
     "@thymian/common-cli": "0.0.0-PLACEHOLDER",


BaggersIO (approved):
**Verdict: Approve** — minimal, correct fix for thymianofficial/thymian-internal#454; I verified the PR's claims independently rather than taking the description at face value.

### What this does

Declares `@scalar/json-magic@^0.9.0` as a direct dependency of `@thymian/plugin-openapi` (plus the matching lockfile edge), fixing the phantom-dependency crash that kills every CLI command under pnpm's strict isolation.

### What I verified

- **The imports are real**: `src/load-openapi.ts:4,10` imports `@scalar/json-magic/bundle` and `@scalar/json-magic/bundle/plugins/node` at runtime, and the package was previously undeclared — exactly the failure mode described.
- **The range floor is safe**: both subpath exports (`./bundle`, `./bundle/plugins/node`) and all named exports used (`bundle`, `fetchUrls`, `parseJson`, `parseYaml`, `readFiles`) exist in json-magic `0.9.0` (checked against the nested copy that `@scalar/openapi-parser` pins) as well as in the hoisted `0.9.6`. So even a resolver that dedupes down to the parser's exact `0.9.0` pin (npm `--prefer-dedupe`, pnpm lowest resolution-mode) resolves every specifier this plugin uses.
- **No resolution changes**: the root already hoists `0.9.6` — via `@thymian/plugin-sampler`, which declares the identical `^0.9.0` range, so the range choice is also consistent with the sibling package. `openapi-parser` keeps its nested pinned `0.9.0`. The lockfile hunk adds only the dependency edge; CI runs `npm ci`, so the green 3-OS × 3-Node matrix confirms lock/manifest consistency.
- **Scope is complete**: I swept all workspace packages for `@scalar/json-magic` imports — only `plugin-openapi` (fixed here) and `plugin-sampler` (already declared) use it, so no sibling package hides the same bug. The remaining bare imports in `plugin-openapi/src` are all declared; `openapi-types` appears only in type-only imports, so its devDependencies placement is also correct under strict installers.
- **Dual-copy concern is benign**: plugin code (`0.9.6`) and `openapi-parser` (`0.9.0`) use separate json-magic copies, but only plain document objects cross that boundary (`bundle()` output → `validate`/`dereference`), and this split is already today's behavior under npm hoisting — this PR doesn't change it.

### Non-blocking follow-ups

- This bug class (undeclared-but-hoisted runtime dep) is invisible to npm-based CI and the current ESLint setup (no `import/no-extraneous-dependencies`, no knip/depcheck). Worth adding one such guard in a separate PR so the next phantom dependency fails in CI instead of in a pnpm user's environment.
- Since the reporter offered to test a patch: a canary release after merge would let them confirm the fix with `pnpm dlx` before it ships in a stable version.

PR #340, opened 2026-07-31
feat(core): enforce that executable rules define their execution functions

matthyk:
## Summary

Implements the two-layer enforcement decided in [thymianofficial/thymian-internal#359](https://github.com/thymianofficial/thymian-internal/issues/359): a rule declaring an executable type (`static`/`analytics`/`test`) without the matching execution function (`lintRule`/`analyzeRule`/`testRule`) registered but silently never ran — the bug class behind thymianofficial/thymian-internal#358.

**Invariant:** every declared executable type must have its execution function, and `informational` must be exclusive and function-free.

### Layer 1 — runtime (hard errors)

- New shared module `packages/core/src/rules/rule-execution-invariant.ts` (exported from `@thymian/core`) with the shape check (`meta.type` missing / non-array / empty / unknown entries) and coverage check, plus curated messages and suggestions.
- `RuleBuilder.done()` throws naming exactly the uncovered types; `.type()` rejects `informational` mixed with executable types.
- The rule loader validates in three stages on both load paths: **shape check → rule filter → coverage check**, wrapped as `InvalidRuleError` with suggestions and a docs ref. Running the coverage check only on rules that pass the filter means a broken rule in a third-party package can always be unblocked by disabling it (severity `off`) or downgrading it to `['informational']` via config.
- Config overrides (severity/type/options) now also apply to rules shipped **inline** in a rule set (previously they were ignored on that path), so the escape hatches work there too.

### Layer 2 — type level

The builder interfaces track which declared types are covered by `.rule()` / `.override*Rule()`; `done` only becomes callable once every executable type is covered, otherwise it resolves to a marker type naming the missing types in the compile error. `'informational'` cannot be combined with executable types at the type level either. Since every rule package compiles with `tsc`, this is enforced in CI for all 402 rfc-9110 rules.

### Also included

- `thymian generate rule` rejects selecting `informational` together with other types (previously scaffolded a file that throws).
- New docs page `references/Errors/invalid-rule-error.md` for the error `ref`.
- A compile-time conformance assertion in `rule-builder.ts` replacing the `implements` clause that the conditional `done` property makes inexpressible.

## Testing

- 13 new unit tests for the invariant module, 7 for the builder, 9 loader tests (broken/hand-built rules, inline rule sets, config overrides, filter interaction, malformed `meta.type` fixtures).
- The existing `loadRules('@thymian/rules-rfc-9110')` test now sweeps all 402 built rules through the load-time validation.
- Verified a probe file against the built types: all four invalid chain shapes fail with the intended compile errors; all valid shapes compile.
- Full workspace build, test, lint, and format pass.

Refs thymianofficial/thymian-internal#359

### DIFF
--- a/astro-docs/src/content/docs/references/Errors/invalid-rule-error.md
+++ b/astro-docs/src/content/docs/references/Errors/invalid-rule-error.md
@@ -1,0 +1,43 @@
+---
+title: 'InvalidRuleError'
+---
+
+## The Cause
+
+A loaded rule declares one or more executable types (`static`, `analytics`, or `test`) but has no execution function for at least one of them. Such a rule would register but silently never run. This can happen when:
+
+- a rule object is constructed by hand (without the `httpRule` builder) and misses the `lintRule`, `analyzeRule`, or `testRule` function for a declared type, or
+- the `type` override for a rule in your Thymian config file points at a type the rule has no execution function for, or
+- a rule combines the `informational` type with executable types — `informational` must be the only type of a rule, or
+- a rule's `meta.type` is malformed: missing, not an array, empty, or containing unknown rule types.
+
+## The Solution
+
+Make sure every declared executable type has a matching execution function:
+
+```typescript
+import { httpRule } from '@thymian/core';
+
+// ✅ .rule() provides the execution function for every declared type
+export default httpRule('my-rule')
+  .severity('warn')
+  .type('static', 'analytics')
+  .rule((context, options, logger) => {
+    /* rule logic */
+  })
+  .done();
+```
+
+If the rule is documentation-only, declare it as `informational` instead — informational rules are the only rules without execution functions:
+
+```typescript
+export default httpRule('my-informational-rule')
+  .severity('warn')
+  .type('informational') // informational rules never execute
+  .description('Documented requirement that cannot be checked automatically.')
+  .done();
+```
+
+If the error points at a `type` override in your Thymian config file, restrict the override to types the rule actually implements, or set the rule's type to `['informational']` to stop executing it.
+
+If the broken rule comes from a package you do not control, disable it by setting its severity to `off` in your Thymian config file — a disabled rule is filtered out before this validation runs — or downgrade it to `['informational']` via the `type` override.
--- a/packages/core/src/rules/index.ts
+++ b/packages/core/src/rules/index.ts
@@ -4,6 +4,7 @@ export * from './contexts.js';
 export * from './rule.js';
 export * from './rule-builder.js';
 export * from './rule-configuration.js';
+export * from './rule-execution-invariant.js';
 export * from './rule-filter.js';
 export * from './rule-fn.js';
 export * from './rule-loader.js';
--- a/packages/core/src/rules/rule-builder.ts
+++ b/packages/core/src/rules/rule-builder.ts
@@ -315,32 +315,32 @@ class RuleBuilder<
   }
 }
 
-// Compile-time conformance check (never called): an `implements` clause
-// cannot express the conditional `done` property, so assert here that
-// RuleBuilder still structurally satisfies the public builder interfaces.
-// Interface states with a non-callable `done` deliberately narrow the class
-// and can never be supertypes of it, so the provable states are asserted:
-// full conformance for informational rules (which also covers all meta-
-// property methods) and the coverage-complete DefineRules end state for
-// executable rules. `severity` and `type` are covered by the `implements`
-// clause on the class.
+// The instantiation fails to compile unless Sub is assignable to Super.
+type Satisfies<Sub extends Super, Super> = Sub;
+
+// Compile-time conformance check: an `implements` clause cannot express the
+// conditional `done` property, so assert here that RuleBuilder still
+// structurally satisfies the public builder interfaces. Interface states
+// with a non-callable `done` deliberately narrow the class and can never be
+// supertypes of it, so the provable states are asserted: full conformance
+// for informational rules (which also covers all meta-property methods) and
+// the coverage-complete DefineRules end state for executable rules.
+// `severity` and `type` are covered by the `implements` clause on the class.
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-function assertRuleBuilderConformance(
-  executable: RuleBuilder<
-    Record<PropertyKey, unknown>,
-    ['static', 'analytics', 'test']
+type AssertRuleBuilderConformance = [
+  Satisfies<
+    RuleBuilder<Record<PropertyKey, unknown>, ['static', 'analytics', 'test']>,
+    DefineRules<
+      ['static', 'analytics', 'test'],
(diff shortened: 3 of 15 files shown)


BaggersIO (commented):
## Code Review — enforce that executable rules define their execution functions

### Overview

This closes a real bug class (thymian-internal#358): a rule that declares `type: ['static' | 'analytics' | 'test']` but has no matching `lintRule`/`analyzeRule`/`testRule` registers successfully and then silently never executes. The PR adds two enforcement layers:

- **Runtime** — a new shared invariant module (`rule-execution-invariant.ts`) doing a shape check (`meta.type` malformed/empty/unknown) and a coverage check (every declared executable type has its execution function; `informational` is exclusive and function-free). It's wired into `RuleBuilder.type()`/`.done()` and into the rule loader at all three load paths (single rule, inline rule-set array, glob-discovered rule-set files), consistently as **shape → filter → coverage**.
- **Type-level** — `RuleBuilder`'s chain interfaces now thread a `Covered` type parameter so `.done()` is only callable once every declared executable type has been given an execution function via `.rule()`/`.override*Rule()`, with a marker type naming the missing types in the compile error otherwise.
- Plus a CLI validation fix (`generate rule`), a new `InvalidRuleError` docs page, and an unrelated one-line flaky-test fix in `plugin-reporter`.

I fetched the branch's full `rule-loader.ts`, `rule.ts`, `rule-meta.ts`, and `load-rules.test.ts` (beyond the diff hunks) to trace all three load paths and cross-check fixture changes against pre-existing tests.

### Strengths

- **Defense in depth is real, not cosmetic.** Even where the type-level `Covered` tracking does something clever (the class widens its own return type to the full `RuleType` union via unsafe casts and lets the *interface* narrow per chain-state — see the `AssertRuleBuilderConformance` comment), the runtime check in `.done()` is unconditional and independent of the type gymnastics. A hole in the type algebra (or an `as any` bypass) still gets caught at runtime.
- **The `a.rule.mjs` fixture change is deliberate, not incidental.** The pre-existing `overrides type from config` test overrides fixture `a`'s type to `['analytics']`. Before this PR, `a` only had `testRule` (from `.type('test').rule(fn)`), which would now fail the new coverage check. The PR fixes this by changing the fixture to `.type('test', 'analytics')` so `.rule(fn)` populates both slots — this reads as a deliberate check against the existing suite rather than a latent break left behind.
- Three-stage validation (shape → filter → coverage) is applied identically on both the single-rule and inline-rule-set-array load paths, and the glob/pattern path recurses through `loadRules()` per file so it gets the same treatment for free.
- Good test breadth: builder unit tests, invariant-module unit tests, loader tests (including the escape hatches — disable via `off`, downgrade via `informational` override, malformed declarations bypassing the filter), and the full 402-rule `rules-rfc-9110` sweep now exercises load-time validation for real.

### Issues / Suggestions

1. **Builder-time errors discard the curated suggestions (minor-moderate).** `RuleBuilder.type()` and `.done()` both do:
   ```ts
   const { message } = describeRuleExecutionInvariantViolation(...);
   throw new Error(message);
   ```
   Only `message` is used; the `suggestions` array `describeRuleExecutionInvariantViolation` computes is thrown away, and the error is a bare `Error` (no `ref` to the new `invalid-rule-error` docs page). The loader path (`throwInvalidRuleError`) uses the full `ThymianBaseError` with suggestions + ref for the *same* violations. This matches this file's pre-existing convention (other builder throws are also plain `Error`), so it's not a new inconsistency — but since curated suggestions are the PR's own stated goal, and a rule author hitting `.done()` is exactly who'd benefit from "declare `.type('informational')` if this is documentation-only," it seems worth reusing the richer error here too.

2. **Config-induced shape violations don't get the config-override hint (minor).** In `rule-loader.ts`, `assertRuleTypeDeclaration(rule, source)` is always called with a hardcoded `[]` for extra suggestions, while `assertRuleExecutionInvariant` receives `typeOverrideSuggestions(rule, options)`. So if a user's config overrides `type` with something malformed — e.g. `type: 'static'` (string instead of array), a fairly plausible YAML/JSON typo — they hit `invalid-type-declaration` and get a correct but un-pinpointed message, missing the "Check the `type` override for rule … in your Thymian config file" hint that a coverage violation from the same override would get. Passing `typeOverrideSuggestions(rule, options)` into the `assertRuleTypeDeclaration` call too would close this gap.

3. **One of the four "invalid chain shapes" lacks a permanent compile-time regression test (minor).** `rule-builder.test.ts` has `@ts-expect-error` assertions for `done()` with zero/partial coverage and for `rule()` on an informational builder — three of the four shapes the PR description says were verified against a manual "probe file." The fourth, `.type('informational', 'static')` resolving to the `InformationalMixedWithExecutableTypes` marker at the *type* level, only has a runtime-throw test (`.type()` throws before returning, so the runtime test alone doesn't prove the compile-time marker type is still correct). Since the probe file isn't checked in, a future refactor could silently regress that specific type-level guarantee without CI catching it. Suggest adding a checked-in `@ts-expect-error` case for this shape too.

4. **Nit — scope creep, harmless.** The `plugin-reporter` CSV test fix (`,af-1,` vs `af-1` to avoid colliding with UUID substrings) is unrelated to this PR's subject and package. It's a legitimate, well-justified flakiness fix, just worth calling out in case the team prefers such drive-bys split into their own PR/commit.

5. **Nit — no exhaustiveness guard on the violation switch.** `describeRuleExecutionInvariantViolation`'s `switch` has no `default`. If a new `RuleExecutionInvariantViolation` variant is added later without updating this function, TS will only catch the gap if `noImplicitReturns` is on; otherwise it silently returns `undefined` at runtime. The file already uses an explicit exhaustiveness trick elsewhere (`Satisfies`/`AssertRuleBuilderConformance`) — a `default` case with a `never`-typed exhaustiveness check here would be consistent and cheap insurance.

### Correctness verification performed

- Traced `checkRuleExecutionInvariant`'s dedup (`[...new Set(types)]`) and confirmed duplicate declared types (e.g. `['static','static']`) don't produce duplicated entries in the missing-types message.
- Confirmed `ruleFnPropertyByType`'s `static→lintRule / analytics→analyzeRule / test→testRule` mapping matches `Rule`'s actual optional fields in `rule.ts`, and `ruleTypes` in `rule-meta.ts` matches the shape check's known-types list.
- Confirmed the new docs page's URL casing (`references/errors/invalid-rule-error/` lowercase vs. on-disk `references/Errors/` capitalized) matches the existing, working convention used by every other error page in the repo — not a broken link.
- Confirmed no double-application of config overrides between the inline-rules loop and the recursive glob/pattern path.

### Risk assessment

No blocking correctness, security, or performance issues found. This is an additive, well-tested invariant with a sensible escape hatch (disable or downgrade to `informational`) for rules you don't control. The points above are worth a follow-up pass but shouldn't block merge.

---
🤖 Review generated with [Claude Code](https://claude.com/claude-code)

BaggersIO (approved):
Approving — the notes in my review above are non-blocking suggestions for a follow-up, not blockers.

PR #341, opened 2026-08-03
feat(repo): global site navigation & responsive chrome (Epic 13)

BaggersIO:
## Summary

Ships the full navigation/chrome rework as one PR (Epic 13, Stories 13.1-13.3 - thymian-internal#468, #469, #470):

**13.1 - Navigation data model + three-band desktop header**
- `src/data/navigation.ts`, a single typed source of truth for site nav links, driving the header pill row instead of hardcoded anchors.
- Reorders the pill row to Events / Resources / Blog / Contact / Enterprise (CTA moves from first to last), adds the net-new Blog pill, switches the nav landmark to `aria-label="Site"`.
- Fixes the 800-1150px logo-clipping bug and adds compact-then-72rem pill sizing.
- Pins the Enterprise CTA's dark-mode fills via new `--tnav-*` custom properties; guards a CSS specificity collision between the active-pill and CTA rules.
- Re-gates header socials to the >=1152px band and adds a net-new footer social row for 800-1152px; removes the header `LanguageSelect`.

**13.2 - Sub-800px chrome: menu button, shared panel, drawer parity**
- New shared `<thymian-site-nav>` component (menu button + panel) on every `hasSidebar=false` route (home, enterprise, events, resources) - previously a complete mobile dead end below 800px.
- ARIA state lives on the button itself (not a host element); Esc closes + returns focus; any link activation closes the panel (including the non-navigating mailto Contact link); inert + scroll-locked while open; resets cleanly across the 800px breakpoint.
- Docs/blog routes keep Starlight's native drawer, now re-fed from `navigation.ts` instead of a separate 3-link hardcoded fork that was missing Events and Resources entirely.
- Also fixes two pre-existing bugs surfaced while building this: an unlayered CSS declaration that silently defeated the pill row's `sl-hidden`/`md:sl-flex` visibility utilities (pills were never actually hidden below 800px), and a scroll-lock mechanism that targeted the wrong DOM element for this site's scrolling model.

**13.3 - Navigation e2e viewport matrix + mobile CI project**
- Adds a real mobile Playwright project (390x844, chromium-native context flags) alongside the existing Desktop Chrome project - previously no mobile assertion could ever run in CI.
- Three viewport x route matrix suites (V1/V2/V3 x 6 route classes) covering every navigation AC, plus 46 axe WCAG 2.1 AA scans in both themes, including the sub-800px panel-open and drawer-open states for the first time.
- Surfaced (and mostly fixed) several pre-existing, unrelated accessibility gaps this broadened coverage exposed - tracked in thymianofficial/thymian#347.

## Post-review fixes (manual QA on this PR)

- `deb0b080` - the mobile menu button was its own top-level flex child of the header, so `justify-content: space-between` spread it away from the search icon as the viewport widened (50px gap at 390px wide, growing to 255px at 799px). Moved it to render inside the same flex group as the search icon instead, so it shares a fixed gap.
- `cc31dec6` - the menu icon was pinned top-left instead of centered, and looked too small, once the button grew past Starlight's native 32px. Added explicit centering.
- `e463ec11` - **deliberate, acknowledged deviation, not a bug**: at explicit user request, shrunk the button from 44px back to 32px to match the native docs-page button exactly. This reintroduces the WCAG 2.5.8 44px touch-target gap that 13.2's own Dev Notes originally required the button NOT have (that's why it was 44px in the first place). Flagged here and in 13-469's Dev Agent Record for reviewer visibility.

**Known open item**: user reports the search (magnifier) icon is misaligned on non-docs routes; not yet reproduced despite measuring bounding boxes/vertical centering across widths 320-799px and both themes (all identical to the docs page in every check so far). Still being investigated - will push a follow-up commit once root-caused.

## Test plan
- [x] `npx astro build` - links validator passes, 121 pages
- [x] `npx astro check` - 14 errors (baseline established in 13.1 was 13; the +1 is the same pre-existing virtual-module-resolution tooling quirk affecting every `virtual:starlight/components/*` import in this codebase, not a new defect class)
- [x] `npx vitest run` - 157/157
- [x] `npx playwright test` (both projects) - 150 tests, 145 passed, 5 tracked skips (thymianofficial/thymian#347, out of scope for this epic), re-run multiple times with zero flakiness, ~25s wall-clock
- [x] `npx nx affected -t lint` - 0 errors across 17 projects
- [x] `npx nx format:check` - clean except pre-existing, untouched files from Epic 9

### DIFF
--- a/astro-docs/e2e/accessibility.spec.ts
+++ b/astro-docs/e2e/accessibility.spec.ts
@@ -1,5 +1,6 @@
-import AxeBuilder from '@axe-core/playwright';
-import { expect, type Page, test } from '@playwright/test';
+import { expect, test } from '@playwright/test';
+
+import { expectNoViolations, forceDarkTheme } from './navigation/nav-shared';
 
 /**
  * Automated accessibility tests using axe-core.
@@ -11,41 +12,6 @@ import { expect, type Page, test } from '@playwright/test';
  *
  * The Playwright config starts the Astro preview server automatically.
  */
-
-/** Run the axe WCAG 2 A/AA scan and assert zero violations. */
-async function expectNoViolations(page: Page): Promise<void> {
-  const results = await new AxeBuilder({ page })
-    .withTags(['wcag2a', 'wcag2aa', 'wcag21a', 'wcag21aa'])
-    .analyze();
-
-  const violations = results.violations.map((v) => ({
-    id: v.id,
-    impact: v.impact,
-    description: v.description,
-    nodes: v.nodes.length,
-  }));
-
-  expect(violations, `Found ${violations.length} a11y violation(s)`).toEqual(
-    [],
-  );
-}
-
-/**
- * Force Starlight's dark theme. Starlight keys its theme off a `data-theme`
- * attribute on `<html>` and persists the choice in the `starlight-theme`
- * localStorage key; seeding both before navigation means Starlight's own inline
- * theme script applies dark on load (and stays dark through any client re-run).
- */
-async function forceDarkTheme(page: Page): Promise<void> {
-  await page.addInitScript(() => {
-    try {
-      localStorage.setItem('starlight-theme', 'dark');
-    } catch {
-      /* localStorage may be unavailable; the attribute below still applies. */
-    }
-    document.documentElement.dataset.theme = 'dark';
-  });
-}
 
 const pages = [
   { name: 'Home', path: '/' },
--- a/astro-docs/e2e/navigation-header.spec.ts
+++ b/astro-docs/e2e/navigation-header.spec.ts
@@ -1,0 +1,248 @@
+import { expect, type Page, test } from '@playwright/test';
+
+/**
+ * Component-level assertions for spec ACs 2, 4, 5 (V2/V3 legs), 13, 14, 15
+ * (docs/ux-site-navigation/EXPERIENCE.md @ commit 5823816). Runs under the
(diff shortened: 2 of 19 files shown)


matthyk (approved):
Approving — clean, well-tested nav rework. Two minor, non-blocking a11y nits on the same `matchMedia` resize-close path in `SiteNav.astro`, both stemming from `.addEventListener('change', () => this.setExpanded(false))` not doing what the `Escape` path does. Neither triggers on a real phone (fixed viewport); both need a responsive resize / tablet rotation across the 800px boundary with the panel open.

**1. Focus is dropped to `<body>` when the panel is closed by a resize while focus is inside it.** `setExpanded(false)` sets `panel.hidden = true`, so a focused nav `<a>` becomes `display:none` and focus falls to the document body — the next Tab restarts from the top. `closeOnEscape` handles this correctly by calling `this.button.focus()`; the resize path doesn't. One-line fix: restore focus to the button when the panel contained `document.activeElement`.

**2. The panel stays painted mid-viewport for the ~200ms close transition on that same resize.** Because the close is animated (`transition-property` includes `display` with `transition-behavior: allow-discrete`), `display:none` is deferred to the end of the fade, so the full-width fixed panel briefly overlays the now-desktop layout (with no visible toggle affordance, since the button is already `display:none`). Purely cosmetic, self-clears.

Both are below the merge bar — noting for a follow-up, not gating.

PR #344, opened 2026-08-03
feat(rules-rfc-9110): educational explanations for conditional requests, content negotiation & authentication (§13, §12, §11)

matthyk:
## What

Adds a user-facing `.explanation(...)` — a plain-language description of the rule plus why it matters — to **33 rules across the Conditional Requests, Content Negotiation, and Authentication chapters** of `@thymian/rules-rfc-9110`. Part of the work for thymianofficial/thymian-internal#289.

**RFC 9110 §13 — Conditional Requests**, **§12 — Content Negotiation**, and **§11 — HTTP Authentication.**

- **Conditional Requests (§13):** `If-Match` (§13.1.1), `If-None-Match` (§13.1.2), `If-Modified-Since` (§13.1.3), `If-Unmodified-Since` (§13.1.4), `If-Range` (§13.1.5), and precondition evaluation / ordering (§13.2).
- **Content Negotiation (§12):** `Accept`, `Accept-Charset`, `Accept-Encoding`, and `Vary` (§12.5).
- **HTTP Authentication (§11):** authentication challenges & credentials, protection space / realm, `WWW-Authenticate` / `Authorization`, and authentication-parameter syntax.

## Scope & guarantees

## Verification

### DIFF
--- a/packages/rules-rfc-9110/src/rules/authentication/origin-server/authentication-info/proxy-must-not-modify-authentication-info.rule.ts
+++ b/packages/rules-rfc-9110/src/rules/authentication/origin-server/authentication-info/proxy-must-not-modify-authentication-info.rule.ts
@@ -17,6 +17,9 @@ export default httpRule('rfc9110/proxy-must-not-modify-authentication-info')
   .description(
     'A proxy forwarding a response is not allowed to modify the Authentication-Info field value in any way.',
   )
+  .explanation(
+    "When a proxy passes a response along, it must forward the Authentication-Info header exactly as it received it, without changing, reordering, adding, or dropping anything in the value. This field carries scheme-specific information the origin server sends after accepting the client's credentials (for example a finalization message that authenticates the server). If a proxy alters it, that end-to-end signal can be corrupted and the authentication handshake between client and origin server can break.",
+  )
   .appliesTo('proxy')
   .rule((ctx) =>
     ctx.validateCapturedHttpTraces((trace, location) => {
--- a/packages/rules-rfc-9110/src/rules/authentication/origin-server/authorization/proxy-must-not-modify-authorization.rule.ts
+++ b/packages/rules-rfc-9110/src/rules/authentication/origin-server/authorization/proxy-must-not-modify-authorization.rule.ts
@@ -16,6 +16,9 @@ export default httpRule('rfc9110/proxy-must-not-modify-authorization')
   .description(
     'A proxy forwarding a request MUST NOT modify any Authorization header fields in that request.',
   )
+  .explanation(
+    "When a proxy relays a request onward, it must leave the Authorization header untouched and pass the client's credentials through byte-for-byte. The Authorization field is how a user agent proves its identity to the origin server, and many authentication schemes bind those credentials to their exact bytes. If a proxy rewrites the header, the origin server sees different credentials than the client sent, so authentication fails and the client is wrongly rejected.",
+  )
   .appliesTo('proxy')
   .rule((ctx) =>
     ctx.validateCapturedHttpTraces((trace, location) => {
--- a/packages/rules-rfc-9110/src/rules/authentication/origin-server/www-authenticate/proxy-must-not-modify-www-authenticate.rule.ts
+++ b/packages/rules-rfc-9110/src/rules/authentication/origin-server/www-authenticate/proxy-must-not-modify-www-authenticate.rule.ts
@@ -16,6 +16,9 @@ export default httpRule('rfc9110/proxy-must-not-modify-www-authenticate')
   .description(
     'A proxy forwarding a response MUST NOT modify any WWW-Authenticate header fields in that response.',
   )
+  .explanation(
+    "When a proxy passes a response back toward the client, it must forward every WWW-Authenticate header exactly as it received it, without editing, adding, or removing any challenge. This field is the origin server's authentication challenge, telling the client which schemes, realms, and parameters to use. If a proxy alters it, the client may build the wrong credentials or pick the wrong scheme, breaking the authentication exchange with the origin server.",
+  )
   .appliesTo('proxy')
   .rule((ctx) =>
     ctx.validateCapturedHttpTraces((trace, location) => {
--- a/packages/rules-rfc-9110/src/rules/authentication/origin-server/www-authenticate/server-may-send-www-authenticate-in-other-responses.rule.ts
+++ b/packages/rules-rfc-9110/src/rules/authentication/origin-server/www-authenticate/server-may-send-www-authenticate-in-other-responses.rule.ts
@@ -12,6 +12,9 @@ export default httpRule(
   .description(
     'A server MAY generate a WWW-Authenticate header field in other response messages to indicate that supplying credentials (or different credentials) might affect the response.',
   )
+  .explanation(
+    'A server is allowed to include a WWW-Authenticate header on responses other than a 401, as an optional hint that sending credentials, or different credentials, might change the outcome. It is not required to do so. This lets a server nudge a client toward authenticating even when it did not outright reject the request, which can help clients discover that a better-authorized response is available.',
+  )
   .appliesTo('server')
   .rule((ctx) =>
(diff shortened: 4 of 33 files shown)


BaggersIO (approved):
## Overview

This PR adds a `.explanation(...)` call to all 33 user-facing rules across three RFC 9110 chapters: **Conditional Requests (§13)** — If-Match, If-None-Match, If-Modified-Since, If-Unmodified-Since, If-Range, and precondition evaluation/precedence; **Content Negotiation (§12)** — Accept, Accept-Charset, Accept-Encoding, Vary; and **HTTP Authentication (§11)** — challenges/credentials, realm, WWW-Authenticate/Authorization/Authentication-Info, and auth-parameter syntax. Part of the broader thymianofficial/thymian-internal#289 effort (3rd of 5 sibling PRs split by chapter).

Diff shape matches the PR description exactly: 33 files changed, +99/-0 — a uniform 3-line `.explanation(...)` block added to each file, always inserted right after the existing `.description()`/`.summary()` call and before `.appliesTo()`/`.rule()`/`.overrideTest()`. No other lines are touched anywhere in the diff.

## Code quality & style

- The prose consistently explains **what** the rule requires **and why it matters**, rather than paraphrasing `.description()`/`.summary()`. E.g. `proxy-must-not-modify-authentication-info`: the description just says a proxy can't modify the field; the explanation adds what the field actually carries (a post-credential-acceptance, scheme-specific message, e.g. a server-authentication finalization) and the concrete failure mode if violated (the end-to-end signal is corrupted and the handshake breaks). That pattern holds consistently across the set.
- Length: computed word counts for all 33 explanations programmatically. Range is 66–95 words, average ~77.5 — all but one comfortably inside the ~40–90 target (see suggestion below).
- Voice: 32/33 explanations are third person ("a client...", "the server..."). Exactly one (`realm-parameter-must-use-quoted-string-syntax`) uses second person ("When you emit a realm authentication parameter..."). Per the known, already-tracked cross-effort style split, this is a minor non-blocking point, and at 1/33 it's clearly not pervasive enough to read as inconsistent within this PR specifically.
- Grammar/clarity: clean throughout — no typos or garbled sentences in any of the 33.

## Specific suggestions (non-blocking)

1. **`sender-should-send-q-parameter-last`** — 95 words, the only one over the ~90-word guideline. Consider trimming the last sentence ("For the same reason, a media type is not allowed to define its own parameter named q...") or splitting it off. Content itself is accurate (see Correctness), this is purely a length nit.
2. **`server-must-ignore-preconditions-for-non-2xx-412-responses`** — first sentence reads "A server **should** only evaluate a request's preconditions when..." but the underlying RFC 9110 §13.2.1 rule is a MUST ("A server MUST ignore all received preconditions if..."). The explanation's second sentence correctly uses "must," so the paragraph as a whole isn't misleading, but the first sentence's "should" is a softer normative register than the rule warrants. Copilot's automated pass flagged the same line — I independently agree it's a good tightening opportunity, not a blocker. Suggest swapping to "must."
3. **`origin-server-should-generate-vary-header`** — "causing cache poisoning and wrong content being served" is a bit more security-loaded than RFC 9110's own framing for this rule (the RFC just describes a cache serving a mismatched cached representation). Not inaccurate — missing/incorrect Vary is a legitimate contributor to cache-based cross-user content leakage — just slightly stronger phrasing than strictly necessary. Optional softening.
4. Two incidental, **pre-existing, out-of-scope** observations noticed while cross-checking (neither introduced by this PR — the lines aren't touched by this diff — so not blocking, just flagging for a possible follow-up ticket):
   - `if-unmodified-since/cache-or-intermediary-may-ignore-if-unmodified-since` and `if-unmodified-since/client-may-send-if-unmodified-since-header` both have a pre-existing `.url()` pointing at `#section-13.1.1` (If-Match) instead of `#section-13.1.4` (If-Unmodified-Since) — looks like a copy/paste artifact that predates this PR. The new `.explanation()` text in both is correctly grounded in §13.1.4 (the rule's actual topic) regardless, so this doesn't affect explanation correctness, but the `.url()` itself looks wrong.
   - `if-range/client-must-not-generate-if-range-header-containing-http-date`'s new explanation correctly states the RFC's nuanced exception (HTTP-date allowed only with no entity-tag *and* a strong validator), but the rule's pre-existing `.summary()` reads as an unconditional "MUST NOT ... containing an HTTP-date" with no exception — Copilot flagged this as a "contradiction." I'd frame it the other way: the new explanation is *more* accurate than the old summary; a nice target for a follow-up tightening the summary, not a defect introduced here.

## Issues & risks

**Correctness (main risk surface for this PR).** I read RFC 9110 §11, §12, and §13 directly from the local text and checked all 33 explanations against the normative text and surrounding rationale in the section each rule's (pre-existing) `.url()` cites. Representative checks:
- `sender-should-send-q-parameter-last` (§12.5.1 Accept) — the rule flagged in the pre-cut adversarial pass for an invented "obsolete accept-parameter grammar." Current text has no such grammar; it correctly states recipients treat any `q`-named parameter as weight regardless of position, and that the media-type registry disallows a parameter literally named `q` — both directly from §12.5.1. Confirmed accurate; concur with the earlier spot-check.
- `client-must-not-generate-if-range-with-weak-etag` (§13.1.5) — correctly reflects that If-Range uses the *strong* comparison function (§8.8.3.2) and that a weak tag only promises semantic equivalence, not byte-identity. Accurate.
- `origin-server-must-not-perform-method-when-if-match-fails` / `-if-unmodified-since-fails` (§13.1.1 / §13.1.4) — both correctly tie the MUST-NOT-perform-method rule to the "lost update" problem the RFC itself names in the surrounding prose. Accurate.
- `recipient-must-ignore-if-modified-since-when-if-none-match-present` and its If-Match/If-Unmodified-Since analog (§13.1.3/§13.1.4) — both correctly state entity-tag comparison is treated as the more accurate/precise test and that the two headers are combined only for older-intermediary interop, matching the RFC's own wording almost verbatim.
- `origin-server-should-send-response-without-content-coding` and `server-must-not-include-accept-encoding-for-non-content-coding-415-errors` (§12.5.3) — both correctly preserve the RFC's SHOULD/MUST-NOT strength and the identity-unacceptable exception.
- `proxy-must-not-generate-vary-wildcard` and `origin-server-should-generate-vary-header` (§12.5.5) — both accurate on what `*` and Vary do and the caching consequences.
- Authentication chapter (§11.2, §11.5, §11.6.x) — all 6 explanations accurately reflect their respective subsections; `realm-parameter-must-use-quoted-string-syntax`'s "many long-deployed clients only accept the quoted form" rationale isn't verbatim in §11.5 but is directly supported by §11.2's adjacent statement that sender formats are sometimes restricted "when it is known that deployed implementations will fail when encountering one of the two formats" — not a fabrication.

No misstated normative requirements or invented/unsupported rationale found anywhere across the 33.

**Convention-following.** Confirmed via the diff and by fetching all 33 files at the PR head commit:
- Purely additive: +99/-0 across 33 files = exactly 3 added lines per file, no deletions, matching the PR description.
- Only `.explanation(...)` calls were added — checked every hunk; no other builder call, `.rule()` body, or metadata field was touched anywhere.
- No informational-only rules touched — checked `.type(...)` on all 33 fetched files; every one includes `'static'`, `'test'`, and/or `'analytics'` (several combine two or three), never `'informational'` alone or at all.

**Test coverage.** N/A — doc/metadata-only change (a new string literal passed to an existing, already-tested builder method). No test files are among the 33 changed files, and none should be, since `.explanation()` isn't exercised by `.rule()` validation logic.

**Security.** No `.rule()` validation-logic function bodies were touched anywhere in the diff — confirmed by inspecting every hunk's surrounding context. The 6 authentication-chapter explanations (proxy non-modification of Authentication-Info/Authorization/WWW-Authenticate, WWW-Authenticate-on-other-responses, auth-parameter uniqueness, realm quoting) were read carefully given the topic; none give guidance that could mislead a reader into an insecure practice — they consistently reinforce byte-exact forwarding of credentials/challenges and correct scoping of realms, matching §11. No scope creep found.

## Decision

Approving. No misstated RFC requirements, fabricated rationale, wrongly-touched rules, or off-scope edits — just the minor, non-blocking prose points above.

PR #345, opened 2026-08-03
feat(rules-rfc-9110): educational explanations for representation metadata, message context & identifiers (§8, §10, §4)

matthyk:
## What

Adds a user-facing `.explanation(...)` — a plain-language description of the rule plus why it matters — to **39 rules across the Representation Metadata, Message Context, and Identifiers chapters** of `@thymian/rules-rfc-9110`. Part of the work for thymianofficial/thymian-internal#289.

**RFC 9110 §8 — Representation Data and Metadata**, **§10 — Message Context**, and **§4 — Identifiers.**

- **Representation Data & Metadata (§8):** `Content-Type` (§8.3), `Content-Encoding` (§8.4), `Content-Language`, `Content-Length` (§8.6), and validator fields — `ETag` (§8.8.3) and `Last-Modified` (§8.8.2).
- **Message Context (§10):** request-context fields `Expect`, `From`, `Referer`, `TE`, `User-Agent` (§10.1) and response-context field `Server` (§10.2).
- **Identifiers (§4):** the `http` and `https` URI schemes (§4.2), deprecation of `userinfo` in `http(s)` URIs (§4.2.4), and URI references.

## Scope & guarantees

## Verification

### DIFF
--- a/packages/rules-rfc-9110/src/rules/identifiers/http/recipient-must-reject-http-uri-without-host.rule.ts
+++ b/packages/rules-rfc-9110/src/rules/identifiers/http/recipient-must-reject-http-uri-without-host.rule.ts
@@ -9,6 +9,9 @@ export default httpRule('rfc9110/recipient-must-reject-http-uri-without-host')
   .description(
     `A recipient that processes a 'http' URI reference with empty host MUST reject it as invalid.`,
   )
+  .explanation(
+    'An http URI must name a host (like example.com) in its authority component; that host identifies the origin server that owns the resource. If a server receives an http URI whose host is empty, it must treat the URI as invalid and reject it rather than trying to process it. An http URI with no host does not identify any server, so accepting it would leave the request pointing nowhere and could open the door to ambiguity or misrouting.',
+  )
   .appliesTo('server')
   .rule((ctx, opts, logger) =>
     ctx.validateHttpTransactions(
--- a/packages/rules-rfc-9110/src/rules/identifiers/http/sender-must-not-generate-http-uri-with-empty-host.rule.ts
+++ b/packages/rules-rfc-9110/src/rules/identifiers/http/sender-must-not-generate-http-uri-with-empty-host.rule.ts
@@ -11,6 +11,9 @@ export default httpRule(
   .description(
     `A sender MUST NOT generate an 'http' URI with an empty host identifier.`,
   )
+  .explanation(
+    'Whenever a client produces an http URI, that URI must include a non-empty host (such as example.com), because the host is what identifies the origin server responsible for the resource. Emitting an http URI with an empty host is forbidden. A hostless URI points to no server at all, so any recipient would be unable to route or resolve it and is required to reject it as invalid, breaking the request.',
+  )
   .appliesTo('client')
   .rule((ctx, opts, logger) =>
     ctx.validateHttpTransactions(
--- a/packages/rules-rfc-9110/src/rules/identifiers/https/recipient-must-reject-https-uri-without-host.rule.ts
+++ b/packages/rules-rfc-9110/src/rules/identifiers/https/recipient-must-reject-https-uri-without-host.rule.ts
@@ -9,6 +9,9 @@ export default httpRule('rfc9110/recipient-must-reject-https-uri-without-host')
   .description(
     `A recipient that processes a 'https' URI reference with empty host MUST reject it as invalid.`,
   )
+  .explanation(
+    'An https URI must name a host (like example.com) in its authority component, because that host identifies the origin server that owns the resource. If a server processes an https URI whose host is empty, it must treat the URI as invalid and reject it rather than acting on it. A hostless https URI identifies no server, so accepting it would leave the request pointing nowhere and could enable ambiguity or misrouting.',
+  )
   .appliesTo('server')
   .rule((ctx, opts, logger) =>
     ctx.validateHttpTransactions(
--- a/packages/rules-rfc-9110/src/rules/identifiers/https/sender-must-not-generate-https-uri-with-empty-host.rule.ts
+++ b/packages/rules-rfc-9110/src/rules/identifiers/https/sender-must-not-generate-https-uri-with-empty-host.rule.ts
@@ -11,6 +11,9 @@ export default httpRule(
   .description(
     `A sender MUST NOT generate an 'https' URI with an empty host identifier.`,
   )
+  .explanation(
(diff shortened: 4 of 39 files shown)


BaggersIO (approved):
## Overview

Adds a user-facing `.explanation(...)` to 39 rules spanning three RFC 9110 chapters: **§8 Representation Data and Metadata** (Content-Type, Content-Encoding, Content-Language, Content-Length, ETag, Last-Modified), **§10 Message Context** (Expect, From, Referer, TE, User-Agent, Server), and **§4 Identifiers** (http/https URI schemes, userinfo deprecation, URI length). Per the diff, this is purely additive: 117 additions / 0 deletions across 39 files, one `.explanation()` call inserted per file, nothing else touched.

## Code quality & style

- The prose is genuinely explanatory rather than a restatement of `.description()` — each one restates the rule in plain language and then adds a distinct "why" (cache-correctness, request-smuggling, phishing/privacy, interoperability, fingerprinting, etc.) grounded in the cited RFC section.
- Length: measured all 39 programmatically — range is 59-82 words, average 72, comfortably inside the ~40-90 target. No outliers either direction.
- Grammar/clarity: clean throughout everything I read; no typos or awkward constructions spotted.
- Voice: third person is the majority, but this PR skews a bit more second-person ("you"/"your") than the effort-wide average — 14/39 (36%) here, clustered noticeably in `validator-fields/` (e.g. `origin-server-should-send-etag`, `origin-server-should-send-last-modified`, `sender-may-send-etag-in-trailer`, `origin-server-should-avoid-backslash-in-entity-tags`, `etag-must-differ-for-different-content-encodings` all use "you"/"your", while sibling rules elsewhere in the PR stay third-person). This is the known, tracked, not-yet-fixed style split — non-blocking per that context, but the concentration is dense enough in this folder that it reads as slightly inconsistent file-to-file, worth a mention for the eventual pass.

## Specific suggestions

- Minor (non-blocking): `recipient-should-treat-x-gzip-as-gzip` describes "x-gzip" as "an older, deprecated alias." RFC 9110 §8.4.1.3 itself doesn't use the word "deprecated" for x-gzip, only that a recipient SHOULD treat it as equivalent to gzip. Not incorrect (the "x-" prefix convention for unregistered tokens is well-established HTTP history), just slightly more asserted than the letter of the cited section — could soften to "legacy" if a reviewer wants to stay strictly literal.
- FYI, pre-existing and out of this PR's scope: `content-language-may-be-applied-to-any-media-type.rule.ts` and `multiple-languages-may-be-listed-for-multiple-audiences.rule.ts` have no `.url()` call at all (confirmed via the diff this predates the PR — nothing was removed). I independently verified both explanations against RFC 9110 §8.5 and they're accurate, but the missing `.url()` means these two can't be mechanically traced the way the other 37 can. Might be worth a quick follow-up ticket to backfill it.

## Issues & risks

### Correctness

This is the main risk surface for this content type, so I sampled well beyond the requested 12-15 — I read the full text of RFC 9110 §4, all of §8 (including every Validator Fields subsection), and all of §10 (both request- and response-context fields), and checked all 39 explanations against the section each rule's own `.url(...)` cites (for the 2 files lacking `.url()`, against the section located via the RFC's table of contents instead). Highlights:

- **Identifiers (§4)**: `recipient-must-reject-http(s)-uri-without-host` / `sender-must-not-generate-http(s)-uri-with-empty-host` (§4.2.1/4.2.2) — accurate, "host identifies the origin server" matches the RFC directly. `sender-must-not-generate-userinfo-in-uri` / `recipient-should-treat-userinfo-in-uri-from-untrusted-source-as-error` (§4.2.4) — accurate, phishing/deprecation rationale matches directly. `sender-recipient-should-support-8000-octet-uris` (§4.1) — accurate; added interoperability rationale is a reasonable, non-contradicted inference.
- **Message Context (§10)**: both Expect rules (§10.1.1) — accurate, "(presumably large)" language lifted directly from the RFC. Both From rules (§10.1.2) — accurate, matches "excessive, unwanted, or invalid requests" and the privacy/logging rationale almost verbatim. All three Referer rules (§10.1.3) — accurate, correctly reflects fragment/userinfo stripping and the secure-to-insecure/cross-origin privacy rules including the "unless explicitly allows" exception. TE (§10.1.4) — accurate hop-by-hop framing. All three User-Agent rules + Server (§10.1.5/§10.2.4) — accurate, correctly cites the fingerprinting/latency rationale and the internal-implementation-disclosure security caveat directly from the RFC.
- **Representation Metadata (§8)**: both Content-Type rules (§8.3) — accurate, including the "privilege escalation" MIME-sniffing risk quoted directly from the RFC. Content-Encoding rules (§8.4) — accurate. Both Content-Language rules (§8.5) — accurate, including the Maori/English multi-audience example and the "Latin primer" single-audience counter-example, both lifted directly from the RFC. All 8 Content-Length rules (§8.6) — accurate throughout, including the integer-overflow/request-smuggling rationale and the 1xx/204/CONNECT/HEAD/304 special cases; cross-checked `user-agent-should-not-send-content-length-without-content`'s method list (GET/HEAD/DELETE/CONNECT/OPTIONS/TRACE) against the rule's own untouched `.description()` and validation logic — consistent. All 4 ETag rules + both Last-Modified rules (§8.8.2/§8.8.3) — accurate; `etag-must-differ-for-different-content-encodings` is a particularly good paraphrase of the exact Note in §8.8.3.3 about cache updates and range requests.

No misstated RFC requirements and no fabricated/unsupported rationale found anywhere in the 39.

### Convention-following

Confirmed via the diff and full-file reads of all 39 touched files at the PR head commit:
- 0 deletions; exactly 39 hunks, each a 3-line `.explanation(...)` insertion.
- Every added line is either the `.explanation(` call, its string literal, or the closing `)` — no other builder calls, no rule logic, no `.description`/`.summary`/`.url` edits anywhere.
- `.type()` checked on all 39 files: every one resolves to some combination of `static`/`test`/`analytics`; none is purely `informational`.
- No duplicate/copy-pasted explanation text across the 39 (checked programmatically).

### Test coverage

N/A as expected — this is a doc-only/metadata-only change to a string field that isn't consumed by rule logic or assertions. The existing rule test suites (which exercise `.rule()`/`.overrideAnalyticsRule()` behavior) are untouched and unaffected by this change, so no new tests are needed here.

### Security

Confirmed no `.rule()` / `.overrideAnalyticsRule()` validation-logic function bodies were touched anywhere in the diff — every hunk is a pure `.explanation()` insertion, nothing else. No scope creep found.

## Decision

**Approving.** All 39 explanations are accurate, well-grounded in their cited RFC 9110 sections, correctly scoped to non-informational rules, and the change is cleanly additive with zero deletions. Non-blocking notes for a future pass: the second-person voice concentration in this PR's `validator-fields/` cluster, and backfilling `.url()` on the two Content-Language rules.

PR #346, opened 2026-08-03
feat(rules-rfc-9110): educational explanations for routing, message abstraction & fields (§7, §6, §5)

matthyk:
## What

Adds a user-facing `.explanation(...)` — a plain-language description of the rule plus why it matters — to **35 rules across the Routing, Message Abstraction, and Fields chapters** of `@thymian/rules-rfc-9110`. Part of the work for thymianofficial/thymian-internal#289.

**RFC 9110 §7 — Routing**, **§6 — Message Abstraction**, and **§5 — Fields.**

- **Routing (§7):** target resource, `Host` / authority, `Connection` and connection options (§7.6.1), `Max-Forwards` (§7.6.2), `Via` (§7.6.3), `Upgrade` (§7.8), and message forwarding & transformations (§7.6–7.7).
- **Message Abstraction (§6):** message metadata and the `Trailer` field / trailer-section handling (§6.5–6.6).
- **Fields (§5):** field names (§5.1), field order (§5.3), and field-value construction — date/time formats and list-based field syntax (§5.6).

## Scope & guarantees

## Verification

### DIFF
--- a/packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-generate-timestamps-in-imf-fixdate-format.rule.ts
+++ b/packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-generate-timestamps-in-imf-fixdate-format.rule.ts
@@ -79,6 +79,9 @@ export default httpRule(
   .summary(
     'Sender MUST generate HTTP-date timestamps in the IMF-fixdate format.',
   )
+  .explanation(
+    "Whenever a sender writes a date in a header defined as HTTP-date (such as Date, Expires, or Last-Modified), it must use exactly one format: IMF-fixdate, for example 'Sun, 06 Nov 1994 08:49:37 GMT'. Two older date formats exist but must never be generated. Pinning senders to this single fixed layout, always in GMT, means every recipient can parse timestamps reliably, which is essential for caching, expiration, and conditional requests to work correctly across implementations.",
+  )
   .rule((ctx) =>
     ctx.validateHttpTransactions(
       or(...httpDateResponseHeaders.map((name) => responseHeader(name))),
--- a/packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-not-generate-additional-whitespace-in-http-date.rule.ts
+++ b/packages/rules-rfc-9110/src/rules/fields/defining-field-values/date-time-formats/sender-must-not-generate-additional-whitespace-in-http-date.rule.ts
@@ -61,6 +61,9 @@ export default httpRule(
   .summary(
     'Sender MUST NOT generate additional whitespace in HTTP-date beyond the grammar.',
   )
+  .explanation(
+    'An HTTP-date must contain only the exact single spaces the grammar prescribes; a sender must not add extra whitespace, such as tabs, doubled spaces, or leading or trailing padding. The date format is deliberately rigid so recipients can parse it with a simple fixed pattern. Even harmless-looking extra spaces can cause a strict parser to reject the timestamp, which then breaks caching and expiration logic that depends on reading these dates.',
+  )
   .rule((ctx) =>
     ctx.validateHttpTransactions(
       or(...httpDateResponseHeaders.map((name) => responseHeader(name))),
--- a/packages/rules-rfc-9110/src/rules/fields/defining-field-values/lists/sender-must-not-generate-empty-list-elements.rule.ts
+++ b/packages/rules-rfc-9110/src/rules/fields/defining-field-values/lists/sender-must-not-generate-empty-list-elements.rule.ts
@@ -52,6 +52,9 @@ export default httpRule('rfc9110/sender-must-not-generate-empty-list-elements')
     'In any production that uses the list construct, a sender MUST NOT generate empty list elements.',
   )
   .summary('Sender MUST NOT generate empty list elements.')
+  .explanation(
+    'For any header defined as a comma-separated list (such as Vary or Allow), a sender must not emit empty entries, meaning no leading comma, no trailing comma, and no two commas with nothing but whitespace between them. Although recipients are told to tolerate such gaps, producing them is sloppy and risks confusing stricter parsers. Keeping list values clean, with a real value between every comma, avoids ambiguity and interoperability problems across differing implementations.',
+  )
   .rule((ctx) =>
     ctx.validateHttpTransactions(
       or(...listTypedResponseHeaders.map((name) => responseHeader(name))),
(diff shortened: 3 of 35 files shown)


BaggersIO (approved):
## Overview

This PR adds a single `.explanation(...)` call to each of 35 rules spanning RFC 9110 Routing (§7), Message Abstraction (§6), and Fields (§5) chapters of `@thymian/rules-rfc-9110`, as part of the umbrella effort in thymian-internal#289. The diff is exactly what the description promises: 105 additions / 0 deletions across 35 files, one 3-line `.explanation(...)` block inserted per file between the existing `.summary(...)` and the next builder call (`.appliesTo(`, `.rule(`, or `.overrideAnalyticsRule(`). No rule logic, `.description`, `.summary`, `.url`, `.severity`, `.type`, or test files were touched.

## Code quality & style

Read all 35 added explanations. They consistently do the two things asked of them — restate the rule in plain language, then explain *why* it matters (interoperability, caching/freshness, loop prevention, routing correctness) — rather than parroting `.summary()`. Prose is grammatically clean throughout.

- **Length**: computed word counts for all 35 — min 67, max 92, average ~78. All but one land inside the ~40-90 target; `sender-must-not-generate-trailer-unless-permitted.rule.ts` is 92 words, trivially over. Non-blocking.
- **Voice**: the majority are third-person ("a proxy must...", "an intermediary should..."). Roughly 5-6 of the 35 (~15%) lean second-person — `sender-should-generate-trailer-header-when-sending-trailers`, `client-must-not-use-special-request-target-forms-with-other-methods`, `sender-must-list-connection-specific-field-in-connection-header`, `sender-must-not-send-end-to-end-fields-as-connection-options`, and a "you" mid-clause in `intermediary-must-check-and-update-max-forwards`. That's a lower share than the effort-wide ~48/171 (~28%), so it doesn't read as inconsistent across this PR as a whole — but two of them (`client-must-not-use-special-request-target-forms...`, `sender-must-not-send-end-to-end-fields-as-connection-options`) switch from third- to second-person mid-paragraph, which reads slightly awkward within a single explanation. Given this is tracked, acknowledged follow-up work effort-wide, treating as a style note, not a blocker.

## Specific suggestions

- `origin-server-may-generate-date-for-1xx-5xx.rule.ts`: asserts the MAY-vs-MUST split exists "because these response classes are less useful for caching than 2xx/3xx/4xx" — RFC 9110 §6.6.1 doesn't state this rationale explicitly (plausible and true, but inferred rather than textual). Consider a version that doesn't imply the RFC itself gives that reason.
- `intermediary-should-remove-known-hop-by-hop-fields.rule.ts`: "...leading to connection errors or request smuggling risks" — accurate and well-known in practice, but "request smuggling" is security/industry terminology that §7.6.1 itself doesn't use. Fine to keep as real-world context; flagging only that it's supported by broader HTTP security knowledge rather than the cited section text.
- Both are minor rationale-scope nits, not misstatements — nothing here contradicts the RFC.

## Issues & risks

**Correctness (main risk surface for this PR)**
Went beyond the requested 10-15 sample and RFC-checked all 35 explanations against the local `rfc9110.txt`, covering every section this PR touches: §5.1 (Field Names), §5.3 (Field Order), §5.6.1 (Lists), §5.6.7 (Date/Time Formats), §6.5.1/6.5.2 (Trailer Fields), §6.6.1 (Date), §6.6.2 (Trailer), §7.1 (Determining the Target Resource), §7.2 (Host and :authority), §7.6/7.6.1/7.6.2/7.6.3 (Message Forwarding, Connection, Max-Forwards, Via), §7.7 (Message Transformations), §7.8 (Upgrade). Representative spot-checks:

- `sender-must-not-generate-trailer-unless-permitted.rule.ts` (§6.5.1) — the field list ("framing, routing, authentication, request modifiers, response controls, or content format") is a direct, accurate paraphrase of the RFC's own list. Match confirmed.
- `sender-must-not-send-end-to-end-fields-as-connection-options.rule.ts` (§7.6.1) — the Cache-Control example is lifted directly from the RFC's own example ("Cache-Control is never appropriate as a connection option"). Match confirmed.
- `proxy-must-not-transform-content-with-no-transform-directive.rule.ts` (§7.7) — "medical images and scientific results" / integrity-checks-and-signatures rationale matches the RFC's stated concern almost verbatim. Match confirmed.
- `user-agent-must-generate-host-or-authority-header.rule.ts` (§7.2) — "server that hosts many sites on one address" accurately restates "distinguish among resources while servicing requests for multiple host names." Match confirmed.
- `intermediary-must-not-forward-message-to-itself.rule.ts` (§7.6) — this explanation makes an implementation-specific claim ("The Via chain is used here to detect a hop the message has already passed through"), so I checked it against the rule's actual `.rule()` body, not just the RFC: the code does exactly that (flags a duplicate `received-by` entry in the parsed Via chain). Match confirmed.
- `sender-must-not-generate-empty-list-elements.rule.ts` (§5.6.1) — "such as Vary or Allow" checked against the rule's actual `listTypedResponseHeaders` array (`vary`, `allow`, `content-encoding`, `accept-ranges`); both named examples are in fact tested by this rule. Match confirmed.
- Pre-existing issue, already resolved: Copilot's first review pass (on commit `fb7c426`) correctly flagged that `proxy-must-forward-unrecognized-header-fields.rule.ts` originally weakened the RFC's MUST to "it **should** pass along header fields...". I diffed the file between that commit and the current head (`1d78012a`) via the GitHub contents API and confirmed it now reads "it **must** pass along...", matching the rule's `error` severity and its own `.description`/`.summary`. Copilot's second pass (on the current head) found no new comments, consistent with this being fixed. Nothing further to flag here.

No misstated normative requirements or fabricated/unsupported rationale found anywhere across the 35 explanations.

**Convention-following**
Confirmed via the diff: 105 additions / 0 deletions, exactly 3 lines added per file, inserted consistently between `.summary(...)` and the next existing call. No other builder calls, descriptions, summaries, URLs, severities, types, or rule logic changed. Also fetched all 35 files at the PR head and confirmed every one has a `.type(...)` combination that includes `analytics`, `static`, and/or `test` — none is purely `informational`. Fully compliant with the stated scope.

**Test coverage**
N/A — metadata-only change (a documentation string read by nothing at runtime except display), so no test changes are expected. The PR description states the full 171-rule superset was verified green (tsc/eslint/prettier/105 tests) before splitting; this PR is a strict additive subset of that.

**Security**
No `.rule()`/`.overrideAnalyticsRule()` validation-logic function bodies were touched anywhere in the diff — confirmed by inspecting every hunk (all are pure insertions of an `.explanation()` block, 0 deletions total). No scope creep found.

## Decision

Approving. All 35 explanations are accurately grounded in the RFC sections their rules cite (cross-checked against the primary text, not just style-reviewed, plus two implementation-level spot-checks), the one substantive issue Copilot flagged was already fixed before this review, and the change is cleanly additive with no informational-only rules or off-scope edits. The two "specific suggestions" above are optional tightening, not blockers.

PR #348, opened 2026-08-03
feat(core): recommended rule-configuration profiles on RuleSet (#467)

matthyk:
## Summary

Adds a first-class mechanism for rule packages to ship curated, adoptable rule configurations, and ships the RFC 9110 noise-tuned set as the first instance. Implements the decision recorded in **ADR-0016** (see `thymianofficial/thymian-internal#467`).

A package's exported `RuleSet` gains an optional **`profiles`** field — named override maps (`recommended` / `strict` / `minimal`) that reuse the exact same per-rule shape as the config `rules:` block. Users select one per rule set:

```yaml
ruleSets:
  - '@thymian/rules-rfc-9110'                       # bare string → recommended (the default)
  - name: '@thymian/rules-api-description-validation'
    profile: strict                                 # object form → explicit profile
```

A bare string (or an object with no `profile`) resolves to **`recommended`**. There is no opt-out flag — to run spec-faithful, set `profile: strict`.

## How it works

**Resolution order per rule:** shipped default → selected **profile** override → user `rules:{}` (**user always wins**). The `ruleSeverity` floor filter still runs afterwards on the fully-resolved severity — it was not moved into the merge. A profile only ever touches rules belonging to the rule set that declares it, so composition across packages is safe by construction; a missing/unknown profile resolves to an empty override (no throw).

**Additive threading:** the profile selection travels into the workflow input as a parallel `ruleProfiles` map and is applied inside `loadRules`, so the internal `core.lint`/`core.test`/`core.analyze` detail-action `rules` payload is unchanged (ADR-0007 contract preserved).

## Changes

- **`@thymian/core`** — `profiles?: Record<string, RulesConfiguration>` on `RuleSet`; loader applies the selected profile before user config (`applyProfileThenConfig`, `resolveProfileConfig`); optional `ruleProfiles` on the workflow inputs + action schemas.
- **`@thymian/common-cli`** — `ruleSets` entries accept `string | { name, profile? }` (`RuleSetEntry`); JSON-Schema `oneOf` in `thymian-config-schema.ts` and its hand-maintained `.json` twin, kept in lockstep; `mergeRuleSets` normalizes to `{ name, profile }`.
- **`thymian`** CLI — `lint`/`test`/`analyze`/`rules list` split the merged rule sets into `rules` + `ruleProfiles`; `generate config` documents that scaffolded entries use `recommended`.
- **`@thymian/rules-rfc-9110`** — `recommended` = the 16-change noise tuning (12 rules `off`, 2 narrowed from `static` to `analytics`/`test`, 2 demoted to `hint`); `strict = {}`; `minimal` reuses `recommended` as a documented stopgap. A package test guards that every profiled rule id resolves to a real rule (the seed analysis had a `-sever-` typo, corrected here).
- **Docs** — `ADR-0016` + architectural-decisions index row; config schema reference regenerated.

## Testing

`nx run-many -t typecheck lint test -p core common-cli thymian rules-rfc-9110` — all green (core 296, common-cli 244, thymian 56, rules-rfc-9110 109). New tests cover profile-before-user resolution, user-wins, unknown-profile/id no-ops, `type` narrowing, floor-filter interaction, schema validation of the object form, and the RFC 9110 drift/typo guard.

## Follow-ups (deferred, non-blocking)

Tracked in the planning repo: `thymianofficial/thymian-internal#472` (diagnostic for explicitly-selected-but-absent profile names + action-schema enum alignment), `#473` (pre-existing `skipOrigins` no-op), `#474` (silent drop of duplicate ruleSet with conflicting profile), `#475` (profile-sourced override error misattribution). A curated `minimal` core (currently aliasing `recommended`) is also a known follow-up.

Refs `thymianofficial/thymian-internal#467`, `thymianofficial/thymian-internal#324`.


BaggersIO (approved):
I read this in full — pulled complete file context for `rule-loader.ts`, `thymian.ts`, both config schemas, and the rules-rfc-9110 rule corpus — rather than trusting the description at face value.

**Correctness**: `applyProfileThenConfig` (rule-loader.ts:195) chains `applyRuleConfiguration` twice, correctly giving default → profile → user-wins. The severity floor filter runs *after* this resolution (confirmed at the `ruleFilter(rule)` call sites in `loadRuleSet`/`loadRules`), not merged into it, matching the ADR. Unknown profile name and a missing `profiles` map both correctly no-op via `resolveProfileConfig` (:184).

**Backward compat**: verified in `thymian.ts` — the `core.lint`/`test`/`analyze` payloads (~424/452/495) still only carry `{format, rules: Rule[], rulesConfig, options, ...}`. `ruleProfiles` is consumed solely inside `loadRules` before that payload is built; old callers omitting it correctly default to `recommended` per rule set.

**Schema lockstep**: the new `ruleSets` `oneOf` block is byte-identical between the `.ts`/`.json` twins. Side finding: `validateConfig()`, the function actually gating CLI usage, only ever compiles the `.json` file — the `.ts` export isn't consumed by any runtime or doc-gen path, only by this PR's own new integration test — so nothing guards future drift between the two going forward.

**"-sever-" typo**: cross-referenced all 16 `recommended` ids against the real rule corpus; all resolve correctly. Unrelated bonus finding: `origin-sever-may-accept-connect-request.rule.ts` carries the identical typo in a different, untouched rule — worth its own follow-up issue, doesn't block this PR.

**Security**: `resolveProfileConfig`'s `profiles?.[profileName]` isn't own-property-guarded (`__proto__` resolves to `Object.prototype` rather than `{}`), but the only consumer does narrow keyed reads (never spread/merge) for real, namespaced rule ids — inert in practice. Worth an `Object.hasOwn` guard as cheap insurance, not a blocker.

The 4 deferred follow-ups (thymian-internal#472-475) all check out as genuinely non-blocking.

Two gaps worth a follow-up test, not worth blocking on: no test loads two rule sets with different profiles together to pin down the cross-package isolation claim, and no CLI-level test threads a config's `profile:` selection through to `Thymian.lint()` (the mock harness in `config-resolution.integration.test.ts` would make this cheap).

Approving.

PR #349, opened 2026-08-03
feat(common-cli): wrap CLI output to terminal width

matthyk:
## What & why

Custom-rendered CLI output (reports, `validate`, `rules list`, `request`, etc.) was emitted without any width management, so long lines ran past the terminal edge and wrapped inconsistently with oclif's own help/error output. This PR makes our custom render layer wrap to the terminal width the same way oclif does, then hardens that wrapping so it never corrupts machine-readable or copy-pasted output.

Closes #471.

## How it works

A small render layer in `packages/common-cli/src/render/utils.ts`:

- **`terminalWidth()`** mirrors oclif's own resolution order — explicit `OCLIF_COLUMNS`, then `settings.columns`, then the clamped real `process.stdout` width — computed per call so tests and live terminals both work.
- **`wrap()`** wraps prose to that width, ANSI-aware (colour codes survive, wrapping is measured by visible width).
- **`wrapIndented()`** wraps leaf prose with a *hanging* indent: the caller supplies the first line's prefix (indent + optional tree glyph, e.g. `"  ├── "`) and continuation lines align under the content without repeating the glyph.

These are wired through the report renderer, `validate`, `rules list`, `request`, `plugins list`, `feedback`, `generate`, and `serve`.

## Safety hardening (from code review)

Wrapping user-facing output must never make it wrong. The following make wrapping copy-paste- and pipe-safe:

- **Soft wrapping (`hard: false`)** — unbreakable tokens (file paths, URLs, JSON blobs, rule identifiers) are never split mid-token, so they stay copy-paste- and `grep`-able. Only whitespace-separated prose reflows.
- **Non-TTY output is emitted verbatim** — `terminalWidth()` reports no finite width when stdout is not a TTY (and no width is pinned), so `thymian … > out.txt` and `… | grep` receive unwrapped output. Machine consumption of piped/redirected output is preserved.
- **Bogus width guarded** — only a *positive* `OCLIF_COLUMNS`/`settings.columns` is honoured, so `-1`/`0`/non-numeric no longer collapses every line to one column per line.
- **Raw HTTP wire values are never wrapped** — in `request.ts` the status/request line, `Host`, and header values are printed verbatim so values containing significant whitespace are not mutated.
- **Width-aware run-heading divider** — the `tool · type · ────` separator in the report is sized from `terminalWidth()` (80 fallback) instead of a hardcoded 70, so it wraps uniformly with everything else.
- **No dangling glyph lines** — `wrapIndented('')` renders nothing instead of a lone tree/bullet glyph.
- **Dead indent removed** — `rules list` `formatRule` no longer builds a leading indent that the caller stripped and re-added; `wrapIndented` owns the indent.

## Testing

- `common-cli`, `thymian`, and `core` unit suites pass.
- `terminal-width.test.ts` covers soft-wrap keeping tokens intact, non-TTY verbatim output, non-positive `OCLIF_COLUMNS`, indent reservation, and the empty-text guard.
- `cli-report-renderer.test.ts` asserts every line (prose and the now width-aware divider) stays within the pinned width.
- Lint + build clean for the touched packages.

### DIFF
--- a/e2e-tests/src/env-utils.ts
+++ b/e2e-tests/src/env-utils.ts
@@ -17,6 +17,12 @@ export function getCleanEnv(): Record<string, string> {
   }
   env['FORCE_COLOR'] = '0';
 
+  // Pin the wrap width so custom-rendered CLI output wraps at a deterministic
+  // 80 columns regardless of the test runner's TTY (mirrors oclif's own
+  // `stdtermwidth` source). Without this, `terminalWidth()` would read the
+  // runner's real/absent TTY width and vary between environments.
+  env['OCLIF_COLUMNS'] = '80';
+
   // Force production mode so that oclif's ts-path.js does NOT attempt to
   // remap the compiled `dist/commands` directory back to the `src/commands`
   // source path.  The published npm package only ships `dist/`, so the
--- a/package-lock.json
+++ b/package-lock.json
@@ -31791,6 +31791,7 @@
         "@inquirer/prompts": "^7.6.0",
         "@oclif/core": "^4.3.3",
         "@thymian/core": "0.0.0-PLACEHOLDER",
+        "string-width": "^4.2.3",
         "tslib": "^2.3.0",
         "wrap-ansi": "^7.0.0",
         "yaml": "^2.8.3"
--- a/package-lock.json
+++ b/package-lock.json
@@ -31792,6 +31792,7 @@
         "@oclif/core": "^4.3.3",
         "@thymian/core": "0.0.0-PLACEHOLDER",
         "tslib": "^2.3.0",
+        "wrap-ansi": "^7.0.0",
         "yaml": "^2.8.3"
       }
     },
--- a/packages/common-cli/package.json
+++ b/packages/common-cli/package.json
@@ -46,6 +46,7 @@
     "@inquirer/prompts": "^7.6.0",
     "tslib": "^2.3.0",
     "yaml": "^2.8.3",
+    "wrap-ansi": "^7.0.0",
     "@oclif/core": "^4.3.3",
     "@thymian/core": "0.0.0-PLACEHOLDER"
   }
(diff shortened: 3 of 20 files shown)


BaggersIO (approved):
Read through `render/utils.ts` and every call site, and checked the edge cases this touches against the exact pinned `wrap-ansi@7.0.0`/`string-width@4.2.3` versions (ANSI survival across wrap points, non-SGR escapes, tokens far longer than the wrap width). All of that holds: color codes survive wrapping correctly, and the non-TTY / positive-only-width guards are correctly implemented and tested (including decimals, negatives, and `"80cols"`).

Two things worth addressing:

1. `create-execution-renderer.ts:245` and `:302` — the rule-id reference line (`'› ' + rule.id` / `'› ' + execution.ruleId`) never goes through `wrap`/`wrapIndented`, unlike every sibling line in the same functions (I confirmed this directly — the status line, heading, and label/message lines right next to it are all converted, this one isn't). Real rule ids in `rules-rfc-9110` run to ~85 chars (e.g. `rfc9110/final-recipient-should-exclude-sensitive-request-data-from-response-to-trace`), so this line still overflows a standard 80-col terminal in the default report view — the exact class of bug this PR fixes everywhere else. Not a regression (same overflow existed before this PR), but it's in-scope and easy to miss because no wrap-test fixture attaches a `ruleId` long enough to trip it. Worth a fast follow-up given it's the primary output path.
2. `positiveInt()` (`utils.ts:41`) uses strict `Number(...)`, but the actual installed `@oclif/core@4.13.0` parses `OCLIF_COLUMNS` via lenient `Number.parseInt`. So `OCLIF_COLUMNS=80cols` wraps oclif's own `--help` at 80 but leaves this custom output unwrapped — a real exception to the "consistent with oclif" framing, probably the right call (oclif's own handling of e.g. `-1` is arguably worse) but worth a docstring caveat.

Minor: `findings.ts`'s `expected`/`actual` JSON blobs do get split at whitespace inside string values, despite the "JSON blobs never split mid-token" framing in the description — the soft-wrap guarantee only covers whitespace-free tokens, so a JSON value with a prose string inside it can still wrap across lines.

None of this blocks merge — core module is correct and well-tested against the hardest edge cases. Approving.

PR #35, opened 2025-09-28
ci(repo): temporarily disable build matrix

Markus-Ende:

PR #350, opened 2026-08-03
feat(core): human-readable plugin & config validation errors

matthyk:
## Summary

Closes [#35](https://github.com/thymianofficial/thymian-internal/issues/35).

Ajv's raw validation output (`must have required property 'x'`, `must be object`, `must NOT have additional properties`) is hard for users to act on. This PR introduces a shared `formatAjvErrors` helper in `@thymian/core` that wraps raw Ajv errors with [`@segment/ajv-human-errors`](https://github.com/segmentio/ajv-human-errors), and routes every validation consumer through it so error messages are consistent and readable across the codebase.

## Changes

- **`@thymian/core` — `ajv.ts`**
  - Configure the shared Ajv instance with `allErrors: true` + `verbose: true` (both required by `@segment/ajv-human-errors` to render messages and to list every failing constraint).
  - Add `formatAjvErrors(errors)` returning `{ message, details }`:
    - `details` is a bounded, per-error list (capped at 20 to keep output readable on deeply-invalid / adversarial input).
    - Falls back to Ajv's own `errorsText` when the library can't render a given error shape (e.g. `additionalProperties: false` without `properties`, `patternProperties`).
    - Returns an `"Unknown validation error"` message for an empty/null error list — never an empty message.
- **`@thymian/core` — `thymian.ts`**
  - Emit human-readable detail for invalid plugin options.
  - Migrate `#assertValidWorkflowInput` from raw `errorsText` to `formatAjvErrors` so WS action-input errors (#300 client) use the same friendly phrasing.
- **`@thymian/common-cli` — `validate-config.ts`**
  - Render one bullet per error, and fall back to the summary `message` when no per-error `details` are available — restoring the previous non-empty-message guarantee (the old `?? 'Unknown error'` behaviour).
- **`@thymian/plugin-har` — `har-schema.ts`**
  - Surface human-readable HAR validation errors via the shared helper instead of raw `errorsText`.

## Tests

- New `packages/core/test/ajv.test.ts` covering `formatAjvErrors` (rendering, fallback, empty-list, and cap behaviour).
- Updated assertions in `core`, `common-cli`, and `plugin-har` test suites to match the new human-readable phrasing.

All affected suites pass; `prettier --check` and core `tsc` are clean.

## Review notes

This PR incorporated findings from a high-effort code review:
- Fixed a blank-message regression in `validate-config` when the Ajv error list is empty.
- Made the workflow-input and HAR validation paths reuse the shared helper for consistency.
- Kept `allErrors` and `verbose` on the shared Ajv instance intentionally (multi-error listing + library requirement).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8374,6 +8374,15 @@
         "node": ">=20"
       }
     },
+    "node_modules/@segment/ajv-human-errors": {
+      "version": "2.16.0",
+      "resolved": "https://registry.npmjs.org/@segment/ajv-human-errors/-/ajv-human-errors-2.16.0.tgz",
+      "integrity": "sha512-cHNfZcbHrmuYOA7/Sn7HlIDHanamiRTZtngfxcAuFaKQjP7cSqsVHjLz38FI2FQ8JDLz3syGLaz10Gn2ddo7+w==",
+      "license": "MIT",
+      "peerDependencies": {
+        "ajv": "^8.0.0"
+      }
+    },
     "node_modules/@shikijs/core": {
       "version": "4.3.1",
       "resolved": "https://registry.npmjs.org/@shikijs/core/-/core-4.3.1.tgz",
@@ -31801,6 +31810,7 @@
       "license": "AGPL-3.0-only",
       "dependencies": {
         "@fastify/deepmerge": "^3.1.0",
+        "@segment/ajv-human-errors": "^2.16.0",
         "ajv": "^8.18.0",
         "ajv-formats": "^3.0.1",
         "chalk": "^5.4.1",
--- a/packages/common-cli/src/validate-config.ts
+++ b/packages/common-cli/src/validate-config.ts
@@ -1,6 +1,6 @@
 import { EOL } from 'node:os';
 
-import { ajv } from '@thymian/core';
+import { ajv, formatAjvErrors } from '@thymian/core';
 
 import thymianSchema from './thymian-config-schema.json' with { type: 'json' };
 
@@ -21,11 +21,14 @@ export function validateConfig(config: unknown): ConfigValidationResult {
   if (valid) {
     return { valid };
   } else {
+    const { message, details } = formatAjvErrors(validationFn.errors);
+
     return {
       valid,
       message:
-        validationFn.errors?.map((e) => EOL + '   * ' + e.message).join('') ??
-        'Unknown error',
+        details.length > 0
+          ? details.map((detail) => EOL + '   * ' + detail).join('')
+          : message,
     };
   }
 }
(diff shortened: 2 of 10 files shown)


BaggersIO (approved):
Read through the diff plus the actual `@segment/ajv-human-errors@2.16.0` source (pulled the real tarball from the registry, since the linked `segmentio/ajv-human-errors` repo 404s — the real source lives in `segmentio/action-destinations`) to check the secret-leakage question directly rather than assume it, given `plugin-har` validates HAR files that routinely contain auth headers and cookies.

**Verdict: no leak.** `getMessage()` in `formatting.ts` only ever reduces `error.data` to `humanizeTypeOf(data)` (typeof label), `data.length`, or `Object.keys(data).length` — the raw value never reaches a message string. `additionalProperties`/`required`/`propertyNames` echo schema property *names*, never values. `formatAjvErrors` (`ajv.ts:42`) constructs `new AggregateAjvError(errors)` with no options, so the library's `includeData`/`includeOriginalError` both default to `false`, and even then `formatAjvErrors` only ever reads `.message` off each error, never `.data`. Good bar cleared.

One real bug worth fixing: `thymian.ts:224` produces a double period. `AggregateAjvError`'s message is already period-terminated (`aggregate-ajv-error.ts:27`: `` this.message = `${capitalize(message)}.` ``), and the template at :224 appends another `.` on top — I confirmed this directly against the library source. E.g. "...but it was undefined.." The sibling call at `thymian.ts:232` (`register()`) doesn't have this. A little ironic for a "human-readable errors" PR, and no test catches it since existing assertions are all substring/regex, never end-of-string.

A few smaller things, none blocking:
- The `allErrors`/`verbose` config change is on the *shared* core Ajv instance, but at least two consumers never benefit from it: `plugin-websocket-proxy`'s `isClientToServerMessage` (used on every inbound WS message) and `rule-loader.ts:143`'s rule-options validation both only consume a boolean and never read `.errors`, so they now pay the collect-everything/retain-data cost for nothing.
- `rule-loader.ts:143`'s `InvalidRuleOptionError` wasn't migrated to `formatAjvErrors` despite being the same shape of problem as the plugin-options case migrated in the same file — the description's "every validation consumer" is slightly overstated.
- `ajv.ts:45`'s `.filter(Boolean)` after iterating `AggregateAjvError` looks dead — the library already drops null-message ("redundant") errors internally before yielding them.
- No test pins "the rendered message never contains the raw invalid value" — true today (verified above) but resting on an unpinned (`^2.16.0`) caret range against a library with an opt-in `includeData` flag. Worth a one-line regression test given the codebase already has precedent for this exact concern elsewhere (`schema-error.ts`'s `isPrimitive(error.data)` guard).

The cap/truncation logic, empty-list guarantee, and `validate-config.ts` fallback fix are all correct. Approving; the double-period fix would be a trivial, worthwhile fast-follow.

Issue #347, opened 2026-08-03
Home page: widespread WCAG AA color-contrast failures in dark mode


BaggersIO:
## Summary

The home page (`/`) fails axe's WCAG 2.1 AA color-contrast check in dark mode with **66 affected nodes**, spanning at least the `quick-start` and `rules-drift-prevention` components (muted/small text using gray tokens like `#898c8f` on `#25272a`, `#56585b` on `#171819` - ratios as low as 2.49:1 against a 4.5:1 requirement).

## How this was found

While building Story 13.3 (navigation e2e viewport matrix, thymian-internal#470), the new matrix suite scans every route in light AND dark mode. The existing `e2e/accessibility.spec.ts` only ever scanned Home in **light** mode - this is the first time Home has been axe-scanned in dark mode, and it surfaced a genuine, pre-existing gap unrelated to the navigation work.

## Scope note

This is unrelated to Epic 13 (site navigation/chrome) and was not introduced by it. Fixing it requires actual design rework of muted-text color tokens across multiple unrelated marketing components, which is out of scope for the navigation epic. Filed here for separate triage rather than silently skipped or bundled into an unrelated PR.

## Repro

```
npx astro build
npx playwright test --project=mobile-chromium -g "no violations, dark"
```
(from `astro-docs/`, on the `story/468-13-468-navigation-data-model-three-band-desktop-header` branch / thymian-internal#468 PR - the matrix suite will be upstreamed to main once that PR merges)

BaggersIO:
## Additional finding (Story 13.3 V2 = 810px band)

Testing the 800-1152px viewport band surfaced two more pre-existing, unrelated issues, same root cause as the original report (widths/themes never axe-scanned before):

1. **Home, light, `.lifecycle-scroll-fade`** (`scrollable-region-focusable`): same class of issue as the original report, same component (`lifecycle-grid`) - fixed directly (added `tabindex="0"`, zero visual change) since it's the identical trivial remediation already applied to the other 4 scrollable elements in this PR.
2. **Enterprise page, light AND dark, `color-contrast`** (9 nodes): `.ep-incident-card--red`, `.ep-severity--red`, and related "incident card" elements - red-on-white/red-on-dark text as low as 2.44:1 against a 4.5:1 requirement. This is a genuine color/design fix (the red severity palette needs actual rework), not a one-line accessibility-attribute fix like the others - left unfixed and skipped in the nav epic's test suite pending separate triage.

[...]

PR #353, opened 2026-08-04
feat(repo): homepage promo strip (Epic 10)

BaggersIO:
## Summary

Ships the full homepage promo strip as one PR (Epic 10, Stories 10.1-10.3 - thymian-internal#444, #445, #446):

**10.1 - Build the PromoStrip component (upcoming + latest)**
- New `src/components/promo-strip/promoStripMeta.ts`: a pure, `buildDate`-injected selection module (`selectPromoItems`) composing up to 3 nearest Upcoming Events + the single latest Resource. `PROMO_EVENT_LIMIT` is a cap, never a fill target - fewer Upcoming Events means a shorter strip, never topped up from Past.
- New `src/components/promo-strip/PromoStrip.astro`: renders the selection as a compact card grid, self-seeding its own `--ev-*` design-token subset (light + dark) since the homepage has no `.events-root`/`.resources-root` ancestor to inherit from. Deliberately does not reuse `EventCard`/`ResourceCard` (wrong altitude, tokens wouldn't resolve).
- Not wired into any page yet by design - this story ships the component only.

**10.2 - Never-empty deterministic fallback**
- Completes FR-15's full precedence in `selectPromoItems`, purely additive to 10.1's contract: 0 Upcoming falls back to the most-recent Past events; 0 Events but >=1 Resource still shows the latest Resource (not the evergreen case); only 0 Events AND 0 Resources shows a static evergreen card.
- The Past arm is an `else`, never a top-up - tenses are never mixed in one render.
- Still unreachable via `astro build` with today's live content (2 Upcoming events exist) - unit-test-only coverage by design, same as 10.1.

**10.3 - Integrate the strip into the homepage**
- The only product-code page change: slots `<PromoStrip />` into `src/content/docs/index.mdx` directly after the `QuickStart` block.
- Adds the rendered/visual/a11y/responsive verification 10.1/10.2 deliberately deferred: link/discovery e2e assertions (every card resolves to `/events/` or `/resources/`, both 200, surfaced titles present on the target index - SM-2, <=2 clicks), a `.promo-strip`-scoped axe scan in light + dark + at a 390x844 mobile viewport, and full nav e2e matrix regression coverage (no new horizontal-overflow failures).
- A review pass on this story additionally wrapped the strip in its own `<section aria-labelledby>` with a heading, demoting card titles to `<h3>` so they nest correctly now that the component is actually mounted (previously bare `<h2>`s with no bucket heading - flagged as a to-decide item in 10.1's own Dev Notes).

## Test plan
- [x] `npx nx run astro:test` - 177/177 passing (157 baseline + 20 new across the three stories)
- [x] `npx nx run astro:astro -- check` - 14 errors, identical to the pre-existing baseline (0 new)
- [x] `npx nx build astro` - succeeds, 122 pages, `dist/index.html` contains the strip, link-validator clean
- [x] `npx nx run astro:e2e` - 156 total: 151 passed, 5 skipped (all pre-existing `test.fixme` cells tracked against `thymianofficial/thymian#347`, unrelated dark-mode contrast bug, untouched), 0 failed - re-verified independently multiple times with zero flakiness

Screenshot 2026-08-10 at 08 43 25

### DIFF
--- a/astro-docs/e2e/promo-strip.spec.ts
+++ b/astro-docs/e2e/promo-strip.spec.ts
@@ -18,10 +18,32 @@ import { forceDarkTheme } from './navigation/nav-shared';
 const PROMO_EVENT_LIMIT = 3;
 
 /**
- * Story 10.3 (Epic 10, thymian-internal#446): `PromoStrip.astro` (10.1/10.2)
- * renders on the homepage for the first time here, so its rendered/visual/
- * a11y/responsive verification lands in this story - 10.1/10.2 deliberately
- * shipped unit-test-only coverage.
+ * Both pages that mount `PromoStrip`: Home (`index.mdx`) renders it as the
+ * last content section, directly above the Get Started/Enterprise CTA row;
+ * Enterprise (`enterprise.mdx`) mounts it as the page's last element
+ * entirely. Every discovery + scoped-a11y check below runs once per page so
+ * neither mount point is ever silently uncovered. If a future change adds or
+ * removes a mount point, update this list by hand to match (mirrors how
+ * `PROMO_EVENT_LIMIT` above is kept in sync by hand).
+ */
+const PROMO_STRIP_PAGES = [
+  { path: '/', label: 'home' },
+  { path: '/enterprise/', label: 'enterprise' },
+];
+
+if (PROMO_STRIP_PAGES.length === 0) {
+  throw new Error(
+    'PROMO_STRIP_PAGES is empty - this would silently register zero promo-strip tests instead of failing loudly.',
+  );
+}
+
+/**
+ * Originally Story 10.3 (Epic 10, thymian-internal#446): `PromoStrip.astro`
+ * (10.1/10.2) rendered on the homepage for the first time here, so its
+ * rendered/visual/a11y/responsive verification landed in this file -
+ * 10.1/10.2 deliberately shipped unit-test-only coverage. It later also
+ * gained an Enterprise mount point (see `PROMO_STRIP_PAGES` above); the scope
+ * below applies identically to both pages.
  *
  * Scope:
  *  - Discovery (SM-2, <=2 clicks): every card link resolves to a real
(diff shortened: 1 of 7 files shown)


matthyk (commented):
## Review — homepage promo strip (Epic 10)

Very well-documented and defensive PR. **No correctness bugs survived verification** — the module logic, helper contracts, and tests all check out (no shared-array mutation; `latestResource` correctly independent of the event branch; stretched-link/attribution `z-index` reasoning holds; all accessed fields are schema-required or refine-guaranteed). The findings below are all maintainability/cleanup, dominated by CSS duplication. None are blockers.

### 1. `--ev-*` token palette is now a third independent copy — `PromoStrip.astro`
The WCAG-tuned light+dark teal contract is defined in full in `EventsLayout.astro` (`.events-root`) and `ResourcesLayout.astro` (`.resources-root`), and this adds a third subset copy. Scoping the tokens to `.promo-strip` is justified (the homepage has no `.events-root`/`.resources-root` ancestor) — but *copying the hex values* rather than applying a shared custom-property sheet to that class is a drift risk: when a contrast value is re-tuned (the dark `--ev-accent-text` was already re-derived to `#74cfbf` once), an editor updates the two layouts that visibly sit together and silently misses the homepage strip, which can drift sub-AA at margins the scoped axe scan won't catch.

### 2. Badge markup re-implements the section-agnostic `.badge` — `PromoStrip.astro`
`.promo-card-badge`/`-alt` duplicate `.badge`/`.badge--type-alt` (`EventCard.astro`) byte-for-byte (padding/radius/weight/letter-spacing/text-transform). Not reusing the full-width `EventCard` is defensible (wrong altitude), but the badge itself is layout-independent — a shared badge class/partial would avoid a third divergent copy on the very page advertising those sections.

### 3. `PROMO_EVENT_LIMIT` is hand-mirrored in the e2e spec — `e2e/promo-strip.spec.ts:24`
The comment correctly explains the source module can't be imported (transitive `astro:content` virtual module), but the constant itself is a dependency-free integer. Bumping the source limit silently staleness the `count <= PROMO_EVENT_LIMIT + 1` guard (false failure or masked regression). Extracting the bare constant into a tiny `astro:content`-free module importable by both removes the manual-sync seam.

### 4. Full events array classified twice on the fallback path — `promoStripMeta.ts`
`allEvents.filter(classify … 'upcoming')`, then on 0 upcoming `allEvents.filter(classify … 'past')` re-runs `classify` over the whole set and duplicates the predicate. Minor (build-time, small N); a single partition pass would be one `classify` per event. Lowest priority.

---
*Checked and cleared: shared-array mutation (filtered/spread copies), the AD-6 dangling-ref throw's "widened blast radius" (build already fails at `/resources` via `assertOriginEventsResolve`), `place`/`location` undefined (schema-refine-guarded), and the hardcoded `#017b64` hover shadow (identical to `EventCard.astro:134` — pre-existing idiom, not introduced here).*

matthyk (approved):

PR #356, opened 2026-08-04
feat(repo): raise Node support floor to 22+

Dariant31:
Closes thymianofficial/thymian-internal#406

Enforces a Node.js **≥ 22** support floor across the published packages, CI and the docs, so the
documented runtime promise matches what we actually build, test and support.

Approved via the Sprint Change Proposal 2026-07-20 (correct-course; drift analysis in
thymianofficial/thymian-internal#404 §3 row 5). Before this change the sources disagreed:

| Source | Said |
| --- | --- |
| PRD | Node 20+ |
| Architecture | pinned 22.x |
| Astro 6 (astro-docs) | requires ≥ 22.12 |
| CI matrix | tested 20, 22, 24 |
| Published packages | **no `engines` field at all** |

## Changes

**`engines` on every published package** — `"engines": { "node": ">=22" }` added to all 16
packages under `packages/*` (`thymian` + all `@thymian/*`), which is exactly the set `nx release`
publishes (`release.projects: ["packages/*"]`).

Deliberately *not* added to the root `@thymian/source` (`private: true`, never published), nor to
`astro-docs` / `e2e-tests` (workspace projects, not published artifacts).

> The count was 15 for most of this PR's life. `@thymian/plugin-spectral` landed on `main` after
> this branch covered the original 15, and the merge from `main` brought in a 16th published
> manifest without an `engines` field. Caught in review and fixed in `fc57c38c`; it is not
> `private`, is `publishConfig.access: "public"` and matches `release.projects`, so it publishes on
> the next `nx release` like the rest.

**`package-lock.json` synced** (`7a9a6496`) — npm mirrors `engines` into the lockfile's workspace
entries, and the lock had never been regenerated after the manifests changed, so all 16
`packages/*` entries were missing it. `npm ci` does not fail on this (it reconciles dependencies,
not `engines` metadata), which is why it went unnoticed — but every `npm install` regenerated the
same diff for whoever ran one next. The regenerated diff is engines-only: no dependency version or
`resolved`-URL drift.

**CI matrix** — `version: [20, 22, 24]` → `[22, 24, 26]`. With the Node 20 lane gone, the astro
build exclusion it existed to work around is obsolete, so the Build step collapses back to a plain
`npx nx affected:build`:

```diff
-        # Astro 6 requires Node >=22.12, so skip the astro project on the Node 20 lane.
-        run: npx nx affected:build ${{ matrix.version == 20 && '--exclude astro' || '' }}
+        run: npx nx affected:build
```

The Windows-pin comment just above it referenced "Node 20/22's npm"; narrowed to "Node 22's npm"
since the rationale still holds for 22 but the 20 lane no longer exists. No other CI job
referenced Node 20 — lint, license-check and e2e were already on 22, release on 24.

**Docs** — Node 22+ stated in `README.md` (new Requirements section), `CONTRIBUTING.md` (new
Prerequisites section), `docs/introduction/getting-started.mdx` and
`docs/tutorials/sampler-getting-started.mdx` (Node 20+ → 22+). The runtime notes state the floor
("22 or newer") plus the versions CI exercises — **22**, **24** and **26**.

The "supported _LTS_ versions" framing is deliberately gone from the README and the getting-started
prerequisites: Node 26 does not reach LTS until October 2026, so listing it under that heading would
be inaccurate. Stating the floor plus the tested versions is true of all three regardless of LTS
status.

## Node 26 lane

`>=22` is an open range, so it already permits Node 26. Leaving 26 untested would reintroduce the
exact permit-but-do-not-verify gap this story exists to close, so 26 joins the matrix (9 lanes:
3 versions × 3 OSes).

Verified locally on **26.7.0** before adding the lane:

| Check | Result |
| --- | --- |
| `npm ci` (clean, npm 11.19) | ✅ 2053 packages, no `EBADENGINE` |
| `nx run-many -t build` | ✅ 16 projects, astro included |
| `nx run-many -t lint` | ✅ 17 projects |
| `nx run-many -t test` | ✅ 157 tests |
| `nx e2e e2e-tests` | ✅ 9 files, 55 tests |

The e2e run is the meaningful one — it publishes every package to Verdaccio, does a real global
`npm install` of the CLI and shells out to the installed binary, so 26 is exercised through the
user path rather than only under vitest. Native deps are fine: the ABI moves 127 → 147 between 22
and 26, and `better-sqlite3` loads unchanged because it is N-API with an in-tarball prebuild.

Two things worth knowing, neither blocking:

- `better-sqlite3` is the de-facto ceiling — it is the one dependency that enumerates majors
  (`20.x || 22.x || … || 26.x`) instead of using an open range, so it will `EBADENGINE` on Node 28
  until upstream adds it. Advisory only, since nothing sets `engine-strict`.
- npm ≥ 11 (bundled with Node 24 and 26) no longer runs install scripts by default, so
  `better-sqlite3`'s `prebuild-install || node-gyp rebuild` is skipped with a warning on both the
  workspace and global installs. Nothing breaks today only because the prebuilt binary ships
  inside the tarball.

## Release impact: patch `0.1.14`

Per review, the feature commit is a plain `feat(repo):` with **no** breaking-change marker, so this
ships as a patch and does **not** open a new minor. The matrix change is `ci(repo):`, which is not
a releasable type and adds no bump of its own.

Worth recording why the marker had to be removed in two places rather than one. nx treats the `!`
and the footer as independent triggers — `utils/git.js`:

```js
const isBreaking = parsedMessage.breaking || commit.body.includes('BREAKING CHANGE:');
```

That is an `||`, and the second operand is a plain substring check anywhere in the body. Dropping
only the `!` would have left `isBreaking === true` via the footer and still produced `0.2.0`
(`major` → `minor` at `0.x` per `adjustSpecifierForZeroMajorVersion`). Both signals are gone, so
the specifier resolves to `minor` → adjusted to **`patch`** → `0.1.13` → **`0.1.14`**.

Consequence to be aware of: with no breaking marker, nx will not emit a "Breaking Changes" section,
so the migration note below is **not** auto-propagated into the generated release notes. It lives
in the commit body as a `Migration note:` line and here in this description. If it should appear in
the `0.1.14` release notes, it needs adding to the GitHub release by hand.

> **Migration:** Node 20 is no longer supported; upgrade to Node 22+.

Note also that no `.npmrc` sets `engine-strict`, so `engines` makes npm **warn** on Node < 22
rather than hard-fail the install. A hard gate would be a separate, deliberate decision.

## Verification

Re-run on Node 22.19.0 after merging `main` and adding `plugin-spectral` (project counts are up by
one against the earlier run, which predates that package):

- `nx run-many -t build` — 17 projects ✅
- `nx run-many -t lint` — 18 projects ✅
- `nx run-many -t test` — 17 projects, 1410 passed / 1 todo ✅

Also verified: all **16** published manifests report `engines.node === ">=22"`, and so do all 16
`packages/*` entries in `package-lock.json`; root, `astro-docs` and `e2e-tests` have none.

`nx format:check` no longer passes on this branch, but **not** because of anything here — it flags 7
files (2 `astro-docs` pages, 5 `docs/arc42` pages) that are byte-identical to `origin/main` and
already unformatted there. Confirmed by running prettier against both refs. Pre-existing on `main`
and left alone rather than swept into this diff.

The Node 26.7.0 table above also predates `plugin-spectral`, so it covers 15 packages, not 16. Not
re-run on 26 — the addition here is a manifest field, and the CI matrix exercises 26 on every push.

## Notes for maintainers

**1. `docs/project-context.md` in thymian-internal needs regenerating** after this merges — its
Technology Stack section currently records the floor as "decided but not yet implemented".

**2. AC4 of the story conflicts with this PR as merged.** The story text specifies a *minor* bump
carrying an auto-generated migration note. The review called for no new minor, so this ships as a
patch instead — story #406 should be amended to match, so the next reader is not left reconciling
the two.

**3. The `volta` key in the root `package.json` is now unreferenced.** Removing the Volta paragraph
from `CONTRIBUTING.md` leaves `volta.node: "22.19.0"` as the only mention of Volta anywhere in the
repo, documented nowhere. Left in place because story #406 explicitly scoped it out ("Volta pin
stays as-is"), but if the team does not use Volta it is dead config worth deleting separately.

**4. Two pre-existing issues found while verifying, deliberately left out of this PR:**

- Running a full `nx build` regenerates the checked-in docs under
  `astro-docs/src/content/docs/references/`, and the regenerated output embeds the **build
  machine's absolute path** as a flag default (`--cwd=<value> [default: /Users/…/thymian]`). So
  any full build dirties the tree, and committing that output would leak a local path into the
  published docs. The committed versions on `main` are also stale (they predate the `--cwd` flag).
  I reverted that directory here so it stays out of this diff.
- The CI lint job runs `npx nx affected -t format:check`, but no project defines a `format:check`
  target — it reports "No tasks were run", i.e. the formatting gate is a silent no-op. The
  effective check is the built-in `nx format:check` (which I ran above and which passes).


CONTRIBUTING.md
```diff
@@ -1,5 +1,11 @@
 # Contributing to Thymian
 
+## Prerequisites
+
+Thymian requires **Node.js 22 or newer**. All published packages declare `"engines": { "node": ">=22" }`, and CI
+builds and tests on Node **22** and **24**. The repository pins a development version via [Volta](https://volta.sh/)
```
matthyk: We do not use volta

matthyk (changes requested):
Please do not use an exclamation mark in the commit message. This would produce a new minor version.

Dariant31:
Removed — good catch, that paragraph is gone.

To be clear about how it got in: the root `package.json` has a `volta` key (`"node": "22.19.0"`), and I inferred from its presence that Volta was the team's toolchain manager, then wrote that inference into contributor-facing onboarding docs. A config key existing is not evidence anyone uses it, and I should not have documented a workflow I had not confirmed. The AC only asked for the runtime floor to be stated, so the tooling advice was unrequested scope on my part as well.

`## Prerequisites` is now just the verifiable part:

> Thymian requires **Node.js 22 or newer**. All published packages declare `"engines": { "node": ">=22" }`, and CI builds and tests on Node **22** and **24**.

Worth noting for a separate cleanup: with this paragraph gone, `volta.node` in the root `package.json` is the only Volta reference left in the repo and is documented nowhere. I left it alone because story #406 explicitly scoped it out ("Volta pin stays as-is"), but if the team does not use Volta it looks like dead config.

Dariant31:
Done — force-pushed as `feat(repo): raise Node support floor to 22+`, no `!`.

**One thing worth flagging, because removing the `!` alone would not have had the effect you wanted.** nx treats the `!` and the `BREAKING CHANGE:` footer as two independent triggers — `nx/src/command-line/release/utils/git.js`:

```js
const isBreaking = parsedMessage.breaking || commit.body.includes('BREAKING CHANGE:');
```

That is an `||`, and the right-hand side is a plain substring check over the whole body. My original commit had *both* markers, so dropping just the `!` would have left `isBreaking === true` and still produced `0.2.0`. I removed the footer too and reworded it to a `Migration note:` line, so:

| | specifier | after `adjustSemverBumpsForZeroMajorVersion` | version |
|---|---|---|---|
| before | `major` | `minor` | `0.2.0` |
| now | `minor` | `patch` | **`0.1.14`** |

**Two consequences to decide on:**

1. **The migration note no longer reaches the release notes automatically.** Without a breaking marker nx emits no "Breaking Changes" section, so `Node 20 is no longer supported; upgrade to Node 22+.` now lives only in the commit body and the PR description. If it should show up in the `0.1.14` notes it needs adding to the GitHub release by hand.

2. **This contradicts AC4 of the story**, which specifies a minor bump precisely *because* 0.x breaking → minor. Your call overrides it for this PR, but thymianofficial/thymian-internal#406 should be amended to match so the two do not stay out of sync.

Happy to restore the markers if the intent was only to change how it reads rather than to change the bump.

Dariant31:
Added Node 26 to the matrix in `250b8f20` — `version: [22, 24, 26]`, so 9 lanes.

Rationale for putting it in this PR rather than a follow-up: `>=22` is an open range and already permits 26, so shipping the floor without a 26 lane would reintroduce the same permit-but-never-verify gap this story exists to close.

I ran the full suite locally on **26.7.0** before adding the lane, rather than just asserting it works:

| Check | Result |
| --- | --- |
| `npm ci` (clean, npm 11.19) | ✅ 2053 packages, no `EBADENGINE` |
| `nx run-many -t build` | ✅ 16 projects, astro included |
| `nx run-many -t lint` | ✅ 17 projects |
| `nx run-many -t test` | ✅ 157 tests |
| `nx e2e e2e-tests` | ✅ 9 files, 55 tests |

The e2e run is the one that actually matters here — it publishes all 15 packages to Verdaccio, does a real global `npm install` of the CLI and shells out to the installed binary, so 26 is covered through the user path and not only under vitest. Native side is fine too: ABI moves 127 → 147 between 22 and 26, and `better-sqlite3` loads unchanged because it is N-API with an in-tarball prebuild.

**I left the docs advertising LTS 22 and 24 only.** Node 26 is `lts: false` until October 2026, so we now test more than we promise, which seems like the right asymmetry — happy to flip the docs too if you would rather advertise it.

Two things I noticed that are worth knowing, neither blocking:

- `better-sqlite3` is our de-facto ceiling. It is the one dependency that enumerates majors (`20.x || 22.x || … || 26.x`) instead of using an open range, so it will `EBADENGINE` on Node 28 until upstream adds it. Advisory only, since nothing sets `engine-strict`.
- npm ≥ 11 (bundled with **24** as well as 26) no longer runs install scripts by default, so `better-sqlite3`'s `prebuild-install || node-gyp rebuild` is skipped with a warning on both the workspace and the global install. Nothing breaks today only because the prebuilt binary ships inside the tarball — worth remembering if that ever stops being true.

If you would rather 26 not be able to block a merge while it is still the Current line, `continue-on-error: ${{ matrix.version == 26 }}` would make it informational until October. I left it as a normal gate on the assumption you want a real break there to be visible.

Dariant31:
Texts updated in `485788a8`, and the branch is now merged up to `main` (`64228336`).

| File | Now reads |
| --- | --- |
| `README.md` | Thymian requires **Node.js 22 or newer**. We build and test on Node **22**, **24** and **26**. |
| `CONTRIBUTING.md` | … `engines: { node: ">=22" }`, and CI builds and tests on Node **22**, **24** and **26**. |
| `getting-started.mdx` | You need **Node.js 22** or newer — we build and test on **22**, **24** and **26** |

One deliberate wording change: I dropped the "supported _LTS_ versions" framing from the README and the getting-started prerequisites. Node 26 is `lts: false` until October 2026, so listing it under an LTS heading would have been inaccurate. The notes now state the floor plus the versions we actually exercise, which is true of all three regardless of LTS status.

`sampler-getting-started.mdx` already said "Node.js 22+ installed" and stays as-is — still correct.

---

## ⚠️ Unrelated: `main` is currently broken

Heads-up, because it will turn this PR's checks red for a reason that has nothing to do with it. `origin/main` at `64228336` does not build:

```
packages/thymian/src/commands/explain/rule.ts(45,7): error TS2345:
  Argument of type 'NormalizedRuleSet[]' is not assignable to parameter of type 'string | string[]'.
```

I verified this on a **clean detached checkout of `origin/main`** with none of my changes present — `nx build thymian` exits 1. For completeness, my `packages/` tree differs from `main` only by the 15 `engines` blocks, which cannot affect `tsc`.

It looks like a semantic conflict between two PRs that were each green in isolation:

- **#348** (rule-configuration profiles) introduced `toRuleSetInputs()` and changed callers to split rule sets into specifiers + profiles — `rules/list.ts` now does `const { rules: ruleSpecifiers, ruleProfiles } = toRuleSetInputs(mergeRuleSets(...))`.
- **#358** (`thymian explain rule`) passes `mergeRuleSets(...)` straight into `loadRules()`, which still takes `string | string[]`.

Neither branch's `nx affected` saw the other's change, so CI passed on both and the break only materialised once both were on `main`.

The fix looks like making `explain/rule.ts` follow the same shape as `rules/list.ts`:

```ts
const { rules: ruleSpecifiers, ruleProfiles } = toRuleSetInputs(
  mergeRuleSets(this.thymianConfig.ruleSets, this.flags['rule-set']),
);

const rules = await loadRules(
  ruleSpecifiers,
  () => true,
  this.thymianConfig.rules,
  this.flags.cwd,
  ruleProfiles,
);
```

I have deliberately **not** fixed it here — it belongs to the `explain rule` feature rather than a Node-support-floor PR, and burying it in this diff would hide it from whoever owns those changes. Happy to open a separate PR for it if useful; this one can then merge on a green `main`.

matthyk (commented):
One completeness gap before this is good to merge:

**`packages/plugin-spectral/package.json` is missing the `engines` floor.**

The description says `engines` was added to "all 15 packages under `packages/*` … exactly the set `nx release` publishes." But `packages/` actually contains **16** package directories. `@thymian/plugin-spectral` has `publishConfig.access: "public"`, is not `private`, and matches `release.projects: ["packages/*"]` — so it publishes on the next `nx release`, yet it is the one published manifest without an `"engines"` field (confirmed on this branch).

Left as-is, `@thymian/plugin-spectral` ships without the Node ≥ 22 declaration, reopening the exact permit-but-don`t-declare gap this story closes for the other 15 packages.

Fix (one line):
- Add `"engines": { "node": ">=22" }` to `packages/plugin-spectral/package.json`
- Update "15 packages" → "16" in the PR description

Everything else checked out clean — CI matrix bump, dropped astro/Node-20 build exclusion, and the doc updates all look correct.

Dariant31:
Good catch — confirmed and fixed.

`@thymian/plugin-spectral` landed on `main` after this branch had covered the original 15, and the merge from `main` brought it in as a 16th published manifest with no `engines` field. Not `private`, `publishConfig.access: "public"`, matches `release.projects: ["packages/*"]` — so exactly as you describe, it would have shipped without the Node ≥ 22 declaration.

- **`fc57c38c`** — `"engines": { "node": ">=22" }` on `packages/plugin-spectral/package.json`, same position as its siblings (after `publishConfig`, before `keywords`).
- PR description updated: 15 → 16, with a note on why the count moved.

**One more thing surfaced while verifying it** (`7a9a6496`): npm mirrors `engines` into the lockfile's workspace entries, and `package-lock.json` had never been regenerated after the manifests changed — so all **16** `packages/*` entries were missing it, not just `plugin-spectral`. `npm ci` does not fail on this (it reconciles dependencies, not `engines` metadata), which is why CI stayed green and it went unnoticed; the cost was that every `npm install` regenerated the same diff for the next person. Regenerated it — diff is engines-only, no dependency version or `resolved`-URL drift.

Verification on Node 22.19.0, full workspace:

| Check | Result |
| --- | --- |
| `nx run-many -t build` | ✅ 17 projects |
| `nx run-many -t lint` | ✅ 18 projects |
| `nx run-many -t test` | ✅ 17 projects, 1410 passed / 1 todo |
| all 16 manifests + 16 lock entries `engines.node === ">=22"` | ✅ |
| root / `astro-docs` / `e2e-tests` have no `engines` | ✅ |

**`nx format:check` is worth flagging:** it does not pass on this branch, but nothing here caused it. It flags 7 files (2 `astro-docs` pages, 5 `docs/arc42` pages) that are byte-identical to `origin/main` and already unformatted there — verified by running prettier against both refs. Pre-existing on `main`, so I left them rather than sweeping unrelated reformatting into this diff. Happy to fix separately if you want.

I did not re-run the Node 26 suite — the change is a manifest field and the matrix exercises 26 on every push. The 26.7.0 table in the description predates `plugin-spectral` (15 packages, not 16); annotated as such rather than restated as if re-measured.

matthyk (approved):

PR #358, opened 2026-08-04
feat(thymian): add thymian explain rule command

matthyk:
This PR adds the `thymian explain rule <rule-name>` command.

For example running `thymian explain rule rfc9110/intermediary-must-parse-and-remove-connection-fields` prints

```txt

RULE
rfc9110/intermediary-must-parse-and-remove-connection-fields

SUMMARY
Intermediary MUST parse and remove Connection fields before forwarding.

DESCRIPTION
Intermediaries MUST parse a received Connection header field before a message is forwarded and, for each connection-option in this field, remove any header or trailer field(s) from the message with the same name as the connection-option, and then remove the Connection header field itself (or replace it with the intermediary's own control options for the forwarded message).

EXPLANATION
Before a proxy passes a message on, it must read the incoming Connection header, and for every field name listed there it must strip out that matching header or trailer, then drop the Connection header itself (or replace it with its own options for the next hop). Those listed fields are meant only for the current connection ("hop-by-hop"); if they leak downstream they can misconfigure or confuse the next recipient. Removing them keeps connection-specific control information from being blindly forwarded.

SEVERITY
✖ error

APPLIES TO
intermediary

REFERENCE
https://www.rfc-editor.org/rfc/rfc9110.html#name-connection
```

to the console.

### DIFF
--- a/packages/thymian/src/commands/explain/rule.ts
+++ b/packages/thymian/src/commands/explain/rule.ts
@@ -12,7 +12,9 @@ import {
 
 export default class ExplainRule extends BaseCliRunCommand<typeof ExplainRule> {
   static override description =
-    'Explain a single rule: what it checks, why it matters, and how to act on it.';
+    'Explain a single rule: what it checks, why it matters, and how to act on it. ' +
+    'Any rule in the configured rule sets can be explained regardless of its ' +
+    'severity, so the inherited --rule-severity flag has no effect here.';
 
   static override examples = [
     '<%= config.bin %> <%= command.id %> rfc9110/origin-server-should-send-501-response-for-unrecognized-method',
@@ -52,9 +54,10 @@ export default class ExplainRule extends BaseCliRunCommand<typeof ExplainRule> {
       );
     }
 
-    // Canonical lookup keyed by the stable `<ruleset>/<name>` id that reports use.
-    const ruleMap = new Map(rules.map((rule) => [rule.meta.name, rule]));
-    const rule = ruleMap.get(this.args.ruleName);
+    // Look up by the stable `<ruleset>/<name>` id that reports use. `find`
+    // takes the first match, so an earlier rule set wins over a later
+    // same-named rule instead of it being silently shadowed.
+    const rule = rules.find((r) => r.meta.name === this.args.ruleName);
 
     if (!rule) {
       this.error(
@@ -68,9 +71,9 @@ export default class ExplainRule extends BaseCliRunCommand<typeof ExplainRule> {
 
 /**
  * Render focused, educational detail for a single rule. Optional fields are
- * omitted entirely when absent — no empty headings or placeholders. Field order
- * follows UX Decision 11: id → summary → description → "Why this matters"
- * (explanation) → "Recommendation" → Severity → Applies to → Reference.
+ * omitted entirely when absent — no empty headings or placeholders. Section
+ * order (per UX Decision 11): RULE → SUMMARY → DESCRIPTION → EXPLANATION →
+ * RECOMMENDATION → SEVERITY → APPLIES TO → REFERENCE.
  *
  * Content is identical across TTY and non-TTY; only color (stripped by
  * `oclif.ux.colorize` when the stream is not a TTY) differs. Unicode severity
--- a/packages/thymian/src/commands/explain/rule.ts
+++ b/packages/thymian/src/commands/explain/rule.ts
@@ -90,30 +90,38 @@ function formatRuleDetail(rule: Rule): string[] {
 
   const lines: string[] = [];
 
-  lines.push('');
-  lines.push(oclif.ux.colorize('bold', 'RULE'));
-  lines.push(name);
-  lines.push('');
+  // Each section is a blank spacer + bold heading + value. Optional fields are
+  // pushed only when present, so absent fields leave no empty heading (AC3).
+  const section = (heading: string, value: string): void => {
+    lines.push('', oclif.ux.colorize('bold', heading), value);
+  };
 
-  lines.push(oclif.ux.colorize('bold', 'SEVERITY'));
-  lines.push(formatSeverity(severity));
+  section('RULE', name);
+
+  if (summary) {
+    section('SUMMARY', summary);
+  }
+
+  if (description) {
+    section('DESCRIPTION', description);
+  }
 
   if (explanation) {
-    lines.push('');
-    lines.push(oclif.ux.colorize('bold', 'EXPLANATION'));
-    lines.push(explanation);
+    section('EXPLANATION', explanation);
   }
 
   if (recommendation) {
-    lines.push('');
-    lines.push(oclif.ux.colorize('bold', 'RECOMMENDATION'));
-    lines.push(recommendation);
+    section('RECOMMENDATION', recommendation);
+  }
+
+  section('SEVERITY', formatSeverity(severity));
+
+  if (appliesTo && appliesTo.length > 0) {
+    section('APPLIES TO', appliesTo.join(', '));
   }
 
   if (url) {
-    lines.push('');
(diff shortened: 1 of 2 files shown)


BaggersIO (commented):
## Code review

**What it does:** adds `thymian explain rule <rule-name>`, a read-only command that loads every rule from the configured rule sets and prints one rule's metadata (id, summary, description, explanation, recommendation, severity, applies-to, reference) as a sectioned, colorized block. Two new files, no changes to existing code.

### Verification

I fetched the branch and staged the two files onto the exact merge-base (`73894779`, so all imports resolve), then cleaned up afterwards:

- `nx build thymian` — **compiles clean** under `tsconfig.lib.json`
- `vitest run test/commands/explain/rule.test.ts` — **6/6 pass**
- End-to-end via `bin/run.js`: renders correctly, exit 0; unknown rule → **exit 2** with the intended message; `thymian explain` topic help resolves
- **Threshold independence actually works**: a `hint` rule explains fine under the default `ruleSeverity: error`, and a rule set to `off` in `thymian.config.yaml` explains and renders `off`. This is the PR's core design claim and it holds against real rule sets, not just the mock.

The command itself is in good shape. Everything below is test quality plus small polish — no correctness bug in shipped behavior.

---

### 1. Three real TypeScript errors in the test file (invisible to CI)

`packages/thymian/test/commands/explain/rule.test.ts:150-152`

```ts
const filter = mockedLoadRules.mock.calls[0]![1];
expect(filter(makeRule({ name: 'w', severity: 'warn' }))).toBe(true);
```

`loadRules`' second parameter has a default (`ruleFilter: RuleFilter = () => true`), so `mock.calls[0]![1]` is `RuleFilter | undefined`. Under the repo's `strict` + `noUncheckedIndexedAccess`, calling it is `TS2722: Cannot invoke an object which is possibly 'undefined'` — I confirmed 3× TS2722 with a targeted `tsc` run.

It doesn't fail CI only because test files are never typechecked: `tsconfig.lib.json` includes `src/**` only, and vitest strips types via esbuild. (The package's `tsconfig.json` *does* include `test/**`, but it's pre-existing broken — `rootDir: src` makes it error on every test file, this PR's and the six existing ones alike.)

Fix: `const filter = mockedLoadRules.mock.calls[0]![1]!;`, or `expect(filter).toBeDefined()` first.

### 2. The non-TTY test is tautological — it asserts nothing about the command

The `'emits plain, symbol-preserving output in non-TTY mode'` test asserts `stdout` contains no `ESC`. But `captureOutput` defaults to `stripAnsi: true` (`@oclif/test/lib/index.js:61`) and strips escapes **unconditionally**, TTY or not. I proved it with a throwaway test that writes raw `ESC[1mBOLD` straight to `process.stdout`: it passes `not.toContain(ESC)`, and only retains escapes under `{ stripAnsi: false }`.

So this test passes even if the command force-colorized in a pipe — it can never fail for the reason it claims to guard. The inline comment (*"Captured (non-TTY) output carries no ANSI color escapes"*) attributes the stripping to the wrong layer; it's the capture helper, not the TTY.

Fix: `captureOutput(fn, { stripAnsi: false })`, then assert no escapes. That actually exercises `ux.colorize`'s TTY detection. The `ERROR_SYMBOL` half of the test is fine and worth keeping.

### 3. No test asserts field order

Field order is the PR's documented contract (*"Field order follows UX Decision 11"*, `rule.ts:69-78`), but every assertion is `toContain` — order-insensitive. Reordering the `section()` calls would not fail a single test.

A `toMatchInlineSnapshot()` on the full stdout in the fully-populated case would lock down order *and* the blank-line spacing, and would replace ~15 `toContain` lines.

### 4. Two untested branches

- `rules.length === 0` → `'No rules are loaded.'` (`rule.ts:49-53`)
- `severity === 'off'` → dimmed render (`rule.ts:138-140`). I verified this manually; nothing locks it in.

### 5. Mocking `loadRules` elides the one thing most likely to regress

`vi.mock` replaces `loadRules` wholesale, so the filter test asserts what the *filter* returns, not that `loadRules` with `() => true` actually yields `off`/`hint` rules. That integration is the command's whole premise. Given `test/commands/` already hosts `*.integration.test.ts` files, one unmocked test against `@thymian/rules-rfc-9110` explaining a known `hint` rule would cover it — I did exactly that by hand and it passes.

---

### Smaller points

- **`rule.ts:55-57`** — a `Map` is built over all rules for a single `.get()`. `rules.find((r) => r.meta.name === this.args.ruleName)` is simpler, skips the allocation, and takes the *first* match instead of silently letting a duplicate rule name from a later rule set shadow an earlier one. The comment *"Canonical lookup keyed by..."* oversells a one-shot lookup.
- **`rule.ts:69-78`** — the doc comment promises headings `"Why this matters"` and `"Recommendation"`, but the code emits `EXPLANATION` and `RECOMMENDATION`. Align one to the other.
- **`--rule-severity` is advertised but inert here.** It's a `baseFlags` entry so it shows in `explain rule --help`, yet the command deliberately ignores it. The inherited-flag noise is pre-existing (`rules list` carries `--spec`/`--traffic` too), but `--rule-severity` on a command whose selling point is ignoring the threshold is actively misleading. A clause in `description` would cost nothing.
- **Style consistency win worth noting:** using `SEVERITY_COLORS`/`SEVERITY_SYMBOLS` from core is better than the hand-rolled `colorizeSeverity` switch in `commands/rules/list.ts:161-174`. Follow-up opportunity: collapse that sibling onto the same shared constants.
- **Performance:** explaining one rule imports every rule in every configured rule set (hundreds for rfc-9110), and `shouldAutoload()` still registers all 8 default plugins first. Both are inherent to the existing architecture and identical to `rules list`, so not this PR's problem — just don't expect it to feel instant.

### Security

Clean. The `ruleName` arg is used only as a lookup key and in an error message — it never reaches `require.resolve`, `import()`, or the filesystem, so there's no path-traversal or arbitrary-import vector from the new input. The dynamic imports inside `loadRules` are driven by config-declared rule sets, unchanged by this PR. No secrets, no shell-out, no network.

### Verdict

Good to go once #1 (broken types) and #2 (the test that can't fail) are addressed. #3–#5 are the difference between tests that document intent and tests that enforce it — worth doing while the file is fresh, but they don't block. Conventions are followed closely throughout: `requiresSpecifications = false`, `EOL` joining, `oclif.ux.colorize`, `OCLIF_TEST_ROOT` depth, `vi.mock`-before-import ordering, and the `--no-autoload` test flag all match the surrounding code, and the *why*-oriented comments are better than the repo average.

BaggersIO (approved):
Approving — the shipped command is solid. I built and ran it against the real rule sets on the merge-base: renders correctly, exits 2 with a clear message on an unknown rule, and the threshold-independent loading genuinely works (a `hint` rule and a config-disabled `off` rule both explain fine under the default `ruleSeverity: error`). No security surface — the `ruleName` arg never reaches `require.resolve`, `import()`, or the filesystem.

Everything in my [review comment](https://github.com/thymianofficial/thymian/pull/358#issuecomment) is test quality, not shipped behavior, so I don't think it should gate the merge. Two of them are one-liners I'd still like to see land, ideally before merge:

1. `rule.test.ts:150-152` — `mock.calls[0]![1]` is `RuleFilter | undefined` (the param has a default), so calling it is `TS2722` under the repo's `strict` + `noUncheckedIndexedAccess`. Add the `!`. CI can't catch this today because `tsconfig.lib.json` only typechecks `src/**`.
2. The `'emits plain, symbol-preserving output in non-TTY mode'` test can't fail — `captureOutput` strips ANSI unconditionally (`stripAnsi` defaults to `true`), so it would pass even if the command force-colorized in a pipe. Pass `{ stripAnsi: false }` to make it actually test the TTY detection.

The rest (field-order assertion, the two untested branches, an unmocked integration test) are fine as follow-ups.

PR #362, opened 2026-08-11
feat: `thymian report merge` — merge multiple reports into one Thymian report

atennert:
Implements **thymianofficial/thymian-internal#507**: a `thymian report merge` command that merges N typed report inputs — persisted Thymian JSON reports and/or convertible external formats (e.g. Spectral) — into one Thymian report.

```sh
thymian report merge \
  --report thymian:./my-thymian-report.json \
  --report spectral:./my-spectral-report.json \
  --spec openapi:./api.yaml
```

## What's in here

- **core**: `ConvertedRunFragment` gains an optional `thymianFormat` record (the only contract change — additive, backward-compatible); `Thymian.reportConvert()` unions fragment format maps by hash (first-wins, prototype-safe, junk values skipped) and `finalizeWorkflow()` folds in the spec-loaded format. Persisted `thymianFormat` locations stay resolvable after a merge.
- **plugin-reporter** *(reworked 2026-08-13 — see "Course-correction rework" below; originally a new `@thymian/plugin-thymian-report` package)*: claims `thymian:` inputs on `core.report.convert` (ADR-0016/0017); accepts the JSON formatter's native `Report[]` array or a bare `Report`, validates against the loose `reportSchema` (persisted-report compatibility contract), tolerates a UTF-8 BOM, and passes runs through identity-preserved with their format maps. Empty/run-less files raise input-named errors (team-ratified semantics). The package now owns the persisted-report file boundary in both directions: its JSON formatter writes exactly what its loader reads back.
- **common-cli**: claim-enforcement block extracted from `convert.ts` into shared `enforceReportClaims()` (was flagged drift-prone in the 14.2 review); plugin-reporter (already autoloaded) carries the `thymian:` claim; config-schema twins updated (`reports` drives convert *and* merge).
- **thymian CLI**: `report merge` command — repeatable `--report <type>:<location>` with config fallback, spec passthrough, `handleWorkflowOutcome` (ADR-0015 exit codes). `convert.ts` refactored onto the shared helper, behavior unchanged.
- **e2e**: 15 published-CLI cases incl. a convert→persist→merge chain (endpoint resolution without re-supplying `--spec`), two-different-format-hash resolution, foreign+foreign and thymian+thymian pairings, identity merge, duplicate collapse, determinism (normalized), error paths, stream separation.
- **docs**: new "Merging reports" tutorial + cross-link from the Spectral integration tutorial.

## Merge semantics (team-ratified)

One `Report` whose `runs` array concatenates all input reports' `ToolRun`s in input order (never folded — the `ToolRun` union and per-run `rules[]`/`thymianFormatVersion` scoping forbid it); all involved formats listed in the merged `thymianFormat` record; fresh `reportId`/`createdAt`; no finding dedup in v1; provenance out of scope (model v4 frozen).

## Verification

- Unit: core 338, plugin 12, thymian 271, common-cli 90 — all green; red-green followed per layer.
- E2E: **85/85** (full suite, Verdaccio-published CLI).
- Adversarial code review (3 parallel layers, 25 verified findings): 11 patches applied in this branch, 1 decision ratified, 5 pre-existing items deferred as tracked issues (thymian-internal#516–#520), 8 dismissed.

## Course-correction rework (2026-08-13)

Per the approved Sprint Change Proposal on thymianofficial/thymian-internal#507 (course-correction section of the issue), `@thymian/plugin-thymian-report` is **rescinded** — it introduced a 17th package without an accepted ADR naming it. Rework commits `7fb4f4d9..b2af55a0` (5 commits since the last review pass):

- **Relocation, behavior-identical**: the loader (`load-thymian-report.ts`) and its 12 tests moved into `@thymian/plugin-reporter`, which now also listens on `core.report.convert` and claims `thymian:` inputs (handler in `src/thymian-report-input.ts`, registered before the first `await` so formatter-setup failure can't strand the declared claim). Error strings byte-identical; identity-preserved passthrough unchanged.
- **Package deletion**: `packages/plugin-thymian-report/` removed with all wiring (tsconfig references, CLI dependency, commitlint scope, license allowlist, default-config autoload, e2e fixture entry, lockfile). Back to **16 packages**; zero in-tree mentions remain.
- **ADR-0017 amended** (Edits A–C from the proposal): claims live in existing plugins where the domain fits; a **new package requires an accepted ADR naming it before implementation**; native `thymian:` → `@thymian/plugin-reporter`. Header carries the amendment signal; status stays Accepted.
- **Review pass on the rework diff** (adversarial, 2 hunters): 5 patches applied (registration ordering, ADR provenance markers, manifest read-side metadata, neutral fixture producer name, new JSON-formatter→loader round-trip unit test); 4 items deferred as thymian-internal#553–#556.
- **Verification**: 66 Nx tasks green (plugin-reporter 73/73 unit tests), full e2e **85/85** via published CLI, format clean. The one observed e2e flake (AC 6 determinism, feedback-tip line ordering) is pre-existing and now tracked as thymian-internal#556.


packages/core/src/thymian.ts
```diff
@@ -602,7 +610,24 @@ export class Thymian {
...
+      return matches.map((fragment) => {
+        if (fragment.thymianFormat) {
+          for (const [hash, serialized] of Object.entries(
+            fragment.thymianFormat,
+          )) {
+            // Skip junk values (e.g. null from a hand-edited persisted map)
+            // rather than copying them into the merged report.
+            if (!serialized || typeof serialized !== 'object') {
+              continue;
+            }
```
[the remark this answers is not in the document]
atennert: Fixed in c642cfa5 — the junk guard now also rejects arrays (`Array.isArray`), and the hardening test covers an array map value.

packages/core/src/thymian.ts
```diff
@@ -687,10 +716,18 @@ export class Thymian {
...
-      : undefined;
+    let thymianFormat: Report['thymianFormat'];
+
+    if (format || additionalFormats) {
+      thymianFormat = { ...additionalFormats };
+      if (format) {
+        // Keep an already-present entry (same hash ⇒ same graph).
+        thymianFormat[format.attributes.hash] ??= format;
+      }
+    }
```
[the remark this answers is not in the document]
atennert: Fixed in c642cfa5 — `finalizeWorkflow()` now assembles `thymianFormat` via `Object.assign(Object.create(null), additionalFormats)`, so the null-prototype hardening survives assembly and the `??=` hash lookup can't hit inherited `Object.prototype` members.

packages/thymian/src/commands/report/merge.ts
```diff
@@ -0,0 +1,65 @@
...
+  static override requiresSpecifications = false;
+
+  static override flags = {
+    report: reportFlag(),
+  };
+
+  override async run(): Promise<void> {
+    // Flags-over-config for --report, mirroring Step C's --spec handling;
+    // resolved here because --report is command-level, not a baseFlag.
+    const reports: ReportInput[] = this.flags.report?.length
```
matthyk: Is this behavior intended? By this the reports given via CLI would override the reports from the config. We could also discuss if we want to append them?

packages/thymian/src/commands/report/merge.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    // Merge = convert-then-assemble: every input is normalized to ToolRuns
+    // via core.report.convert (thymian: inputs pass through unchanged, with
+    // their format maps; foreign inputs are converted, against --spec when
+    // given) and core assembles exactly one Report in input order. A single
+    // input is a valid identity merge; duplicates collapse in core.
+    const outcome = await this.thymian.run(() =>
+      this.thymian.reportConvert({
+        reports,
+        specification: this.thymianConfig.specifications ?? [],
```
matthyk: **`merge` loads every config `specifications` entry even when no input needs a spec.**

With `specifications: [openapi: ./api.yaml]` in `thymian.config.yaml`, `thymian report merge --report thymian:a.json --report thymian:b.json` (two pure Thymian reports, neither needing a spec):

1. If `api.yaml` is missing or invalid, `reportConvert` → `loadFormat` fails and the merge dies with a spec error even though no input required one.
2. If it loads, `finalizeWorkflow` adds its hash to `report.thymianFormat`, so the output JSON carries a serialized graph no run points at.
3. Worst: it flips endpoint resolution. With only fragment formats the map has 1 entry, so `resolveThymianFormatForRun`'s sole-entry fallback renders endpoints for runs whose `thymianFormatVersion` is absent. Add the config spec and there are 2 entries, the fallback switches off, and those locations degrade to `format:<elementId>` — the rendered output of the same merge changes based on whether an unrelated config key is present.

Consider loading specs lazily, or only when at least one claimed input actually needs one.

packages/thymian/src/commands/report/merge.ts
matthyk: **This command is a verbatim clone of `convert.ts`.** The flags-over-config resolution, the empty-input error string, the `reportConvert` call, `enforceReportClaims`, and `handleWorkflowOutcome` are byte-identical — only `description`, `examples`, and one comment differ.

The PR extracted `enforceReportClaims` precisely because the duplicated claim block was flagged drift-prone in the 14.2 review, then left the ~30 surrounding lines duplicated — including the user-visible `'No report input found. Provide one with --report or add reports to your configuration file.'` in two places.

Any future change to `--report` precedence (e.g. moving to append semantics to match `mergeSpecifications`/`mergeTraffic` in `packages/common-cli/src/merge-inputs.ts`) or to that error text now has to be made twice, and whoever edits one command gets no signal a twin exists. A shared `BaseReportConvertCommand`, or a `resolveReportInputs(command)` helper in common-cli next to `enforceReportClaims`, would fix it at the right depth.

packages/common-cli/src/report-claim-enforcement.ts
```diff
@@ -0,0 +1,81 @@
+import type { Command } from '@oclif/core';
+import type { ReportInput } from '@thymian/core';
+
+/**
+ * Claim enforcement for `--report` inputs (ADR-0017): a type is supported
+ * exactly when >=1 registered plugin claims it, so the supported list is
+ * derived from this run's claims. Errors with exit 2 (usage error) when any
+ * input went unclaimed; call before any report rendering. Shared by
```
matthyk: **Two things on this module.**

**1. The ordering promise is not kept — a usage error still writes a report.** The docstring says "call before any report rendering", but both call sites invoke `enforceReportClaims` *after* `this.thymian.run(...)` has already emitted `core.report` and flushed the file formatters.

`thymian report merge --report thymian:ok.json --report bogus:x.json` with the json/markdown/csv formatters enabled (as in `e2e-tests/fixtures/report-merge/thymian.config.yaml`): `Thymian.run()` closes the instance, `core.close` runs `flushReporters()`, and `.thymian/reports/report.{md,json,csv}` land on disk containing a **partial** merge (only `ok.json`'s runs). *Then* this errors with exit 2. A CI job treating exit 2 as "nothing produced" publishes the truncated artifact as the full merge. The e2e suite asserts no-files-written for the loader-throw path (`cli-report-merge.test.ts:279`) but never for the unclaimed path.

**2. No unit test in this package** — and `packages/common-cli`'s vitest config sets `passWithNoTests: true`, so the gap is invisible. Every branch here (empty-`supportedTypes` path and its suggestion, the unsupported-vs-unclaimed-of-supported-type split, singular/plural variants, the dedup-aware `supportedTypes` derivation) is exercised only indirectly from two oclif command test files in a *different* package. The plural branches (`unsupported.length > 1`, `unclaimedOfSupportedType.length > 1`) have no coverage at all. The point of the extraction was to make this an owned unit — it should carry `packages/common-cli/test/report-claim-enforcement.test.ts`.

e2e-tests/src/cli-report-merge.test.ts
```diff
@@ -0,0 +1,421 @@
...
+        'openapi:test.openapi.yaml',
+        '-o',
+        '@thymian/plugin-reporter.formatters.json.path=converted.json',
+      ],
+      { cwd: getTempDir() },
+    );
+    expect(convert.exitCode).toBe(1); // findings, not an error
+    expect(existsSync(join(getTempDir(), 'converted.json'))).toBe(true);
+
+    // Step 2: merge the persisted report WITHOUT --spec — endpoint
```
matthyk: **This test cannot fail, and the comment on this line is not true.** `copyFixturesToTempDir` copies `thymian.config.yaml`, whose `specifications: [openapi: test.openapi.yaml]` is picked up by `merge.ts` (`specification: this.thymianConfig.specifications ?? []`) — so a spec *is* supplied, via config rather than `--spec`.

Delete the entire `thymianFormat` passthrough this PR adds (the `fragmentFormats` union in `thymian.ts` and `thymianFormat: report.thymianFormat` in `thymian-report-input.ts`) and re-run: `report merge` still loads `test.openapi.yaml` from the copied config, `finalizeWorkflow` still inserts that format under the identical content hash, `stdout`/`merged.md` still contain `200 OK - */*`, and this passes green.

So the only e2e proof of the PR's headline feature — endpoint resolution without re-supplying the spec — proves nothing. Same weakening applies to the foreign+foreign test at line 170. Dropping `specifications` from the fixture config (or using a fixture dir without it) would make these tests actually load-bearing.

packages/core/src/thymian.ts
```diff
@@ -687,10 +721,23 @@ export class Thymian {
...
   ): Report {
-    const thymianFormat = format
-      ? { [format.attributes.hash]: format }
-      : undefined;
+    let thymianFormat: Report['thymianFormat'];
+
+    if (format || additionalFormats) {
+      // Null-prototype like the fragment union above: hash keys are external
+      // data, so prototype members must not shadow or leak into lookups.
+      thymianFormat = Object.assign(
```
matthyk: **Merge-only hardening pushed into the shared finalizer, changing a public shape for every workflow.**

This branch (`if (format || additionalFormats)`) is taken on every `lint`/`test`/`analyze` run too — where `additionalFormats` is always `undefined` and the input is trusted, spec-derived data, not an external file. `report.thymianFormat` used to be `{ [hash]: format }` with `Object.prototype`; it now has a `null` prototype on all of them.

Any consumer — a third-party plugin on `core.report`, a custom formatter — doing `report.thymianFormat.hasOwnProperty(h)`, `.toString()`, or `instanceof Object` throws or flips result, and `expect(report.thymianFormat).toStrictEqual({...})` fails.

The hardening belongs where untrusted data actually enters: the fragment union at line 599, or `loadThymianReports`.

packages/core/test/thymian.test.ts
```diff
@@ -500,6 +500,147 @@ describe('Thymian.reportConvert()', () => {
...
+            tool: { name: 'hostile-reader' },
+            runType: 'lint',
+            executions: [],
+            thymianFormatVersion: 'constructor',
+          }),
+          // 'constructor' collides with an Object.prototype member; 'bad'
+          // and 'worse' carry junk values a hand-edited persisted map could
+          // contain (arrays are typeof 'object' too).
+          thymianFormat: {
+            constructor: format,
```
matthyk: **This regression test passes without the hardening it claims to cover.** `constructor` is not a discriminating key: `Object.hasOwn({}, 'constructor')` is `false`, and `obj['constructor'] = format` creates an ordinary own property. So `Object.keys(...)` is `['constructor']` and `thymianFormat['constructor'] === format` both hold whether the accumulator is `Object.create(null)` or `{}`.

Revert both `Object.create(null)` calls in `thymian.ts` (lines 599 and 731) to `{}` and this test still goes green.

The actual hazard is untested: with a plain object, a persisted `"thymianFormat": {"__proto__": {...}}` — and `JSON.parse` *does* create `__proto__` as an own property — hits the `__proto__` setter, the entry silently vanishes from `Object.keys`, and the accumulator's prototype is replaced by the serialized format. `__proto__` is the key this test needs.

packages/thymian/test/commands/report/merge.test.ts
```diff
@@ -0,0 +1,349 @@
...
+  afterEach,
+  beforeAll,
+  beforeEach,
+  describe,
+  expect,
+  it,
+  vi,
+} from 'vitest';
+
+process.env.OCLIF_TEST_ROOT = join(import.meta.url, '../../../..');
```
matthyk: **`path.join` on a `file://` URL — this env var does not point where it looks like it does.**

`join('file:///Users/x/repo/packages/thymian/test/commands/report/merge.test.ts', '../../../..')` returns `'file:/Users/x/repo/packages/thymian'` — `path.join` collapses the `//` in the URL scheme, so the result is neither a valid URL nor an existing directory (verified).

The file already imports `fileURLToPath` and uses it correctly two lines below:

```suggestion
process.env.OCLIF_TEST_ROOT = fileURLToPath(new URL('../../../..', import.meta.url));
```

As written, whatever this line was meant to guarantee about command discovery is not in force — and the `it.skipIf(!existsSync(dist/commands/report/merge.js))` registration test at the bottom silently skips on any unbuilt checkout, so nothing else covers it either.

packages/thymian/test/commands/report/merge.test.ts
```diff
@@ -0,0 +1,349 @@
+import { EventEmitter } from 'node:events';
```
matthyk: **This file is a ~90% clone of `convert.test.ts`** (349 lines): the whole `vi.mock('@thymian/core')`/`MockThymian` block, the `mockState` shape, the `beforeAll`/`afterAll` tmpDir dance, and five near-identical claim-enforcement cases. A `diff` of the two shows only the command import, the fixture report tool names, and three added cases as real differences.

Both files now assert the *same* `enforceReportClaims` messages against the *same* shared implementation — so four duplicated cases here test common-cli code twice, while `enforceReportClaims` has no test in its own package.

Suggest: `MockThymian` → a shared `packages/thymian/test/helpers/mock-thymian.ts`; the claim-enforcement assertions → `packages/common-cli/test/report-claim-enforcement.test.ts`.

packages/plugin-reporter/src/load-thymian-report.ts
```diff
@@ -0,0 +1,81 @@
...
+      `Failed to read Thymian report "${inputLabel}" (resolved to ${filePath}).`,
+      { cause: err instanceof Error ? err : undefined },
+    );
+  }
+
+  let parsed: unknown;
+  try {
+    // Tolerate a UTF-8 BOM (common for files saved by Windows editors) —
+    // JSON.parse rejects it with a misleading syntax error otherwise.
+    parsed = JSON.parse(content.replace(/^\uFEFF/, ''));
```
matthyk: **BOM tolerance exists only for `thymian:` inputs — the same command answers the same file boundary two different ways.**

This file and `thymian-report-input.ts` are near-verbatim copies of plugin-spectral's `load-spectral-report.ts` and its `core.report.convert` listener, and the copies have already diverged right here.

`thymian report merge --report thymian:a.json --report spectral:b.json`, both saved by a Windows editor with a UTF-8 BOM: `a.json` is accepted by this `replace(/^\uFEFF/, '')`, `b.json` fails with `Failed to parse Spectral report "spectral:b.json" as JSON`, and the user has no way to tell why.

Also duplicated verbatim across the two packages: the `isAbsolute(location) ? location : join(cwd, location)` resolution, the read/parse `ThymianBaseError` wrapping with `inputLabel`, and the entire listener skeleton (type filter, empty-reply guard with the same "collect strategy waits for every registered listener (14.1 listener contract)" comment, `String(location)`, `inputLabel`, `fragments.push`, `ctx.reply`).

A shared helper in `@thymian/core` (`readTypedInputFile` + `registerReportInputClaim`) would fix the BOM asymmetry once and keep the next claimant consistent.

e2e-tests/fixtures/report-merge/spectral-findings.json
```diff
@@ -0,0 +1,121 @@
...
+      "end": {
+        "line": 3,
+        "character": 16
+      }
+    },
+    "source": "test.openapi.yaml"
+  },
+  {
+    "code": "operation-description",
+    "path": ["paths", "/users", "get"],
```
matthyk: **These findings name operations that don't exist in the fixture spec, so the endpoint assertions hold only by line-number coincidence.**

`test.openapi.yaml` declares exactly one path, `/api/hello`. This fixture's `path` arrays say `["paths","/users","get"]` and `["paths","/orders","post"]`. Endpoint mapping is line-based (findings at or after an endpoint's location in the same source file attach to that endpoint), so these 0-based ranges happen to land on `/api/hello` and the `200 OK - */*` assertions pass.

Any harmless reflow of the fixture spec — adding a `description`, reordering `info` — shifts the line numbers and breaks tests that look like they're about path mapping.

Relatedly, the comment at `cli-report-merge.test.ts:157` claims "the spectral fixture's source doesn't map onto test2's paths" — but `test2.openapi.yaml` puts `/api/goodbye` at 0-based line 6, exactly where the first `operation-description` finding sits, so those findings *do* map. The comment misdescribes what the test observes.

packages/thymian/tsconfig.json
```diff
@@ -23,10 +23,10 @@
       "path": "../plugin-websocket-proxy"
     },
     {
-      "path": "../plugin-sampler"
+      "path": "../plugin-spectral"
```
matthyk: Unrelated churn: `../plugin-sampler` and `../plugin-spectral` are swapped here and in `tsconfig.lib.json` with no functional effect — neither file gains or loses a reference (`../plugin-reporter` was already present). Almost certainly an `nx sync` artifact. It adds two hunks of noise to a 30-file diff, and the next `nx sync` on a branch that didn't pick it up will swap them back and produce a phantom conflict. Worth reverting.

packages/core/src/thymian.ts
```diff
@@ -591,6 +591,14 @@ export class Thymian {
     }
 
     const unclaimed: ReportInput[] = [];
+    // Union of fragment-carried format maps (#507): first occurrence wins per
+    // hash — equal hashes mean equal serialized graphs, so collisions are
+    // benign. Collected in assembly (input) order alongside the runs.
+    // Null-prototype object: persisted files are external input, so a hash
+    // key like 'constructor' or '__proto__' must behave as plain data.
+    const fragmentFormats: NonNullable<Report['thymianFormat']> = Object.create(
```
matthyk, line 609: **Building this map as a union across inputs breaks the assumption `resolveThymianFormatForRun` relies on.**

`packages/core/src/report/location-format.ts:104-110` (untouched by this PR):

```ts
const entries = Object.entries(formats ?? {});
const serialized =
  runVersion !== undefined && formats?.[runVersion]
    ? formats[runVersion]
    : entries.length === 1
      ? entries[0]?.[1]
      : undefined;
```

That `entries.length === 1` fallback was safe while a report's formats all came from one workflow. This union spans inputs, so it isn't any more.

Merge report A (carries `thymianFormat: {H1}`, runs → H1) with report B whose runs have `thymianFormat`-typed locations but whose `thymianFormatVersion` is absent, or whose hash was never persisted. The merged map has exactly one entry, the fallback fires, and **every one of B's locations renders against A's API graph.** Element ids are semantic content hashes (`ThymianFormat.hash(...)`), so any endpoint hashing identically in both APIs renders B's finding at A's endpoint — wrong endpoint attribution, silently, no warning.

Worth gating the fallback on single provenance, or dropping it once a report can hold formats from more than one source.

packages/core/src/thymian.ts
```diff
@@ -591,6 +591,14 @@ export class Thymian {
...
+      null,
+    ) as NonNullable<Report['thymianFormat']>;
 
     const toolRuns = reports.flatMap((reportInput) => {
       const key = `${reportInput.type}:${String(reportInput.location)}`;
```
matthyk, line 614: **Identity is keyed on path, not content — the same report under two paths yields duplicate `runId`s in one merged report.**

This key (and the matching `seenInputs` filter at lines 547-556) treats `type:location` as the identity of an input. So:

```
cp report.json copy.json
thymian report merge --report thymian:report.json --report thymian:copy.json
```

Two distinct keys, both claimed, and `Report.runs` ends up holding two runs with identical `runId`, `runAt`, and executions. `reportSchema` doesn't enforce uniqueness, so this validates and persists.

The CSV formatter emits `run.runId` as the first column of every row (`packages/plugin-reporter/src/formatters/csv.ts:126,147`), so anything joining or de-duplicating on `runId` downstream silently merges or double-counts the two runs.

The docs warn about doubled *counts* ("Merging the same results twice doubles the summary counts") but say nothing about duplicate run *identities* — and the new tutorial explicitly promises "each run keeps its identity: its `runId` …" (`report-merge.mdx:27`). Content-hashing the input, or de-duplicating on `runId` at assembly, would close it.

astro-docs/src/content/docs/tutorials/report-merge.mdx
```diff
@@ -0,0 +1,91 @@
...
+File output works like every other workflow: enable formatters in your configuration file or override their paths ad hoc:
+
+```sh
+thymian report merge \
+  --report thymian:./a.json --report thymian:./b.json \
+  -o '@thymian/plugin-reporter.formatters.markdown.path=merged.md' \
+  -o '@thymian/plugin-reporter.formatters.json.path=merged.json'
+```
+
+Or persistently, in `thymian.config.yaml` — which can also carry the report inputs themselves, so a bare `thymian report merge` works:
```
matthyk: **This tutorial documents config-carried report inputs for `merge`, but the configuration reference still says that key is convert-only.**

The PR updated the `reports` description in both config-schema twins to say "convert *and* merge" and left `astro-docs/src/content/docs/references/configuration.mdx` saying:

- line 37: ``- `--report <type:path>` overrides `reports` (used by `thymian report convert`)``
- line 127: ``` `reports` declares the pre-generated external reports that `thymian report convert` imports ```

Someone reading the reference concludes `reports` is convert-only and never discovers the behaviour this tutorial depends on (and that the e2e AC-7 test exercises).

Separately, `references/configuration-schema.md` is headed "Auto-generated from the Thymian config JSON schema" yet has **no `reports` row at all** — verified at this head: 5 hits for `specifications`, 0 for `reports`. The schema edit in this PR is therefore invisible there, which means the generator isn't being re-run as part of the change.

packages/thymian/src/commands/report/merge.ts
atennert: As discussed, the logic is to be adjusted, so that report merge only reads CLI-Arguments, not the config.

atennert: Implemented in 1250b379: `report merge` now reads CLI arguments only — neither config `reports` nor config `specifications` feed a merge (general config like plugins/formatters still applies). Enforced by `resolveReportInputs()` in `merge.ts`, unit tests ("ignores config-file reports/specifications"), an e2e test that runs a bare `report merge` against a config full of reports and expects exit 2, and the tutorial now documents CLI-only inputs.

packages/thymian/src/commands/report/merge.ts
atennert: Fixed in 1250b379, on two axes. (1) Config `specifications` can no longer sneak in — merge is CLI-only per the decision above, so a spec is loaded only on an explicit `--spec` (and failing on a broken explicit spec is then correct). (2) The output-shape harms are gone regardless: the workflow spec joins the merged `thymianFormat` map only when a run actually converted against it (no orphan graph), and point 3 is moot because the sole-entry render fallback was removed entirely (see the thread on `thymian.ts:599`).

packages/thymian/src/commands/report/merge.ts
atennert: Fixed in 1250b379: both commands now extend a shared `BaseReportAssemblyCommand` in common-cli that owns the empty-input usage error, the `reportConvert` run, `enforceReportClaims`, and `handleWorkflowOutcome`. The one deliberate difference — input resolution — is the abstract `resolveReportInputs()` (convert: flags-over-config; merge: CLI-only), so the shared wording and sequence can no longer drift.

packages/common-cli/src/report-claim-enforcement.ts
atennert: Both fixed in 1250b379. (1) The ordering promise is now kept where it can be: core withholds the `core.report` emission when any input went unclaimed, so the file formatters never see (and never persist) a partial assembly — the CSV formatter additionally opens its stream lazily on the first report, so not even a header-only artifact lands. The docstring now describes that mechanism, and the unclaimed e2e case asserts no report.md/json/csv on disk. (2) `packages/common-cli/test/report-claim-enforcement.test.ts` now covers every branch: no-claimant hint + suggestion, unsupported singular/plural, supported-but-unclaimed singular/plural, the mixed case, dedup-aware supported-type derivation, and the wrong-location-does-not-mask-support case.

e2e-tests/src/cli-report-merge.test.ts
atennert: Fixed in 1250b379: merge now ignores config `specifications` entirely (CLI-only decision), so this test is load-bearing — the only spec the pipeline ever sees is the one persisted inside converted.json, and deleting the thymianFormat passthrough fails it. The fixture config deliberately keeps its `specifications` entry (with a warning comment) to prove the ignoring; the misleading comment was rewritten.

packages/core/src/thymian.ts
atennert: Fixed in 1250b379: the shared finalizer builds an ordinary plain-prototype object again — `hasOwnProperty`, `instanceof Object`, and `toStrictEqual` against literals all behave as before, asserted by a new core test. The hardening stays where untrusted data enters: the fragment union accumulates on the null-prototype object, and the finalizer copies those hash keys as own data properties (`Object.defineProperty`), so a persisted `__proto__` key lands as plain data without touching the prototype setter.

packages/core/test/thymian.test.ts
atennert: Fixed in 1250b379: the test now builds the hostile map via `JSON.parse('{"__proto__": {...}}')` — the actually discriminating vehicle, with a sanity assertion that the key is an own property — and asserts the entry survives in `Object.keys`, resolves as data, and that the resulting public map keeps an unpolluted `Object.prototype`. Reverting the hardening makes it fail.

packages/thymian/test/commands/report/merge.test.ts
atennert: Fixed in 1250b379 with `fileURLToPath(new URL(...))` — one note on the suggestion: with `new URL` the base file segment is dropped before resolving, so `'../../../..'` lands one level too high (`packages/`); the fix uses `'../../..'`, which resolves to the package root `packages/thymian` (what `@oclif/test`'s `findRoot()` expects). Applied to `convert.test.ts` too, which had the same `path.join` bug on the same line. The skipIf registration test is unchanged — it needs `dist/`, and the test target intentionally does not depend on build.

packages/thymian/test/commands/report/merge.test.ts
atennert: Fixed in 1250b379: `MockThymian`/`mockState` moved to a shared `packages/thymian/test/helpers/mock-thymian.ts` used by both files; the claim-enforcement message matrix moved to `packages/common-cli/test/report-claim-enforcement.test.ts` (where the code lives); `convert.test.ts` keeps one integration case for the path; and `merge.test.ts` is now merge-specific only (CLI-only input resolution + merged-outcome handling), with a comment pointing at where the shared coverage lives.

packages/plugin-reporter/src/load-thymian-report.ts
atennert: Fixed in 1250b379: core now owns the claimant boundary — `readTypedInputJson` (path resolution, BOM tolerance, read/parse `ThymianBaseError` wording) and `registerReportInputClaim` (type filter, always-reply-empty contract, stringified identity tags). Both plugin-reporter and plugin-spectral are refactored onto them, so `spectral:` inputs now accept a BOM'd file exactly like `thymian:` ones (new spectral test proves it), and the next claimant gets the same boundary for free.

e2e-tests/fixtures/report-merge/spectral-findings.json
atennert: Fixed in 1250b379: the findings now name elements that exist in `test.openapi.yaml` — `["paths","/api/hello","get"]` with ranges anchored at the real `get:` position, and a response-content finding at `.../responses/200/content/*/*` anchored inside the response (which is what exercises the `200 OK - */*` endpoint mapping). The misdescribing comment at cli-report-merge.test.ts:157 was rewritten to state the actual mechanism: run 2's findings keep file locations because their `source` (test.openapi.yaml) matches no node loaded from test2.openapi.yaml — `mapLocation` matches on source file before line proximity.

packages/thymian/tsconfig.json
atennert: Reverted to main in 1250b379 (both tsconfig.json and tsconfig.lib.json).

packages/core/src/thymian.ts
atennert: Fixed in 1250b379 by removing the sole-entry fallback from `resolveThymianFormatForRun` — resolution is now strictly by the run's own `thymianFormatVersion`, so a merged single-entry union can never attribute a foreign graph to a run. The forgetting-producer guard it provided moves to assembly, where provenance is actually known: `finalizeWorkflow` completes missing versions from the workflow format on single-workflow runs (lint/test/analyze), and `reportConvert` completes them per fragment when the fragment's own source map has exactly one entry. A run whose provenance genuinely can't be established degrades to the raw `format:<elementId>` text instead of rendering against the wrong API. Covered by new core tests plus the flipped resolve-location tests.

packages/core/src/thymian.ts
atennert: Fixed in 1250b379: assembled runs are de-duplicated on `runId` in `reportConvert` (first occurrence kept, a warning logged when duplicates drop), so `merge report.json copy.json` yields one run. Covered by a core unit test and an e2e test that literally `cp`s the fixture; the tutorial's caution block now states the semantics — same run collapses by identity, findings across *different* runs still union by design.

astro-docs/src/content/docs/tutorials/report-merge.mdx
atennert: Resolved in 1250b379 by the CLI-only decision: the config-schema description change is reverted (both twins back to main's convert-only wording), so `references/configuration.mdx` lines 37/127 are consistent again as they stand, and the tutorial's config section now documents that merge ignores `reports`/`specifications`. With no schema edit left in the PR, the un-regenerated `configuration-schema.md` no longer hides anything — its missing `reports` row predates this branch (worth a separate issue for the generator gap if you want one).

packages/core/src/thymian.ts
```diff
@@ -687,12 +774,57 @@ export class Thymian {
...
+          writable: true,
+          configurable: true,
+        });
+      }
+      thymianFormat = copied;
+
+      if (format) {
+        // Keep an already-present entry (same hash ⇒ same graph).
+        thymianFormat[format.attributes.hash] ??= format;
+      }
```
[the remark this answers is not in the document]
atennert: Fixed in cfb042ae: the workflow-format insertion now uses `Object.hasOwn` for the presence check and `Object.defineProperty` for the write, matching the hardened copies above — no inherited-member reads, no `__proto__` setter exposure.

packages/thymian/src/commands/report/merge.ts
atennert: Scope correction in cfb042ae (per follow-up discussion): only *report* inputs are CLI-only — specifications keep the normal `--spec`-over-config chain for merge, like every other command. The decision is now documented as ADR-0019 (`docs/arc42/adr/0019-report-inputs-are-cli-only-for-merge-and-diff.md`), which also settles input sourcing for the upcoming `report diff` (#502).

packages/thymian/src/commands/report/merge.ts
atennert: Scope correction in cfb042ae: merge does read config `specifications` again (only the `reports` key is CLI-only, ADR-0019). Your points 2 and 3 stay fixed structurally — the spec joins the merged map only when a run converted against it, and the sole-entry fallback is gone. Point 1 remains as documented ADR-0019 trade-off: a configured spec is loaded eagerly even when no input needs it (input types are plugin-claimed, so core cannot know upfront whether a claimant will use the format), so a broken configured spec fails the merge.

e2e-tests/src/cli-report-merge.test.ts
atennert: Follow-up in cfb042ae: since merge resolves config `specifications` again (ADR-0019 pins only report inputs to the CLI), the fixture config now deliberately carries NO `specifications` entry — exactly your suggested fix — so the passthrough test stays load-bearing: the only spec the pipeline can see is the format map inside converted.json.

astro-docs/src/content/docs/tutorials/report-merge.mdx
atennert: Follow-up in cfb042ae: the decision is now an ADR (ADR-0019 — report inputs CLI-only for merge and the future diff; specifications keep the config chain). The tutorial's note was re-scoped accordingly. The config `reports` docs stay convert-only, consistent with schema and reference.

packages/thymian/src/commands/report/merge.ts
atennert: Renumbered in ea013b6e: the ADR is now **ADR-0020** (`0020-report-inputs-are-cli-only-for-merge-and-diff.md`) — 0019 is already taken by another PR. All code/test/fixture references updated; earlier replies on this PR saying ADR-0019 mean this ADR.

packages/core/src/thymian.ts
```diff
@@ -617,8 +685,27 @@ export class Thymian {
...
+        (run) => run.thymianFormatVersion === workflowFormat.attributes.hash,
+      );
+
     return {
-      report: this.finalizeWorkflow(toolRuns, format?.export()),
+      report: this.finalizeWorkflow(
+        uniqueToolRuns,
+        workflowFormatUsed ? workflowFormat : undefined,
+        Object.keys(fragmentFormats).length > 0 ? fragmentFormats : undefined,
+        // Unclaimed inputs are a usage error the caller enforces (ADR-0017);
```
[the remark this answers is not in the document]
atennert: Addressed in 8a73fc57 as a contract clarification, deliberately not a completion: at merge assembly an untagged run is indistinguishable from one converted *without* the format — claimants legitimately convert format-free (spectral without `--spec`), and persisted runs may reference a *different* graph — so completing from `--spec` would re-introduce exactly the misattribution the removed sole-entry fallback had. The "completed at assembly" contract applies where provenance is knowable: single-workflow runs (lint/test/analyze) and single-entry fragment maps. For `core.report.convert`, tagging is the claimant's obligation — a claimant that converts against the handed format must set `thymianFormatVersion = format.attributes.hash` (both in-tree claimants do; spectral at index.ts:70). 8a73fc57 makes that contract explicit in the listener contract doc and at the `workflowFormatUsed` gate, and pins the degradation with a new core test: an untagged run stays versionless (renders raw `format:<elementId>` text) and the unused graph is withheld.

docs/arc42/adr/0020-report-inputs-are-cli-only-for-merge-and-diff.md
```diff
@@ -0,0 +1,53 @@
+# ADR-0020: Report inputs for `report merge` and `report diff` come from CLI arguments only
+
+| Status   | Date       | Supersedes | Superseded by |
+| -------- | ---------- | ---------- | ------------- |
+| Accepted | 2026-08-24 | —          | —             |
+
```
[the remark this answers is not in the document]
atennert: Fixed in 8a73fc57: ADR-0020 added to the arc42 ADR index (Accepted, 2026-08-24; no related quality requirements — the input-sourcing decision doesn't map to a 10.2 scenario, so the column carries — like ADR-0015/0017's kin). The 0019 gap is intentional — that number is claimed by another open PR.

packages/plugin-reporter/src/formatters/csv.ts
```diff
@@ -53,44 +65,50 @@ export class CsvFormatter implements Formatter<CsvFormatterOptions> {
...
+      stream.once('error', onError);
 
-      this.stream.on('ready', () => {
-        this.stream.removeListener('error', onError);
-        this.stream.on('error', (err) => {
+      stream.on('ready', () => {
+        stream.removeListener('error', onError);
+        stream.on('error', (err) => {
           this.logger.error(
             `Failed to write CSV report to ${this.options.path}: ${err.message}`,
```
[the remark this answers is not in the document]
atennert: Fixed in 60e1f64b: the first post-'ready' stream error is now recorded (`streamError`) and thrown from the next `report()`/`flush()` — the header write's failure has no callback to reject through, and `flush()`'s `end` callback previously resolved unconditionally, so this closes the exit-clean-beside-a-truncated-CSV path. Pinned by two /dev/full regression tests (opens fine, every write ENOSPCs — the exact post-open class; Linux-guarded with skipIf): the header-only case fails without the fix. On the line-110 remark: Node does deliver fs write failures to the `write()` callback (verified — that test case rejects even pre-fix), but `report()` now also throws a recorded `streamError` up front, so the destroyed-stream path is covered deterministically either way.

packages/core/src/thymian.ts
```diff
@@ -617,8 +685,34 @@ export class Thymian {
...
+        // a partial assembly must not reach `core.report` listeners, or the
+        // file formatters would persist a truncated merge before the caller
+        // gets to fail the command.
+        { emitReport: unclaimed.length === 0 },
```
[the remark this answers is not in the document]
atennert: Fixed in dd874e96 — confirmed real: `finalizeWorkflow` inferred "single workflow" from the absence of `additionalFormats`, but a mixed merge reaches it in exactly that shape (one run tagged with the --spec hash makes the workflow format join, while no fragment carries a map), and a versionless passthrough run then inherited the foreign spec hash. The backfill is now gated on an explicit `completeRunVersions` opt-in that only the lint/test/analyze callers pass; `reportConvert` never opts in, so merge-side completion remains exclusively the provenance-safe per-fragment single-entry case. Pinned by a new mixed-merge core test (spectral run tagged with the spec hash + versionless thymian passthrough): the passthrough stays versionless and the map carries only the spec hash — fails without the fix. Verified: core 346/346, full e2e 86/86 locally.

packages/plugin-reporter/src/formatters/csv.ts
```diff
@@ -32,17 +32,48 @@ function executionLabel(
...
+  async flush(): Promise<string | undefined> {
+    // No report ever arrived: like the json/markdown formatters, leave no
+    // file behind — core withholds the report emission on a failed run
+    // (e.g. an unclaimed report input), and a header-only artifact on disk
+    // would read as a produced result.
+    if (!this.streamPromise) {
+      return undefined;
+    }
+
+    const stream = await this.streamPromise;
```
matthyk, line 59: **Open failure is re-thrown from `flush()` and stalls `core.close`.**

`report()` has already awaited this promise, so if `openStream()` rejected — unwritable path (EACCES), or the path is an existing directory (EISDIR) — `flush()` re-throws that same rejection here, before the `streamError` guard below is ever consulted.

That lands in `packages/plugin-reporter/src/index.ts:129-131`:

```ts
emitter.onAction('core.close', async (_event, ctx) => {
  await flushReporters();
  ctx.reply();            // <- never reached
});
```

Two consequences:

1. `emitAction('core.close')` blocks until the idle timeout instead of failing fast — the CLI hangs rather than reporting a bad `--csv` path.
2. `flushReporters` uses `Promise.all(reporters.map(r => r.flush()))`, which rejects on the first error, so the markdown/JSON flushes are no longer awaited — their `writeFile` can be cut short by process exit.

The round-9 `streamError` handling covers post-open failures (ENOSPC, mount going away) correctly, since that check happens *after* the await. It's specifically the open rejection that short-circuits past it.

Worth noting the failure mode also moved in this PR: the open used to happen in an awaited `init()` (fail fast, before any work), and is now lazy on first report. That's a reasonable trade for streaming rows, but the close path should still surface it as an error rather than a hang.

Suggestion: `Promise.allSettled` in `flushReporters` so sibling formatters still complete, throw an aggregated error afterwards, and `ctx.reply()` in a `finally` so the action always completes.

packages/core/src/thymian.ts
```diff
@@ -602,9 +618,67 @@ export class Thymian {
...
+        return fragment.run;
+      });
     });
 
+    // Run identity is `runId`, not the input path it arrived under: the same
+    // persisted run reaching the merge twice (a copied file, two exports of
+    // one report) must not yield duplicate `runId`s in the assembled report —
+    // downstream consumers join and de-duplicate on that id.
+    const seenRunIds = new Set<string>();
+    const uniqueToolRuns = toolRuns.filter((run) => {
```
matthyk: **Warn-and-drop here can silently change the exit code.**

`runId` is `randomUUID()` (`packages/core/src/report/report-builder.ts:53`), so two runs sharing an id means one of two things:

- **the same run arrived twice** (a report embedded in two inputs) — exactly what this dedup is for, and correct;
- **a copied or hand-edited report file**, where the two runs are genuinely different.

The second case is silently resolved as the first. The dropped run's `executions` never reach the merged report, so `classifyReport` doesn't see its `failed` executions and a merge that should exit 1 can exit 0 — a green CI gate with no error, just a `logger.warn`, and no way to opt out.

The two cases are cheap to distinguish: if the duplicate is deep-equal to the run already seen, drop it silently (it really is the same run); if it differs, that's malformed input and deserves the same treatment as the other input-level errors in this PR — fail naming both inputs, rather than picking a winner by arrival order.

Given UUID generation this isn't a happy-path blocker, but "drops findings and can flip the exit code, warn-only" is a rough failure mode to leave in the command whose whole job is combining inputs from multiple sources.

packages/thymian/src/commands/report/merge.ts
```diff
@@ -0,0 +1,33 @@
...
+  // silently widen or alter an explicit merge request. Specifications keep
+  // the normal resolution chain — Step C has already folded --spec over the
+  // config's `specifications` by the time this runs.
+  protected override resolveReportInputs(): {
+    reports: ReportInput[];
+    specification: SpecificationInput[];
+  } {
+    return {
+      reports: this.flags.report ?? [],
+      specification: this.thymianConfig.specifications ?? [],
```
matthyk: Why we do not allow to declare specifications via the CLI?

atennert: You're right, specifications should be read from Config AND CLI.

packages/core/src/thymian.ts
atennert: Fixed in 5e59e4d6, exactly along your split: an identical duplicate (deep-equal via `util.isDeepStrictEqual`, so key order can't defeat it) still collapses — that's the copied-file case the dedup exists for — while two *different* runs under one runId now fail the merge with a ThymianBaseError naming both inputs (`from thymian:./report.json and thymian:./edited-copy.json`), exit 2 like the other input-level errors. The warn-and-drop is gone, so no execution can silently vanish from `classifyReport`. Comparison happens after per-fragment version completion, so two identical inputs completed identically stay equal. Pinned by a new core test (conflicting same-id runs reject, load-bearing verified against the pre-fix source); the existing collapse tests (unit + both e2e AC 4 cases) still pass unchanged since they use identical content. Tutorial caution block updated to the conflict-fails semantics.

packages/plugin-reporter/src/formatters/csv.ts
atennert: Fixed in 5e59e4d6 — consequence 2 as you suggested: `flushReporters` now uses `Promise.allSettled`, logs any additional failures, and re-throws the first after all siblings settled, so the markdown/json writes always complete before close proceeds. On consequence 1, I measured before changing anything: the open failure already fails fast (~515ms, no idle-timeout stall). The emitter's `onAction` catches a thrown handler error and forwards it as a correlated error event, which counts toward the collect take and is re-thrown by `emitAction` — so a throwing `core.close` handler *is* a completed action, and the CLI reports the bad --csv path as an error. (The open rejection also surfaces even earlier, at `report()` time via the `core.report` listener, which is what actually fails the run.) For that reason I didn't add `ctx.reply()` in a finally: replying *and* throwing would double-emit for one listener and cut the collect short, and throw-don't-reply is the documented action-error contract. Both behaviors are pinned by a new regression test (`reporter-close.test.ts`): csv path pointing at a directory + json sibling → run rejects with EISDIR in <5s and the sibling json report lands complete on disk.

packages/thymian/src/commands/report/merge.ts
atennert: Good news: it already works exactly as ADR-0020 states — `--spec` is a base flag, and Step C (BaseCliRunCommand) folds it over the config's `specifications` before any command's run(), so this line carries BOTH sources with flags winning. Three merge unit tests prove it (mixed --report+--spec forwarding; config specs used; --spec overriding config specs), and the command's own help example shows `--spec openapi:./api.yaml`. The line just didn't say so — 5e59e4d6 adds the inline comment at exactly this spot so the next reader doesn't have to trust the method-level docstring.

packages/common-cli/src/base-report-assembly-command.ts
```diff
@@ -0,0 +1,69 @@
...
+ * `thymian report merge`): resolve the report and spec inputs (the one place
+ * the two commands deliberately differ — see {@link resolveReportInputs}),
+ * fail usage-style on an empty input set, run the core convert workflow,
+ * enforce input claims (ADR-0017), and hand the outcome to the standard
+ * renderer/exit-code path. Extracted so behavior and user-visible wording
+ * can't silently drift between the two commands (#507 review).
+ */
+export abstract class BaseReportAssemblyCommand<
+  T extends typeof Command,
+> extends BaseCliRunCommand<T> {
```
matthyk: Are there other ways to share functions besides inheritance? I think we should not follow the pattern of creating a new subclass for each functionality which is shared by multiple commands

packages/core/src/load-thymian-report.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Thymian versions stay readable as long as the core shape holds. The file
+ * boundary itself (path resolution, BOM tolerance, read/parse error wording)
+ * is the shared `readTypedInputJson` contract all claimants use.
+ *
+ * @param inputLabel the offending input's identity (`type:location`), used in
+ *   every error message so failures trace back to the CLI input.
+ * @throws {ThymianBaseError} if the file is unreadable, not valid JSON, not a
+ *   report object/array, an empty array, or an entry fails `reportSchema`.
+ */
+export async function loadThymianReports(
```
matthyk, line 24: Why the logic for reading thymian reports lives in the formatter plugin?

packages/core-testing/src/mocks/thymian.mock.ts
```diff
@@ -0,0 +1,66 @@
...
+        }
+      );
+    });
+  }
+
+  return {
+    ...actual,
+    Thymian: MockThymian,
+  };
+}
```
matthyk, line 66: We should not create such test helpers over and over again. Place it in the core-testing package and import it from there.

packages/core/src/load-thymian-report.ts
atennert: As discussed .. to be moved into core

packages/common-cli/src/base-report-assembly-command.ts
atennert: Agreed — replaced in de4cfd1b. `BaseReportAssemblyCommand` is gone; the shared spine is now a free function `runReportAssembly()` in common-cli (`packages/common-cli/src/report-assembly.ts`), the same mechanism `lint`/`test` already use to share their run() bodies (and `enforceReportClaims`/`handleWorkflowOutcome` were already structurally-typed helpers). Both commands extend `BaseCliRunCommand` directly and keep only their deliberate divergence: input resolution (flags-over-config for convert, CLI-only for merge per ADR-0020) and the empty-input error wording. One wrinkle: `thymian` and the two base flags travel as an explicit options argument rather than on the structural command type, because `BaseCliRunCommand` declares them `protected` and TypeScript won't let a protected member satisfy an unrelated public interface. ADR-0020's enforcement sentence is updated to describe the new shape.

packages/core/src/load-thymian-report.ts
atennert: Moved in de4cfd1b: `loadThymianReports` now lives in core (`packages/core/src/load-thymian-report.ts`, beside `readTypedInputJson`) — it validates core's own `reportSchema` and carried no plugin-specific state, so it was core knowledge sitting in the plugin. plugin-reporter keeps the `thymian:` claim registration and the write side (the JSON formatter that produces these files); ADR-0017's "owns the file boundary in both directions" wording is narrowed accordingly. The loader also gained focused core unit tests (bare object / array / empty array / invalid entry index / missing file / invalid JSON).

packages/core-testing/src/mocks/thymian.mock.ts
atennert: Done in de4cfd1b: the helper moved to `@thymian/core-testing` (`src/mocks/thymian.mock.ts`), exported from the barrel and additionally via a dedicated `@thymian/core-testing/mocks/thymian` subpath. The `vi.mock('@thymian/core', …)` factories import the subpath rather than the barrel on purpose: the barrel imports `@thymian/core` values at module scope, i.e. inside the very factory that is mocking that module. convert.test.ts/merge.test.ts now consume it from there and `packages/thymian` gained the devDependency. Note: `serve.test.ts` and `config-resolution.integration.test.ts` still carry their own pre-existing inline `MockThymian` variants — left untouched here to keep the diff scoped to files this PR owns; happy to migrate them onto the shared mock in a follow-up.

matthyk (approved):

PR #364, opened 2026-08-15
fix(common-cli): stop plugin-not-found suggestion from dumping stack traces

matthyk:
## Problem
Running e.g. `thymian explain my-rule`, accepting the "did you mean explain rule?" suggestion, then hitting the missing required-arg error prints a large stack trace plus a spurious `Command explain:my-rule not found.` — none of which appears when running `thymian explain rule` directly.

## Root cause
`@oclif/plugin-not-found` re-runs the *suggested* command from inside its hook. When that command fails a `showHelp` parse error, oclif core's top-level `handle()` renders it with `showHelp(process.argv.slice(2))` — but `process.argv` still holds the *original* unknown command, so `showHelp` throws and `handle()`'s emergency catch dumps two raw stack traces.

## Fix
In `BaseCliRunCommand.catch()`, repoint `process.argv` at the command that actually ran (`this.id`) before the error propagates, so `handle()` renders it like a direct invocation (pretty message + usage, no stack trace). Scoped to `showHelp` errors on the fatal-exit path; nothing downstream reads argv after.

## Testing
- `nx build common-cli` and `nx test common-cli` (271 tests) pass.
- `nx lint common-cli` passes.
- Reproduced end-to-end and confirmed the upstream defect with a fixture CLI + failing-vs-clean tests: oclif/plugin-not-found#1132.

### DIFF
--- a/packages/common-cli/src/base-cli-run-command.ts
+++ b/packages/common-cli/src/base-cli-run-command.ts
@@ -32,6 +32,7 @@ import { getConfig } from './get-config.js';
 import type { ThymianSpecSearchResult } from './hooks/spec-search-hook.js';
 import type { ThymianTrafficSearchResult } from './hooks/traffic-search-hook.js';
 import type { ThymianConfig } from './thymian-config.js';
+import { writeErrorRecord } from './write-error-record.js';
 
 const require = createRequire(import.meta.url);
 
@@ -319,29 +320,7 @@ export abstract class BaseCliRunCommand<
   }
 
   protected override async catch(err: CommandError): Promise<void> {
-    await this.feedback?.error();
-    const versionDetails = this.config.versionDetails;
-
-    const pluginVersions = Object.entries(versionDetails.pluginVersions ?? {})
-      .filter(([name]) => !name.startsWith('@oclif'))
-      .map(([name, version]) => ({ name, version: version.version }));
-
-    await this.errorCache?.write({
-      name: err.name,
-      message: err.message,
-      commandName: this.id ?? 'unknown command',
-      timestamp: Date.now(),
-      cause: err.cause,
-      stack: err.stack,
-      argv: process.argv,
-      version: {
-        architecture: versionDetails.architecture,
-        cliVersion: versionDetails.cliVersion,
-        nodeVersion: versionDetails.nodeVersion,
-        osVersion: versionDetails.osVersion,
-      },
-      pluginVersions,
-    });
+    await writeErrorRecord(this, this.feedback, this.errorCache, err);
 
     if (err instanceof ThymianBaseError) {
       const cliError = new CLIError(err.message, {
@@ -358,25 +337,6 @@ export abstract class BaseCliRunCommand<
       }
 
       return super.catch(cliError);
-    }
-
-    // When this command was reached via @oclif/plugin-not-found's "did you
-    // mean …?" suggestion, process.argv still holds the original unknown
-    // command (e.g. `explain my-rule`). oclif core's top-level handle()
-    // renders parse errors with showHelp(process.argv.slice(2)); for the
-    // unknown id that throws "command not found" and the emergency catch dumps
-    // raw stack traces. Repoint process.argv at the command that actually ran
-    // so showHelp resolves it — matching the direct-invocation path. This must
-    // happen last, right before handle() (the only consumer): earlier reads —
-    // e.g. the error-cache write above — must still see the argv the user
-    // actually typed, so the diagnostic record preserves the real invocation.
-    // Upstream bug: https://github.com/oclif/plugin-not-found/issues/1132
-    if ((err as { showHelp?: boolean }).showHelp && this.id) {
-      process.argv = [
-        process.argv[0]!,
-        process.argv[1]!,
-        ...this.id.split(':'),
-      ];
     }
 
     return super.catch(err);
--- a/packages/common-cli/src/base-cli-run-command.ts
+++ b/packages/common-cli/src/base-cli-run-command.ts
@@ -320,7 +320,7 @@ export abstract class BaseCliRunCommand<
   }
 
   protected override async catch(err: CommandError): Promise<void> {
-    await writeErrorRecord(this, this.feedback, this.errorCache, err);
+    await writeErrorRecord(this, err, this.feedback, this.errorCache);
 
     if (err instanceof ThymianBaseError) {
       const cliError = new CLIError(err.message, {
--- a/packages/common-cli/src/base-cli-run-command.ts
+++ b/packages/common-cli/src/base-cli-run-command.ts
@@ -319,6 +319,23 @@ export abstract class BaseCliRunCommand<
   }
 
   protected override async catch(err: CommandError): Promise<void> {
+    // When this command was reached via @oclif/plugin-not-found's "did you
+    // mean …?" suggestion, process.argv still holds the original unknown
+    // command (e.g. `explain my-rule`). oclif core's top-level handle()
+    // renders parse errors with showHelp(process.argv.slice(2)); for the
+    // unknown id that throws "command not found" and the emergency catch dumps
+    // raw stack traces. Repoint process.argv at the command that actually ran
+    // so showHelp resolves it — matching the direct-invocation path. Safe: this
+    // is the fatal-exit path and nothing downstream reads argv afterwards.
+    // Upstream bug: https://github.com/oclif/plugin-not-found/issues/1132
+    if ((err as { showHelp?: boolean }).showHelp && this.id) {
+      process.argv = [
+        process.argv[0]!,
+        process.argv[1]!,
+        ...this.id.split(':'),
+      ];
+    }
+
     await this.feedback?.error();
     const versionDetails = this.config.versionDetails;
 
--- a/packages/common-cli/src/base-cli-run-command.ts
+++ b/packages/common-cli/src/base-cli-run-command.ts
@@ -319,23 +319,6 @@ export abstract class BaseCliRunCommand<
   }
 
   protected override async catch(err: CommandError): Promise<void> {
-    // When this command was reached via @oclif/plugin-not-found's "did you
-    // mean …?" suggestion, process.argv still holds the original unknown
-    // command (e.g. `explain my-rule`). oclif core's top-level handle()
-    // renders parse errors with showHelp(process.argv.slice(2)); for the
-    // unknown id that throws "command not found" and the emergency catch dumps
-    // raw stack traces. Repoint process.argv at the command that actually ran
-    // so showHelp resolves it — matching the direct-invocation path. Safe: this
-    // is the fatal-exit path and nothing downstream reads argv afterwards.
-    // Upstream bug: https://github.com/oclif/plugin-not-found/issues/1132
-    if ((err as { showHelp?: boolean }).showHelp && this.id) {
-      process.argv = [
-        process.argv[0]!,
-        process.argv[1]!,
-        ...this.id.split(':'),
-      ];
-    }
-
     await this.feedback?.error();
     const versionDetails = this.config.versionDetails;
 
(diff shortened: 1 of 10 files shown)


packages/common-cli/src/base-cli-run-command.ts
```diff
@@ -360,6 +360,25 @@ export abstract class BaseCliRunCommand<
...
+    // so showHelp resolves it — matching the direct-invocation path. This must
+    // happen last, right before handle() (the only consumer): earlier reads —
+    // e.g. the error-cache write above — must still see the argv the user
+    // actually typed, so the diagnostic record preserves the real invocation.
+    // Upstream bug: https://github.com/oclif/plugin-not-found/issues/1132
+    if ((err as { showHelp?: boolean }).showHelp && this.id) {
+      process.argv = [
+        process.argv[0]!,
+        process.argv[1]!,
+        ...this.id.split(':'),
```
BaggersIO: **`split(':')` hard-codes the topic separator — breaks for consumers using oclif's default.**

`@thymian/common-cli` is published and `BaseCliRunCommand` is exported from `src/index.ts`. A consumer CLI that omits `topicSeparator` gets oclif's default `':'` (`@oclif/core/lib/config/config.js:130`), not `' '`.

When the separator is `':'`, `Help.showHelp` skips `standardizeIDFromArgv` entirely:

```js
// @oclif/core/lib/help/index.js:206
if (this.config.topicSeparator !== ':') { ... }
```

So `getHelpSubject` returns only the first element — `'explain'` instead of `'explain:rule'` — `findCommand` fails, and you get either the wrong (topic-level) help or `Command explain not found.` thrown. That's the same double stack trace this PR exists to eliminate, just reached by a different route.

oclif exports the separator-aware helper:

```ts
import { toConfiguredId } from '@oclif/core'
...toConfiguredId(this.id, this.config).split(' ')
```

Correct under both separators.

packages/common-cli/src/base-cli-run-command.ts
```diff
@@ -360,6 +360,25 @@ export abstract class BaseCliRunCommand<
...
+    // command (e.g. `explain my-rule`). oclif core's top-level handle()
+    // renders parse errors with showHelp(process.argv.slice(2)); for the
+    // unknown id that throws "command not found" and the emergency catch dumps
+    // raw stack traces. Repoint process.argv at the command that actually ran
+    // so showHelp resolves it — matching the direct-invocation path. This must
+    // happen last, right before handle() (the only consumer): earlier reads —
+    // e.g. the error-cache write above — must still see the argv the user
+    // actually typed, so the diagnostic record preserves the real invocation.
+    // Upstream bug: https://github.com/oclif/plugin-not-found/issues/1132
+    if ((err as { showHelp?: boolean }).showHelp && this.id) {
```
BaggersIO: **Guard is broader than oclif's own `showHelp` gate, so argv gets clobbered on paths where `handle()` never runs.**

`handle()` only calls `showHelp` when all three hold (`@oclif/core/lib/errors/handle.js:41`):

```js
if (err.showHelp && err.parse?.input?.argv && config) { ... }
```

Two gaps:

**(a) json-enabled commands.** `packages/plugin-sampler/src/cli/commands/sampler/validate.ts:32` sets `enableJsonFlag = true`. So `thymian sampler validate --json --bogus` throws `NonExistentFlagsError` (`showHelp=true`), this block rewrites `process.argv`, then `super.catch()` sees `jsonEnabled()`, logs JSON, and **does not rethrow**. `handle()` is never reached — nothing consumes the rewrite — but `_run()` returns normally and oclif goes on to run the `postrun`/`finally` hooks against a corrupted global argv.

**(b)** Errors carrying a truthy `showHelp` but no `parse.input.argv` mutate argv for nothing.

Mirroring handle()'s own condition fixes both:

```ts
if (!this.jsonEnabled() && (err as {showHelp?: boolean}).showHelp && (err as any).parse?.input?.argv && this.id) {
```

packages/common-cli/src/base-cli-run-command.ts
```diff
@@ -360,6 +360,25 @@ export abstract class BaseCliRunCommand<
...
+      process.argv = [
```
BaggersIO: **`process.argv` is mutated globally and never restored — and the comment's "the only consumer" claim doesn't hold.**

The comment above says `handle()` is the only consumer after this point. Three places read it afterward:

1. **Hooks.** After `catch()` returns, oclif still runs `config.runHook('postrun', ...)` on the json path, and *always* `config.runHook('finally', ...)` from `main.run`'s finally block. Third-party hook code now sees a command line the user never typed.
2. **In-process reuse.** This is a published library, and both `oclif.execute` (`execute.js:55`) and `Command.run` (`command.js:156`) default to `process.argv.slice(2)` when no args are passed. A second invocation in the same process runs the previously *suggested* command.
3. **This method itself, on a second error.** A later error in the same process writes the rewritten argv into `last_error.json` at line 336 — which defeats the ordering intent this PR's own comment is at pains to preserve.

Either snapshot and restore:

```ts
const originalArgv = process.argv
try { return await super.catch(err) } finally { process.argv = originalArgv }
```

or better, don't mutate a global at all — see my note in the review body about doing this in a `prerun` hook.

packages/common-cli/src/base-cli-run-command.ts
```diff
@@ -360,6 +360,25 @@ export abstract class BaseCliRunCommand<
...
+        process.argv[0]!,
```
BaggersIO: Minor: the non-null assertions can splice `undefined` into a `string[]` rather than failing loudly. Under `node --eval` or some embedding contexts `process.argv` has fewer than two entries, so `process.argv[1]!` is `undefined` and you end up with `[execPath, undefined, 'explain', 'rule']` still typed `string[]` — any later `argv.join(' ')` yields `'undefined'`, any `argv[1].startsWith(...)` throws.

```ts
process.argv = [...process.argv.slice(0, 2), ...toConfiguredId(this.id, this.config).split(' ')]
```

Shorter and assertion-free.

BaggersIO (changes requested):
Reviewed the argv-repoint fix. The diagnosis is right and the comment explaining *why* the repoint has to happen after the error-cache write is genuinely good — that ordering constraint is subtle and would have been easy to get wrong.

But I think the fix is applied at the wrong altitude, and that has already cost coverage. Details inline; three items below don't land on diff lines.

---

### 1. `ThymianBaseCommand.catch()` has the identical bug and is untouched

`packages/common-cli/src/thymian-base-command.ts:74` is the same error path, without the repoint. So the bug this PR fixes still reproduces for every command extending that base class:

```
thymian generate confi --bogus
  -> not-found suggests `generate config`, user accepts
  -> runCommand('generate:config', ['--bogus'])
  -> NonExistentFlagsError (showHelp=true)
  -> ThymianBaseCommand.catch() rethrows, no repoint
  -> handle() calls showHelp(['generate','confi','--bogus'])
  -> 'Command generate:confi not found.' -> two raw stack traces
```

Affects `generate/rule.ts`, `generate/config.ts`, and `feedback.ts`. Worth confirming by hand before merge.

### 2. This belongs in a hook, not per-base-class

The defect is in oclif's top-level `handle()`, shared by *every* command — including `@oclif/plugin-help`/`-version` and any third-party plugin command that doesn't extend a Thymian base class. Patching one `catch()` means copy-pasting into every base class, which item 1 shows already didn't happen.

A `prerun` hook (oclif passes `{argv, Command}` before the command runs) or a wrapper around `oclif.execute` in `packages/common-cli/src/oclif.ts` fixes it once, everywhere. `packages/common-cli/src/hooks/` already exists for this.

Related: `BaseCliRunCommand.catch()` lines 322-344 and `ThymianBaseCommand.catch()` lines 51-73 are byte-for-byte duplicates of the error-cache/feedback logic. Extracting a shared `writeErrorRecord(command, err)` (or having one extend the other) would have made item 1 structurally impossible.

### 3. No regression test

The fix depends entirely on undocumented `@oclif/core` 4.13.0 internals: `err.showHelp`, `handle()` reading `process.argv.slice(2)`, and `Help.showHelp`'s argv standardization. Any oclif bump — or oclif/plugin-not-found#1132 actually landing — can silently turn this block into a no-op or make it actively wrong, and nothing in `packages/common-cli/test/` or `e2e-tests/` would catch it. A unit test asserting `process.argv` after `catch()` on a synthetic `{showHelp: true}` error would pin the contract without needing a TTY.

### 4. Non-blocking, pre-existing

`base-cli-run-command.ts:336` writes the full `process.argv` verbatim into `last_error.json`. `thymian lint -o 'some-plugin.auth.token=ghp_...'` failing puts that token in a plaintext cache file at default permissions. Not introduced here, but since this PR is already reasoning carefully about exactly which argv that record should hold, redacting flag values is a natural companion change.

---

Items 1 and the two inline correctness comments are the blocking ones for me. Happy to re-review quickly.

matthyk:
Thanks @BaggersIO — this was the right call. Reworked in 312c180e to fix the defect once at the shared `execute`/`handle` boundary instead of per-`catch()`.

**1. `ThymianBaseCommand` had the same bug — fixed.** The repoint no longer lives in any `catch()`. It's now in `oclif.ts`'s `execute` wrapper, the single choke point every command flows through, so `generate config`/`generate rule`/`feedback` (and `plugin-help`/`-version`/third-party) are all covered. Verified by hand via a pty:
- `explain my-rule` → accept → clean `Missing 1 required arg` + usage, no stack trace.
- `generate confi --bogus` → accept → clean `Nonexistent flag: --bogus` + usage, no stack trace, no "not found".

**2. Altitude + duplication.** Fix moved to the `execute`/`handle` boundary, keyed on the resolved command id recovered from the error (`err.parse.input.context.id`) rather than mutating per command. I went with the wrapper over a `prerun` hook deliberately: `prerun` fires *before* the command's `catch()` writes the error cache, so it would have re-clobbered the diagnostic argv we fixed last round; the boundary runs *after* `catch()`, so the record stays correct. Also extracted the byte-for-byte error-record logic into a shared `writeErrorRecord()` used by both base classes, so the two `catch()` bodies can't drift again.

**3. Regression test.** Two, both TTY-free:
- unit test on the pure `argvForHelpError()` helper;
- an e2e test that drives `oclif.execute` against a fixture CLI and asserts clean help / no stack trace — it pins the oclif internals we depend on (`err.parse.input.context.id`, `handle()` reading `process.argv`). Confirmed it fails without the fix.

**Copilot A (dropped argv tail):** intentional and harmless — `handle()` only uses argv to resolve *which* command's help to show; the fixed direct + suggestion paths both render identically. (Their `slice(2 + idParts.length)` suggestion also mis-aligns when the typo and resolved id differ in token count.)

**Copilot B (global argv mutation):** the wrapper now restores `process.argv` in a `finally`, so nothing leaks past `handle()`.

**4. Secrets in `last_error.json`:** agreed it's pre-existing and out of scope here — tracked in thymianofficial/thymian-internal#573.

Happy to re-review whenever.

packages/common-cli/src/write-error-record.ts
```diff
@@ -0,0 +1,51 @@
...
+ * they're `protected` on the command instances — the caller (inside the
+ * command class) reads its own fields and hands them over.
+ *
+ * `process.argv` is recorded verbatim on purpose: it must reflect what the user
+ * actually typed. This runs before any argv repointing at the execute/handle
+ * boundary, so the record is never contaminated by that fix.
+ */
+export async function writeErrorRecord(
+  command: Command,
+  feedback: Feedback | undefined,
```
chaosmonster: nitpicking: optionals should be at the end of the parameters, otherwise they always need to be explicit undefineds

packages/common-cli/src/repoint-help-argv.ts
```diff
@@ -0,0 +1,42 @@
...
+ * applies (not a `showHelp` error, or no resolved id available), in which case
+ * callers should leave `process.argv` untouched.
+ *
+ * Pure and side-effect-free so it can be unit-tested without a TTY.
+ */
+export function argvForHelpError(
+  err: unknown,
+  argv: string[],
+): string[] | undefined {
+  const e = err as {
```
chaosmonster: This feels weird - why not a guard or just cast it?

BaggersIO (approved):

PR #367, opened 2026-08-17
feat(repo): shared HTTP-field foundation — @thymian/common-http-fields (Epic 588)

BaggersIO:
> **Draft — epic-bundled PR.** This branch carries **Epic 588** (`thymian-internal#588`,
> "Shared HTTP-field foundation"). Story 588.1 is complete below; **588.2** (normalized header
> view) and **588.3** (Structured-Fields registry + non-SF guard) will land as further commits on
> this same branch. Ready for review when the epic is done.

## Story 588.1 — establish the `common-*` dependency boundary and scaffold the package

Refs `thymian-internal#593`. Two commits:

| Commit | Scope |
|---|---|
| `feat(repo): create the scope:rules module boundary` | boundary constraint + workspace registration + doc sync |
| `feat(common-http-fields): scaffold the shared HTTP-field package` | the new package + lockfile |

### The constraint had to be **created**, not relaxed

Before this PR, `scope:rules` appeared *only* as a permitted **target** of `scope:cli`. There was
no source constraint on rule packages at all — so "`rules-*` depend only on `@thymian/core`" was
documentation, never lint. The pre-change `depConstraintsProduction` had source entries for
`scope:cli`, `scope:core` and `scope:plugin`, and none for `scope:rules`.

### Scope-tag decision (AC 4): `scope:core`

`common-http-fields` is tagged `scope:core` rather than introducing a new `scope:common` dimension
value. `core-testing` is already `scope:core` and `common-cli` is `scope:cli` — i.e. the existing
convention tags `common-*` packages by **domain**, not by a "common" scope. That made this a
one-entry change instead of three (`scope:rules` source, `scope:common` source, plus admission to
`scope:cli`'s allow-list), and `scope:core`'s own core-only constraint already stops a support
library reaching sideways. The rationale is also recorded inline above the constraint in
`eslint.config.mjs`, so it survives independently of this description.

Consequence worth stating plainly: because the tag is shared, `scope:cli`, `scope:plugin` and
`type:app` sources may also import this package. That is the accepted trade-off of Option A.

### AC 5 evidence — the constraint actually fires

Negative probe, a `rules-*` file importing a `plugin-*` package:

```
packages/rules-rfc-9110/src/__boundary-probe.ts
  2:1  error  A project tagged with "scope:rules" can only depend on libs tagged with "scope:core"  @nx/enforce-module-boundaries

✖ 4 problems (2 errors, 2 warnings)
```

Positive probe, the same file importing `@thymian/common-http-fields` — **zero**
`enforce-module-boundaries` errors, so the constraint admits the intended edge:

```
✖ 3 problems (1 error, 2 warnings)
```

(The one remaining error is `@nx/dependency-checks` for the undeclared dependency, which is
expected — a real consumer declares it. Both probes were reverted; the tree is clean.)

### Two CI blockers found in review and fixed here

- **`package-lock.json` had no entry for the new package.** npm records `packages/*` workspace
  entries in the lockfile, so `npm ci` aborted with
  `Missing: @thymian/common-http-fields@0.0.0-PLACEHOLDER from lock file` *before any step* in all
  eight CI, e2e, release and deploy-docs jobs. Nx lint/build read the manifests off disk and need
  no node resolution, so a fully green `nx run-many` cannot detect this. Same class as `7a9a6496`.
- **`.license-checker.json` did not exclude the new package.** Its `AGPL-3.0-only` license is in
  `bannedLicenses`, and the root `postinstall` runs `check:licenses` — so *every* `npm install`
  exited non-zero, plus the dedicated CI license job. This had to be fixed before the `npm install`
  that regenerates the lockfile, or that install failed too.

### Scaffold notes

- `src/index.ts` is `export {}` by design. No SF registry, no non-SF guard, no normalized header
  view, no `structured-headers` dependency — those are 588.2 / 588.3. No vendored reference data
  ever (PSL and OSHP JSON stay with their sole consumers per AD-7). No docs-generation targets: the
  package ships no commands.
- **`@thymian/core` is deliberately not yet a declared dependency.** With an empty entrypoint
  `@nx/dependency-checks` correctly rejects the unused declaration, and suppressing it would be the
  weaken-the-lint-rule anti-pattern. **588.2 must restore both** the dependency and the
  `../core/tsconfig.lib.json` project reference when it first imports core types.
- `vitest` collects `src/**/*.test.ts` in addition to the `test/**` library convention. The
  adjacent `rules-rfc-9110` colocates its tests in `src/`, which `tsconfig.lib.json` excludes from
  the build — a colocated test here would otherwise be built by nothing and collected by nothing,
  while `passWithNoTests` still reported green.

### Doc sync

The `scope:rules` row in `CONTRIBUTING.md` claimed "constrained via `type`/`npm` dimensions …
depend only on `@thymian/core` in practice" — wrong in both halves now. It and the commit-scope
table (which the file itself asks to be kept in sync) and the arc42 package inventory cited by
ADR-0008 are all updated.

Both tables now key on the **`scope:core` tag, not a name prefix**: `@thymian/common-cli` is
`scope:cli` and is therefore *not* reachable from a rule set even though it is a `common-*`
package. The matching amendment in `thymian-internal/docs/project-context.md` is committed there
separately, per AD-8.

### ⚠️ Release gate — AC 10 is open

`npm run bootstrap-npm-packages` has **not** been run for `@thymian/common-http-fields` (deferred
by decision, batched across all four new `@thymian/*` packages in `#572`). This blocks
**publishing, not merging** — but `nx.json`'s `release.projects: ["packages/*"]` picks the package
up with no opt-in, so the first release after merge fails without it. No `NPM_TOKEN` /
`NODE_AUTH_TOKEN` is introduced; publishing stays on OIDC trusted publishing.

### Verification

`npm ci --dry-run` ✅ · `node scripts/check-licenses.js` ✅ · `nx lint` ✅ (`common-http-fields`,
both `rules-*`, `thymian`) · `nx test common-http-fields` ✅ · `prettier --check` ✅ on every
changed and new file.

### DIFF
--- a/.license-checker.json
+++ b/.license-checker.json
@@ -13,6 +13,7 @@
   "excludePackages": [
     "thymian",
     "@thymian/common-cli",
+    "@thymian/common-http-fields",
     "@thymian/core",
     "@thymian/core-testing",
     "@thymian/plugin-http-analyzer",
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,7 @@ When making commits, use one of the following scopes to indicate the area of the
 | cli                              | Cross-cutting CLI concerns                                                             |
 | thymian                          | The `thymian` CLI product package                                                      |
 | common-cli                       | Shared CLI base library (`@thymian/common-cli`)                                        |
+| common-http-fields               | Shared HTTP field primitives for rule sets (`@thymian/common-http-fields`)             |
 | core                             | Core framework: workflows, contracts, rule system (`@thymian/core`)                    |
 | core-testing                     | Test utilities for core (`@thymian/core-testing`)                                      |
 | plugin-http-linter               | Static lint plugin                                                                     |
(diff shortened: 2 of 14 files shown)


BaggersIO:
Closing unmerged — the epic breakdown for the security rule-sets family (#588–592) is being torn down and restarted from scratch. Nothing here is being merged.

PR #368, opened 2026-08-19
feat(core): security rule set foundations — http-fields + rules-scope boundary (Epic 643)

BaggersIO:
## Summary

Foundations for the HTTP-security rule-set delivery (Epic 643 — thymianofficial/thymian-internal#643, tracker thymianofficial/thymian-internal#572): the shared `@thymian/core/http-fields` substrate every upcoming security rule reads through, plus the missing rules-scope dependency-boundary lint. All three stories of the epic ship in this one PR; each is its own reviewed unit.

**643.1 — Normalized header view in `@thymian/core/http-fields`** (thymianofficial/thymian-internal#650) ✅

- New `packages/core/src/http-fields/` module + `"./http-fields"` subpath export (string form, following the `./utils` / `./http-filter` precedent).
- Runtime view (`normalized-headers.ts`): case-insensitive lookup, same-name multiplicity preserved in order, case-variant duplicate keys merged in encounter order, `undefined` entries skipped, values never comma-split (`Set-Cookie` legally contains commas).
- Declared/lint-side adapter (`declared-headers.ts`): L1 presence + `required` always; L2 pinned-value facts only when the schema pins (`const`/`enum`/`pattern`/`default`) — a loose schema exposes no value, likely values are never inferred.

**643.2 — Natively-SF header registry + non-SF guard** (thymianofficial/thymian-internal#651) ✅

- SF grammar adopted, not built: `structured-headers@^2.0.3` (MIT, 0 deps, RFC 9651 + 8941), re-exported through `http-fields` so `rules-*` packages never import it directly.
- `sf-fields.ts`: `SfHeaderRegistry` over an injected allowlist; the canonical `NATIVELY_SF_HEADERS` table ships EMPTY and frozen by design — rows are authored later with defining-spec citations (`field-grammar/` work), never hardcoded here.
- `sf-parse.ts`: `parseSfField` refuses unregistered fields BEFORE any parsing, with a discriminated outcome so refusal is programmatically distinguishable from a grammar `ParseError` on a registered header.

**643.3 — Enforce `scope:rules → onlyDependOnLibsWithTags ['scope:core']`** (thymianofficial/thymian-internal#652) ✅

- Adds the missing source-side dep-constraint to `eslint.config.mjs` as its own commit. The violation-demo lint output is in the **Story 643.3 — AC-3 evidence** section below.

**643.4 — `getHeader` on the `http-fields` engine, legacy read helpers deprecated** (thymianofficial/thymian-internal#671) ✅

- Decided during review of this PR: core was carrying two case-insensitive header-lookup families. `getHeader` is now a shape-preserving adapter over `fromRuntimeHeaders`' semantics — a single matching key returns its value by the same reference as before, and only case-variant duplicate keys (`Set-Cookie` + `set-cookie`) go through the merge. `getHeader` and `objHasKeyIgnoreCase` carry `@deprecated` notices; the write path (`setHeader`/`deleteHeader`) and `findKeyIgnoreCase`/`getContentType` deliberately do not.
- Also fixes `httpResponseToCommonHttpResponse`, where `?.at(0)` indexed into a *string* content-type and reported `'a'` for `'application/json'`.
- Details in the **Story 643.4** section below.

## Delivery notes

- One bundled draft PR per epic; the stories stay In review until this merges.
- Gate after this epic: thymianofficial/thymian-internal#576 (rule test harness + captured-traffic factories in `@thymian/core-testing`) must merge before the first rule story of the bootstrap epic is authored.

### Story 643.3 — AC-3 evidence

Demonstration that the new `{ sourceTag: 'scope:rules', onlyDependOnLibsWithTags: ['scope:core'] }` constraint in `eslint.config.mjs` fires: temporarily added `import '@thymian/plugin-http-linter';` as the first line of `packages/rules-rfc-9110/src/index.ts` (a `scope:rules` package importing a `scope:plugin` package), ran `nx run rules-rfc-9110:lint`, captured the output verbatim, then reverted the fixture (`git checkout -- packages/rules-rfc-9110/src/index.ts`) before committing only `eslint.config.mjs`.

```

> nx run rules-rfc-9110:lint

> eslint .


/Users/petermuller/dev/thymian/thymian-internal/.worktrees/story-643-1-normalized-header-view-in-thymian-core-http-fields/packages/rules-rfc-9110/package.json
  46:3  error  The "rules-rfc-9110" project uses the following packages, but they are missing from "dependencies":
    - @thymian/plugin-http-linter  @nx/dependency-checks

/Users/petermuller/dev/thymian/thymian-internal/.worktrees/story-643-1-normalized-header-view-in-thymian-core-http-fields/packages/rules-rfc-9110/src/index.ts
  1:1  error  A project tagged with "scope:rules" can only depend on libs tagged with "scope:core"  @nx/enforce-module-boundaries
  1:1  error  Run autofix to sort these imports!                                                    simple-import-sort/imports

/Users/petermuller/dev/thymian/thymian-internal/.worktrees/story-643-1-normalized-header-view-in-thymian-core-http-fields/packages/rules-rfc-9110/src/rules/identifiers/userinfo/recipient-should-treat-userinfo-in-uri-from-untrusted-source-as-error.rule.ts
  5:8  warning  'RuleViolationLocation' is defined but never used  @typescript-eslint/no-unused-vars

/Users/petermuller/dev/thymian/thymian-internal/.worktrees/story-643-1-normalized-header-view-in-thymian-core-http-fields/packages/rules-rfc-9110/src/rules/range-requests/content-range/utils.ts
  38:10  warning  '_' is assigned a value but never used  @typescript-eslint/no-unused-vars

✖ 5 problems (3 errors, 2 warnings)
  2 errors and 0 warnings potentially fixable with the `--fix` option.

Warning: command "eslint ." exited with non-zero status code


 NX   Running target lint for project rules-rfc-9110 failed

Failed tasks:

- rules-rfc-9110:lint

Hint: run the command with --verbose for more details.

  Run duration:      2.7s
  Critical path:     2.6s (1 task)
  Recoverable time:  <1ms

  Recommendations:
    - Speed up or split the longest tasks on the critical path:
        rules-rfc-9110:lint    2.6s
```

Both expected errors are present: `@nx/enforce-module-boundaries` ("A project tagged with `scope:rules` can only depend on libs tagged with `scope:core`") and `@nx/dependency-checks` (unlisted `package.json` dependency). The fixture was reverted immediately after capture and never committed. Full `nx run-many -t lint` on the branch with the constraint in place stays green (0 errors; only the two known pre-existing `no-unused-vars` warnings in `rules-rfc-9110` remain).

### Story 643.4 — `getHeader` consolidation (thymianofficial/thymian-internal#671)

**Why it is in this PR.** Story 643.1 gave core a second case-insensitive header-lookup family beside the legacy `utils.ts` read helpers (~110 call sites across `core` and `rules-rfc-9110`). Leaving both in place with different case-folding and multiplicity semantics is the exact divergence `http-fields` exists to prevent, so the legacy read path was consolidated onto the new engine in the same epic rather than left as a known fork.

**The no-breaking-changes guarantee.** `getHeader`'s single-matching-key path returns `headers[key]` untouched — same value, same shape, same *reference* — which is what the ~110 call sites rest on (`extractMediaType` throws on an array; several `rules-rfc-9110` rules untangle `string | string[]` through local `headerToString` copies). Only when two or more case-variant keys for the queried name both hold a contributing value does it delegate the merge to `fromRuntimeHeaders(...).getAll(...)`. Characterization tests pin every non-duplicate-key input, including reference identity for array and empty-array values and the omitted-argument default.

**One deliberate refinement over the story's AC 2.** The AC reads "2+ matches → return the merged array". Taken literally, `{ 'Content-Type': [], 'content-type': 'text/html' }` counts as two matches and yields a one-element array, which makes `getContentType` throw `Multiple content-type headers found.` for a header with exactly one real value. The implementation instead filters the matches down to those that actually contribute a value — using the same absence test `fromRuntimeHeaders` applies internally, where `null`/`undefined`/`[]` are absent — and only merges when two or more remain. AC 2's own worked example (`{ 'Set-Cookie': ['a=1','b=2'], 'set-cookie': 'c=3' }` → `['a=1','b=2','c=3']`) is unaffected.

**Note on live impact.** Case-variant duplicate keys cannot occur in captured traffic today — both ingestion paths lower-case names and fold repeats under one key (`plugin-har`'s `transformHarHeaders`; undici on the response side). This is contract hardening for hand-built records and future producers, not a live-bug fix. The `mediaType` fix is a real bug, but its known consumers are truthiness checks, so no rule verdict changes.

### Review round 5 — `declared-headers.ts` L2 pin mechanism

Addresses the three open threads on the L2 pin mechanism (@matthyk):

- **`pins: PinFact[]` replaces `pin?: PinFact`.** The single slot ranked a *constraint* above a *value*: `{ pattern: '^v\\d+$', default: 'v1' }` exposed the regex and made `'v1'` unreachable, and `{ enum, default }` lost its default the same way. Since this view is the only route a lint-context rule has to a declared value, that rule had to skip the header or re-read `parameter.schema` itself. Reporting every pin removes the precedence question instead of reordering it — no ordering of one slot is lossless for a multi-keyword schema. Array order is still defined (`const` > `enum` > `default` > `pattern`, value-carriers first), so `pins[0]` yields a value whenever the schema declares one. A loose schema yields `pins: []`. New coverage for `pattern`+`default`, `enum`+`default`, and all four at once — previously untested, since every lint-parity AC in #644–#647 pins its fixture via `const`.
- **Pin values are deep-copied.** The old copy was one level deep, so `const: { csp: { mode: 'strict' } }` handed out the inner object by reference and `enum: [{ a: { b: 1 } }]` copied only each member's top level — one rule normalising a pin in place changed what later rules and the persisted model saw. `structuredClone` now does the copying, guarded: `const`/`default` are `unknown` and a spec model can be TypeScript loaded through jiti, so a non-cloneable pin (function, class instance) falls back to the shallow copy rather than letting a `DataCloneError` escape a read-only lookup. Primitives short-circuit, which is nearly every real pin. The three nesting tests were verified red against the shallow implementation.
- **`PinKind` documents its consumer contract.** `const`/`default` carry a value to evaluate directly (`default` an assumption, not a guarantee), `enum` a candidate set where every member must conform, `pattern` a regex rather than a value at all — and `kind` never implies severity, which comes from the rule's own tier tag.

Also merged `main` into the branch, resolving the `packages/core/package.json` conflict as the union of both sides (`./http-fields` + `./user-module` exports, `structured-headers` + `jiti` dependencies).


packages/core/test/http-fields/package-exports.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import { describe, expect, it } from 'vitest';
+
+import packageJson from '../../package.json' with { type: 'json' };
+import { fromRuntimeHeaders } from '../../src/index.js';
+
+// `nx run core:build` must run before this test for the dist assertions
+// below to pass (see spec Design Notes: this replaces a package-name-import
+// test with an assertion on the exports map + built output, since no
+// existing fixture convention covers a same-package subpath import).
```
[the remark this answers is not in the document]
BaggersIO: Fixed in 75c68008: kept the exports-map + barrel re-export assertions (self-contained, no build needed) and removed the two `dist/` existence checks, which duplicated what `nx run core:build` already validates and only passed by accident of task ordering elsewhere (`nx test core` has no `dependsOn: ['build']`).

packages/core/src/http-fields/declared-headers.ts
```diff
@@ -0,0 +1,106 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const parameter = record[key]!;
+    const pin = derivePin(parameter.schema);
+
+    return {
+      present: true,
+      required: parameter.required,
+      ...(pin ? { pin } : {}),
+    };
+  };
```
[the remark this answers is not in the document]
BaggersIO: Fixed in 75c68008: `fromDeclaredHeaders` now builds a case-folded `Map<string, string>` index once per call, and each returned lookup does an O(1) `Map.get` instead of rescanning `Object.keys(record)`. First-match-wins semantics preserved (only the first occurrence per lower-cased name is recorded).

packages/core/src/http-fields/normalized-headers.ts
```diff
@@ -0,0 +1,89 @@
+/**
+ * Runtime header record shape shared by `HttpRequest`/`HttpResponse` (see
+ * `../http.ts`): header name to a single value, multiple values (same-name
+ * multiplicity), or `undefined` (treated as absent).
+ */
+export type RuntimeHeaderRecord = Record<string, string | string[] | undefined>;
```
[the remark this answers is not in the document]
BaggersIO: Fixed in 75c68008: widened `RuntimeHeaderRecord` to `Record<string, string | string[] | null | undefined>`, matching what `fromRuntimeHeaders` already does at runtime, and dropped the unsafe cast this forced in the test.

packages/core/src/http-fields/declared-headers.ts
```diff
@@ -0,0 +1,136 @@
...
+function derivePin(schema: ThymianSchema): PinFact | undefined {
+  if (schema.const !== undefined) {
+    return { kind: 'const', value: copyPinValue(schema.const) };
+  }
+
+  if (schema.enum !== undefined) {
+    return { kind: 'enum', value: schema.enum.map(copyPinValue) };
+  }
+
+  if (schema.pattern !== undefined) {
```
matthyk: `derivePin` precedence ranks a *constraint* above a *value*.

The four keywords aren't interchangeable for the consumer: `const`, `enum` and `default` carry values a lint rule can evaluate; `pattern` carries a constraint. Ranking `pattern` above `default` means the one schema shape where both appear loses the only value it pins:

```ts
schema: { pattern: '^v\\d+$', default: 'v1' }
// -> pin: { kind: 'pattern', value: '^v\\d+$' }  — 'v1' is unreachable
```

Every topic epic (thymianofficial/thymian-internal#644–#647) tells rule authors "value checks only when the declared schema pins the value via `const`/`enum`/`pattern`/`default` (L2)", so this is the only route a lint rule has to a declared value. A rule meeting that schema either skips the header or re-reads `parameter.schema` itself, which defeats the L2 abstraction.

Two things worth settling before #644 starts authoring against this:

1. Reorder to `const` > `enum` > `default` > `pattern` — value-carriers first.
2. `pin?: PinFact` is a single slot, so a multi-keyword schema drops facts whatever the order. `pins: PinFact[]` (or splitting value-pin from constraint) reports both and lets the rule decide.

Either way this is currently unexercised: every lint-parity AC in #644–#647 pins its fixture via schema `const`, so the `pattern`/`default` interaction ships with no test behind it. A two-keyword case would be cheap to add here.

packages/core/src/http-fields/declared-headers.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/**
+ * Shallow-copies an array or object pin value so callers cannot corrupt the
+ * caller's own declared spec model -- potentially shared across every rule
+ * inspecting this header -- by mutating a returned pin. Primitives (and
+ * `null`) pass through unchanged; they're already immutable, so there is
+ * nothing to protect.
+ */
+function copyPinValue(value: unknown): unknown {
```
matthyk, line 134: The doc says callers "cannot corrupt the caller's own declared spec model", but the copy is one level deep, so nested pins stay aliased:

```ts
schema: { const: { csp: { mode: 'strict' } } }
facts.pin.value.csp.mode = 'report-only'; // mutates the shared spec model
```

Same through `enum: [{ a: { b: 1 } }]` — `.map(copyPinValue)` copies each member's top level only. The declared record is shared across every rule inspecting that header, so one rule normalising a pin in place changes what later rules and the persisted model see. `structuredClone` is available on the Volta-pinned Node 22 and would make the comment true. The test only covers the flat `enum: ['a', 'b']` case.

packages/core/src/http-fields/declared-headers.ts
```diff
@@ -0,0 +1,136 @@
...
+ * `ThymianHttpResponse` (see `../format/nodes/http-request.node.ts` /
+ * `http-response.node.ts`): header name to its declared `Parameter`.
+ */
+export type DeclaredHeaderRecord = Record<string, Parameter>;
+
+/** Which direct schema keyword pinned the header's value. */
+export type PinKind = 'const' | 'enum' | 'pattern' | 'default';
+
+/** The pinned value and which schema keyword pinned it. */
+export interface PinFact {
```
matthyk, line 30: Worth documenting what `kind` is *for*, since severity comes from the rule's tier tag rather than the pin kind: `const`/`default` → evaluate the value directly; `enum` → every member must conform; `pattern` → you hold a regex, not a value. Without that, the first author to reach for `kind` will guess.

matthyk (commented):
Three notes on the L2 pin mechanism in `declared-headers.ts` — it is the only route a lint-context rule has to a declared value (per the "Lint asserts the DECLARED contract" guidance carried by every topic epic, thymianofficial/thymian-internal#644–#647), so its shape is worth settling before #644 starts authoring 20 cookie rules against it. Nothing here blocks the rest of the PR.

packages/core/src/http-fields/declared-headers.ts
BaggersIO: Fixed in 01a3006b — took option 2, which subsumes the reorder.

`pin?: PinFact` is now `pins: PinFact[]`, one fact per keyword the schema actually sets. Reporting all of them settles the question rather than reordering it: as you note, a single slot drops facts whatever the order, so `{ pattern, default }` had no lossless ranking available. Array order is still defined — `const` > `enum` > `default` > `pattern`, value-carriers first — so a consumer taking `pins[0]` gets a value whenever the schema declares one at all, and a loose schema yields `pins: []`.

Doing it now rather than after #644: there are no production consumers yet, and story #650 leaves the shape to the implementer ("a workable shape is …"), so the cost is 10 test assertions today versus 77 rules later.

On the untested interaction — you were right that it shipped with nothing behind it. Added `pattern`+`default` (asserting `default` comes first and `v1` stays reachable), `enum`+`default`, and all four keywords at once.

packages/core/src/http-fields/declared-headers.ts
BaggersIO: Fixed in d10a0e4c — the comment was overclaiming, exactly as described.

`copyPinValue` now deep-copies via `structuredClone`, so both of your cases hold: nested `const` and nested `enum` members are no longer aliased. Verified non-vacuous — the three new nesting tests (nested `const`, nested `enum` member, nested `default`) were run red against the shallow implementation first.

One guard on top of a bare `structuredClone`: `const`/`default` are typed `unknown`, and a spec model can be authored in TypeScript and loaded through jiti, so a pin may legally hold a function or class instance that `structuredClone` rejects with `DataCloneError`. A read-only facts lookup should not be able to throw on that, so a non-cloneable value falls back to the previous shallow copy — weaker protection for an exotic pin, never an exception. Covered by a test. Primitives short-circuit before the clone, which is nearly every real pin (a header value is a string), so the cost is not on the common path.

packages/core/src/http-fields/declared-headers.ts
BaggersIO: Documented in 0674af18, using your framing directly — `const`/`default` carry a value to evaluate directly (with `default` flagged as an assumption, not a guarantee), `enum` a candidate set where every member must conform, `pattern` a regex source string rather than a value at all.

Also stated the negative you were pointing at, since it is the guess a first author is most likely to make: `kind` says what the pin *is*, never how loudly to report it — severity comes from the rule’s own tier tag.

packages/core/test/http-fields/package-exports.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  it('adds structured-headers as a runtime dependency, not a dev dependency', () => {
+    expect(packageJson.dependencies).toHaveProperty(
+      'structured-headers',
+      '^2.0.3',
+    );
+    expect(packageJson.devDependencies).not.toHaveProperty(
+      'structured-headers',
+    );
+  });
+});
```
matthyk: I don't think we should add such tests

matthyk (commented):
Besides the one specific comment, my overall feeling about this PR:

While I can imagine the advantages of this abstraction, it feels a little bit over-engineered. Maybe this comes from not seeing it in action in some rule implementations. Besides this, it also creates another problem. When we introduce these new helpers, we provides the users different interfaces of helpers for writing custom rules. We would need a clear guidance so there is no confusion about the question when to use what. And also a clearly defined internal plan for the existing 400 rules if and when we migrate them to the new helper functions.

This is no blocker from my side, but maybe we should talk about it.

packages/core/test/http-fields/package-exports.test.ts
BaggersIO: Agreed, and it was worse than redundant — removed in 505a3550.

`@nx/dependency-checks` is already active on `packages/core`'s JSON files and `structured-headers` is not in its `ignoredDependencies`, so "used in src but missing from dependencies" is a lint error already. On top of that the test hardcoded `'^2.0.3'`, so it would have failed on the next Dependabot bump while proving nothing new.

I kept the other two assertions in the file, but tell me if you meant the whole thing — I read your anchor as the dependency block:

- the exports-map assertion is there because story #650's AC 1 wants the subpath proven by something other than a relative import. I agree it is thin: it restates `package.json` rather than proving resolution. The honest version is a package-name import through built `dist/` (the `test/rules/fixtures/**.mjs` precedent), which an earlier round removed as build-order-brittle since `nx test core` has no `dependsOn: ['build']`. So what is left satisfies the AC's letter, not its intent — happy to drop it and let `core:build` be the only guard if you prefer that trade.
- the barrel re-export check earns its keep: `http-fields` shipped missing from `src/index.ts` initially (fixed in a2e3a004), which broke the `import { fromRuntimeHeaders } from '@thymian/core'` pattern the docs teach.

PR #369, opened 2026-08-20
feat(core): add shared loadUserModule loader with extension dispatch

chaosmonster:
Adds one shared seam in `@thymian/core` for loading a **user-supplied** module — a rule, a rule set
or a plugin — so that `rule-loader.ts` and `base-cli-run-command.ts` can later migrate onto identical
semantics instead of each reinventing TypeScript loading.

**This PR adds the seam and its tests. It changes no call site.** `packages/core/src/rules/` and
`packages/common-cli/` are byte-identical to `main`; the diff is purely additive
(26 files, +1370/−0).

## What it does

`packages/core/src/load-user-module.ts` exports two functions and one options interface:

```ts
resolveUserModule(specifier, cwd?, options?): Promise<string | undefined>  // never throws
loadUserModule(resolvedPath): Promise<Record<string, unknown>>             // returns the namespace
```

Resolution and loading are separate steps because callers need the resolved path in its own right
(`loadRuleSet` uses it as a glob base) and need "cannot resolve" to stay distinguishable from
"failed to load", so each keeps its own user-facing message.

**Resolution order.** Relative specifiers are anchored at `cwd` before resolution — delegating them
would hand back core's *own* modules as the user's rule. Bare specifiers resolve installed packages
first (the user's project, then core's install directory) and fall back to `<cwd>/<specifier>` only
when nothing is installed under that name, so a same-named local file or directory cannot shadow and
silently execute in place of an installed package. Resolved paths are then gated on being absolute,
normalised to their on-disk casing, and checked against a closed allow-list of loadable extensions.

**Dispatch.** `/\.[cm]?ts$/` goes through jiti; everything else through a plain dynamic `import()`.

**jiti stays off the JavaScript path.** It is imported lazily via a memoised *promise* (memoising the
instance would let concurrent TypeScript loads each start their own import), and the only `jiti`
reference in the built output is inside that dynamic import. This matters because the built-in rules
are JavaScript and load on every single invocation.

**Exactly once.** A module's top-level side effects run once — across both dispatch branches, across
every spelling of its path, and under concurrency:

- jiti's `extensions` is narrowed to TypeScript so it delegates `.js` to Node's loader. With the
  default list a `.js` imported by both a natively-loaded JS rule and a jiti-loaded TS rule lands in
  two registries and evaluates **twice**, handing the two rules non-identical objects.
- Paths are canonicalised through `realpathSync.native`, so `x.ts`, `sub/../x.ts` and a symlink are
  one identity.
- An in-flight promise map de-duplicates concurrent TypeScript loads (jiti populates its cache only
  after a load completes).
- An `AsyncLocalStorage` chain detects a module that re-enters the seam during its own evaluation and
  reports the cycle rather than hanging on a promise only its own return can settle.

## Documented limitations

Each is measured against jiti 2.6.1 and recorded in the file header, not left implicit:

- TypeScript sources must use `export default`. Through jiti's interop, `module.exports = {…}` and
  `export = {…}` are indistinguishable from a module with only named exports, so synthesising a
  default would let a named-only module masquerade as a rule.
- tsconfig `paths` aliases are not honoured in a user TypeScript module. Nested *package* imports
  from the user's own `node_modules` do work; it is specifically the alias case.
- Only JavaScript and TypeScript are loadable. The excluded extensions each carry their reason at
  the constant — including `.node`, which is not importable on the `engines.node` floor at all.
- Both `require.resolve` anchors apply the **CJS condition set**, so a dual-published package
  resolves its CommonJS half and an ESM-only package falls through to jiti. Closing this needs a
  real ESM-conditions resolver; tracked separately.

## Dependency

`jiti@^2.6.1` added to `@thymian/core` — the exact range `plugin-sampler` already uses, so the
lockfile keeps one hoisted copy (the `package-lock.json` diff is a single line). jiti has zero
runtime dependencies and declares no `engines`, so `engines.node` stays `">=22"`.

## Verification

`core:lint` 0 errors / 14 warnings (unchanged from baseline) · `core:test` **382 passed** across 34
files, 49 of them new · `core:build` pass · `nx affected -t test` pass · `nx e2e e2e-tests` pass ·
`nx format:check` clean · `git diff main -- packages/core/src/rules packages/common-cli` empty.

Behaviour was additionally re-checked by running the **built** `dist/load-user-module.js` from
directories outside the monorepo — npm hoisting inside the workspace lets core's `require` see
packages a real consumer's would not, which is precisely what hid several defects from the
in-workspace suite during review.

### DIFF
--- a/packages/core/src/load-user-module.ts
+++ b/packages/core/src/load-user-module.ts
@@ -372,11 +372,14 @@ export interface ResolveUserModuleOptions {
  * `{ ok: false }`, carrying a `reason` when the seam knows something the caller could not work
  * out for itself. On success, `path` is an absolute filesystem path, never a `file://` URL.
  *
- * Resolution order stops at the first hit: `require.resolve`, then a `.mjs`/`.cjs` extension
- * guess ({@link resolveThroughGuessing}), then jiti's `esmResolve` — with the `<cwd>/<specifier>`
- * fallback for bare names slotted in before that last step. `require.resolve` already handles
- * TypeScript with an explicit extension, so the jiti fallback is reached for *resolution* only by
- * extensionless and NodeNext `.js`→`.ts` specifiers.
+ * Resolution stops at the first hit and runs the same three-resolver chain
+ * ({@link resolveLocation}) over at most two candidates: the specifier itself, and — for a bare
+ * name nothing is installed under — `<cwd>/<specifier>`. Within a candidate the order is
+ * `require.resolve`, then a `.mjs`/`.cjs` extension guess ({@link resolveThroughGuessing}), then
+ * jiti's `esmResolve`. `require.resolve` already handles TypeScript with an explicit extension, so
+ * the jiti step is reached for *resolution* only by extensionless and NodeNext `.js`→`.ts`
+ * specifiers. The candidates are exhausted in that order, not the resolvers: a bare name that only
+ * jiti can resolve inside `node_modules` still beats a same-named local file.
  *
  * @param specifier The user's rule, rule-set or plugin specifier.
  * @param cwd Directory that path-like specifiers resolve against, that anchors the jiti fallback,
@@ -403,11 +406,17 @@ export async function resolveUserModule(
     // unambiguously a path.
     resolved = await resolveLocation(path.resolve(cwd, specifier), cwd);
   } else {
-    // Installed packages FIRST — through all three resolvers, not just the first two. jiti is
-    // what resolves an installed package whose entry point is TypeScript, which `require.resolve`
-    // cannot; running it here rather than after the cwd fallback is what keeps the documented
-    // order honest. Preferring `<cwd>/<specifier>` up front is what let a same-named local file
-    // or directory shadow — and silently execute in place of — an installed package.
+    // Installed packages FIRST — through all three resolvers, not just the first two. The third
+    // one earns its place: a bare SUBPATH whose target is TypeScript (`my-pkg/rules`, or its
+    // NodeNext `my-pkg/rules.js` spelling, against a `node_modules/my-pkg/rules.ts`) is refused
+    // by `require.resolve` and answered by jiti — measured, both spellings. A package whose
+    // `main` is TypeScript needs no help here; `require.resolve` already returns it.
+    //
+    // So this step has to run before the cwd fallback, not after it: otherwise a local
+    // `<cwd>/my-pkg/rules.ts` would shadow the installed `my-pkg/rules.ts`, which is the one
+    // thing the installed-first order exists to prevent. Preferring `<cwd>/<specifier>` up front
+    // is what let a same-named local file or directory shadow — and silently execute in place
+    // of — an installed package.
     resolved =
       resolveThroughRequire(specifier, cwd) ??
       resolveThroughGuessing(specifier) ??
--- a/packages/core/src/load-user-module.ts
+++ b/packages/core/src/load-user-module.ts
@@ -15,7 +15,7 @@
  * 4. **jiti only for TypeScript, imported lazily.** `loadRules` runs on every single invocation
  *    for all built-in rules, and those are JavaScript; that path must never pay for jiti.
  *
- * Two limitations are deliberate and measured against jiti 2.6.1, not oversights:
+ * Three limitations are deliberate and measured against jiti 2.6.1, not oversights:
  *
  * - **TypeScript sources must use `export default`.** `module.exports = …` in a `.ts`/`.cts` file
  *   produces a namespace with no `default` key, and jiti's interop makes that shape *identical*
@@ -30,7 +30,8 @@
  *   under `paths: { "@lib/*": [...] }` resolves and then fails at load with a raw
  *   `MODULE_NOT_FOUND`. Nested *package* imports from the user's own `node_modules` DO work; it is
  *   specifically the alias case. Probably the first limitation a real user writing TypeScript
- *   rules meets, so it wants a decision rather than silence.
+ *   rules meets. Recorded as deliberately unsupported in story 34.5's documentation criteria.
+ *
  * Bare specifiers resolve installed packages first — the user's project, then core's own install
  * directory — and only fall back to `<cwd>/<specifier>` when nothing is installed under that name.
  * Two reasons for that order: a package the user named in their config is theirs, so a globally
@@ -54,19 +55,6 @@ import { isRecord } from './utils.js';
 const require = createRequire(import.meta.url);
 
 /**
- * Resolves through Node's own resolver, trying the **user's project before core's install
- * directory**.
- *
- * Anchoring only at core's directory (the single `require` above) meant a package present only in
- * the user's `node_modules` was invisible to it and fell through to the jiti fallback — so a
- * plain-JavaScript rule package paid for jiti, breaking contract item 4 (confirmed with
- * `JITI_DEBUG=1`). It also meant a name the user installed lost to a Thymian dependency of the
- * same name, which for a globally installed CLI is the wrong answer twice over.
- *
- * Only **bare** specifiers are affected: a relative one is already an absolute path by the time it
- * gets here, and both anchors resolve an absolute path identically.
- */
-/**
  * Cwd-anchored resolvers, memoised. `resolveThroughRequire` runs for every specifier — including
  * every built-in rule on every invocation, the path the file header singles out as needing to stay
  * cheap — and `cwd` is stable within a run, so building the anchor each time was pure waste.
@@ -84,6 +72,19 @@ function requireFrom(cwd: string): NodeJS.Require {
   return anchored;
 }
 
+/**
+ * Resolves through Node's own resolver, trying the **user's project before core's install
+ * directory**.
+ *
+ * Anchoring only at core's directory (the bare `require` above) meant a package present only in
+ * the user's `node_modules` was invisible to it and fell through to the jiti fallback — so a
+ * plain-JavaScript rule package paid for jiti, breaking contract item 4 (confirmed with
+ * `JITI_DEBUG=1`). It also meant a name the user installed lost to a Thymian dependency of the
+ * same name, which for a globally installed CLI is the wrong answer twice over.
+ *
+ * Only **bare** specifiers are affected: a relative one is already an absolute path by the time it
+ * gets here, and both anchors resolve an absolute path identically.
+ */
 function resolveThroughRequire(
   location: string,
   cwd: string,
--- a/packages/core/src/load-user-module.ts
+++ b/packages/core/src/load-user-module.ts
@@ -15,7 +15,9 @@
  *    everything else through a plain dynamic `import()`.
  * 3. **Never import-then-retry, and never evaluate twice.** A module's top-level side effects
  *    must run exactly once — across BOTH dispatch branches, across every spelling of its path,
- *    and under concurrency. A cycle is reported rather than deadlocked.
+ *    and under concurrency. A cycle is reported rather than deadlocked — including a cycle formed
+ *    by two concurrent roots waiting on each other, which needs the wait-for graph
+ *    ({@link blockedOn}) and not just the evaluation chain.
  * 4. **jiti only for TypeScript, imported lazily.** `loadRules` runs on every single invocation
  *    for all built-in rules, and those are JavaScript; that path must never pay for jiti.
  *
@@ -494,6 +496,11 @@ function canonicalise(resolvedPath: string): string {
  * its own return can settle, and the process hangs silently — Node exits 13, "Detected unsettled
  * top-level await". `AsyncLocalStorage` carries the chain into module top-level code on both
  * branches (verified), so re-entry is detectable and can be reported instead of hung.
+ *
+ * It catches re-entry on ONE path and nothing more, which is all a per-path structure can do. Two
+ * roots entering concurrently and waiting on each other is the same defect through a door no chain
+ * can see; {@link blockedOn} is what closes that one, and the two work together rather than either
+ * replacing the other.
  */
 const evaluationChain = new AsyncLocalStorage<readonly string[]>();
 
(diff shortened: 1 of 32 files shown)


matthyk (commented):
We should replace all existing dynamic imports with these features in a follow-up PR.

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,481 @@
...
+
+  if (!TYPESCRIPT_EXTENSION.test(canonical)) {
+    return await evaluationChain.run(nested, () => importNatively(canonical));
+  }
+
+  const pending = inFlightTypeScriptLoads.get(canonical);
+
+  if (pending !== undefined) {
+    return await pending;
+  }
```
matthyk, line 806: **Blocker — a concurrent cycle deadlocks here instead of reporting.**

The in-flight dedup short-circuits *before* the chain is extended, so the cycle guard never sees the other root's chain. `loadRules` fans out with `Promise.all` over array input (`rule-loader.ts:285`), so this is reachable rather than theoretical:

- Root A sets `inFlight{a}` with chain `[a]`; root B sets `inFlight{b}` with chain `[b]`
- A's body calls `loadUserModule(b)`: chain `[a]` doesn't contain `b`, but `inFlightTypeScriptLoads.get(b)` hits, so it takes `return await pending` and the chain is never extended
- B's body does the mirror image
- Both promises now await each other → the process hangs and exits 13 (`Detected unsettled top-level await`) instead of raising the cycle error this machinery exists to raise

Root cause is that two different requirements share one lookup: dedup ("evaluate exactly once under concurrency") and cycle detection. The `pending` branch has to consult the chain before returning — or, better, the two concerns get separate state.

Worth reconsidering whether `AsyncLocalStorage` is needed at all: `loadRules`/`loadRuleSet` already thread recursion state as an explicit parameter (`profileConfig`, `rule-loader.ts:277`), and a `chain: readonly string[]` alongside it would catch real rule-set cycles with none of this machinery. ALS is only strictly required when *user top-level code* re-enters the loader — the fixture's shape, but no caller's.

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,481 @@
...
+      // Nothing is installed under that name, so a local file or directory of that name is what
+      // the user meant. Resolving the absolute path (rather than using it verbatim) keeps a
+      // directory holding `index.js` or a `package.json` `main` working, which is the shape
+      // today's `rule-loader` accepts for `rules: ['my-rules']`.
+      const fileLocation = path.resolve(cwd, specifier);
+
+      if (existsSync(fileLocation)) {
+        resolved = resolveThroughRequire(fileLocation, cwd);
+      }
+    }
```
matthyk, line 511: **Blocker — a bare specifier can never reach a local TypeScript file.**

`rules: ['my-rule']` with `<cwd>/my-rule.ts` on disk:

- `resolveThroughRequire('my-rule')` fails (bare = `node_modules` only)
- `existsSync('<cwd>/my-rule')` is false — only `my-rule.ts` exists — so this entire block is skipped
- `resolved ??= await resolveThroughJiti(specifier, cwd)` on line 308 then gets the *bare* name, so it searches `node_modules` again

Result: `undefined` → "cannot resolve". The identical file resolves fine as `./my-rule` (the relative branch passes the absolute `path.resolve(cwd, specifier)` into jiti) and as an absolute path.

Two consequences: extensionless local TS rules are the headline use case for this seam, and this blocks migrating `rule-loader`, whose `existsSync`-then-`require.resolve` path (`rule-loader.ts:298-306`) accepts exactly this shape today.

Fix is probably to hand `fileLocation` rather than `specifier` to the jiti fallback when the cwd candidate is plausible, instead of gating on `existsSync` of the exact spelling.

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,481 @@
...
+      //
+      // `extensions` is narrowed to TypeScript so jiti DELEGATES `.js` to Node's loader instead
+      // of registering it itself. With the default list, a `.js` module imported by both a
+      // natively-loaded JS rule and a jiti-loaded TS rule lands in two registries and evaluates
+      // TWICE, handing the two rules non-identical objects — measured, and the thing contract
+      // item 3 forbids. Narrowing unifies them: one evaluation, `===` identical. Resolution is
+      // unaffected (the list governs extension *guessing*, and every case here was measured
+      // identical), because explicit paths and `.js`→`.ts` NodeNext mapping still work.
+      createJiti(import.meta.url, {
+        extensions: ['.ts', '.mts', '.cts'],
```
matthyk: **Blocker — "Resolution is unaffected … every case here was measured identical" isn't accurate, and nothing tests the gap.**

jiti 2.6.1 defaults to `['.js','.mjs','.cjs','.ts','.tsx','.mts','.cts','.mtsx','.ctsx']` and derives `additionalExts` from it by filtering out only `.js`. Narrowing to `['.ts','.mts','.cts']` makes `additionalExts` exactly that — so `.mjs`, `.cjs`, `.tsx`, `.mtsx`, `.ctsx` drop out of extension *guessing* as well as out of registration.

`resolveUserModule('./my-rule', cwd)` with `<cwd>/my-rule.mjs` on disk: CJS `require.resolve` never tries `.mjs`, and jiti no longer guesses it → `undefined`. Unnarrowed, jiti resolved it.

`declines every extension outside the allow-list` only writes explicit-extension names, so nothing turns red.

The one-registry rationale above is sound and I'd keep the narrowing — the ask is to correct the claim and add extensionless `.mjs`/`.cjs` cases so the real boundary is pinned.

packages/core/test/fixtures/user-modules/cycle-a.ts
```diff
@@ -0,0 +1,9 @@
+// Fixture: re-enters the loader for its partner during its own evaluation. Left undetected this
+// awaits a promise only its own return can settle, and the process hangs.
+import { loadUserModule } from '../../../src/load-user-module.js';
+
+export const partner = await loadUserModule(
+  new URL('./cycle-b.ts', import.meta.url).pathname,
```
matthyk: **Blocker — this fails on the `windows-2022` leg.**

`new URL(...).pathname` is not a filesystem path on Windows (and is percent-encoded everywhere). There it is `/C:/…/cycle-b.ts`:

- `path.isAbsolute` accepts it
- `realpathSync.native` throws, so `canonicalise` falls back to `path.resolve('/C:/…')`, which re-anchors on the current drive → `C:\C:\…`
- the chain check misses, jiti throws a raw `MODULE_NOT_FOUND`, and `expect.objectContaining({ name: 'UserModuleLoadError' })` fails

Use `fileURLToPath(new URL('./cycle-b.ts', import.meta.url))`. Same fix needed on `cycle-b.ts:5`.

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,481 @@
...
+    // case-insensitive volume `./Rule.TS` would otherwise reach a dispatch that matches neither
+    // branch. Symlinks are unaffected: both resolvers already return realpaths.
+    normalised = realpathSync.native(resolved);
+  } catch {
+    return undefined;
+  }
+
+  if (unloadableReason(normalised) !== undefined) {
+    return undefined;
+  }
```
matthyk: **Blocker — the reason is computed and then discarded, so "exists but isn't loadable" is indistinguishable from "not found".**

`unloadableReason` has just produced the precise sentence ("only JavaScript and TypeScript modules can be loaded") and it's dropped on the floor. Point `rules:` at `./rules.yaml` (or `./types.d.ts`) and the user gets `Cannot resolve rule source ./rules.yaml.` — told the file can't be found when it is right there.

The header at lines 25-27 says the allow-list is enforced at resolution "so the caller phrases the error", but the caller is handed nothing to phrase it with.

This gets worse once the three existing call sites migrate: each phrases its own error (`RuleLoadError`, `PluginLoadError`, `HookImportError`), so all three would have to re-derive what this function already knew. Suggest returning a discriminated result — `{ ok: true; path: string } | { ok: false; reason?: string }` — rather than `string | undefined`.

chaosmonster: Fixed in `b3dd9162`. `resolveUserModule` returns the shape you suggested:

```ts
{ ok: true; path: string } | { ok: false; reason?: string }
```

`reason` is optional rather than `string | null` because most failures genuinely have nothing to add beyond "not found" — a caller writes `result.reason ?? <its own default>`. Callers still own the sentence; the seam owns the fact.

AC1's verbatim signature was amended rather than the change deferred, on the reasoning that this is the last cheap moment: the seam ships with no production consumer on `main`, so the blast radius today is one unmerged branch. Story 34.2's branch consumes the old shape at `rule-loader.ts:375` and needs updating when it rebases — and that rebase is also the fix for the `Cannot resolve rule source ./rules.yaml.` message you cited, which can now say what is actually wrong.

Tests assert the **reason**, not just the decline: a bare "is it undefined" check cannot tell the two failure modes apart, which is the conflation being fixed. Green on CI.

packages/core/src/load-user-module.ts
chaosmonster: Fixed in `39bed4b7`. Both halves of your trace confirmed against jiti 2.6.1: bare `my-rule` answers `undefined`, `<cwd>/my-rule` answers `my-rule.ts`.

Both branches now share one chain — `require.resolve` → `.mjs`/`.cjs` guessing → jiti, over an absolute path. The `existsSync` gate is dropped rather than corrected: every resolver in that chain is already existence-checked internally (`require.resolve` throws, the guessing step calls `isFile`, the jiti step ends on `existsSync`), so the gate could only ever remove reachable answers.

One thing worth flagging, because it is load-bearing and your suggested fix would have tripped it. Handing `fileLocation` to the jiti fallback *in place of* the bare name reintroduces shadowing by a narrower door: `require.resolve` refuses a bare TypeScript **subpath** (`my-pkg/rules`, and its NodeNext `my-pkg/rules.js` spelling, against a `node_modules/my-pkg/rules.ts`) and jiti answers both — measured — so a local `<cwd>/my-pkg/rules.ts` would win over the installed one. Resolution is therefore ordered by **candidate** rather than by resolver: the bare name goes through all three resolvers, *then* `<cwd>/<specifier>` is tried as a separate candidate. Two tests pin that and go red if the candidates swap.

One deliberate behaviour change fell out: where a local path is resolvable by require/guessing *and* the bare name only by jiti, the answer flips from local to installed. The file header has always claimed installed-first for bare specifiers; the old order contradicted it.

This also closes Copilot's note at `r3821800962` — same defect, directory spelling. Green on CI including the `windows-2022` legs.

packages/core/src/load-user-module.ts
chaosmonster: Fixed in `922fc040`, green on CI.

Reproduced before fixing, against the built `dist/` from an external cwd: two roots, `Promise.allSettled` over both, nothing settles — Node prints `Warning: Detected unsettled top-level await` and exits with neither root resolved nor rejected. Your trace was exact, step for step.

Your framing was the fix: two requirements were sharing one lookup, and they have separate state now. `inFlightTypeScriptLoads` answers "is this file already loading" and nothing else; a new `blockedOn` wait-for graph answers "would waiting for it close a ring". Before joining another evaluation's load, the seam walks the graph from the awaited path looking for the module it is evaluating right now. The walk returns the path rather than a boolean so the error names the real ring instead of the two modules that happened to notice each other: `cb.ts -> ca.ts -> cb.ts` for two roots, `r3.ts -> r1.ts -> r2.ts -> r3.ts` for a three-root ring (the transitive case, verified). Exactly one root detects — the check and the edge registration are synchronous — and its throw rejects the other's await, so both callers get the framed error.

**On the `AsyncLocalStorage` question** — I kept it, and I'd argue that part is orthogonal rather than an alternative. At the moment root B decides, its chain is `[b]` and it wants `a`; `a` genuinely is not in its chain. What makes it a cycle is that `a`'s evaluation is *already waiting on* `b` — dynamic state owned by another async path. Threading `chain` explicitly alongside `profileConfig` carries exactly the same per-path information and misses this identically, so the choice is about how the chain travels, not about what the chain can know. Happy to revisit if you read it differently.

The false positive this must not become: two roots waiting on one shared dependency is legitimate and common, and waiting there is precisely what makes exactly-once hold. Pinned separately — the shared module evaluates once and both roots receive the `===` same namespace.

**Your comment also surfaced something bigger, which is worth your attention independently of this fix.** My first version of the new test passed *with the fix reverted*. Cause: a fixture doing `import { loadUserModule } from '../../../src/load-user-module.js'` is loaded by **jiti**, whose registry is separate from the one Vitest used for the test file — so the fixture receives a **second copy of the seam**, with its own in-flight map, evaluation chain and wait-for graph. Measured directly: the two `loadUserModule` functions are not `===`.

So the pre-existing `cycle-a`/`cycle-b` test was never exercising the module under test — including through round 4, which hardened those fixtures. Its green was real but described the jiti copy. Fixtures now take the loader from `globalThis` via a documented `fixtures/user-modules/loader-under-test.ts`, and I re-checked afterwards that the sequential cycle test still times out when the chain check is removed, so it kept its purpose and gained a real target.

Both new tests are mutation-checked and separable: reverting the wait-for check reddens the concurrent-cycle test by timeout — the defect is a hang, so the test can only assert that both roots settle at all — and leaves the shared-dependency test green; making the walk unconditional does the reverse.

One thing deliberately not covered: a ring closed *partly* outside this seam (a jiti-loaded module reaching its partner through a plain `import` rather than through `loadUserModule`) is jiti's registry to resolve and outside what this graph can observe. No caller shape produces it today.

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,549 @@
...
+      const fileLocation = path.resolve(cwd, specifier);
+
+      if (existsSync(fileLocation)) {
+        resolved =
+          resolveThroughRequire(fileLocation, cwd) ??
+          resolveThroughGuessing(fileLocation);
+      }
+    }
+
+    resolved ??= await resolveThroughJiti(specifier, cwd);
```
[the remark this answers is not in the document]
chaosmonster: Fixed in `39bed4b7`, together with `r3821456158` — the same defect, and your directory spelling (`<cwd>/my-rules/index.ts`) is the case that made it clear one fix covers both.

The cwd fallback now runs the full chain — `require.resolve` → `.mjs`/`.cjs` guessing → jiti — over `path.resolve(cwd, specifier)`, sharing that chain with the relative branch so the two cannot drift again. The `existsSync` gate is gone rather than corrected: every resolver in the chain already checks existence itself, so gating on the extensionless spelling could only remove reachable answers.

Verified against the built `dist/` from a cwd outside the monorepo: bare `my-rule` → `<cwd>/my-rule.ts`, bare `ts-dir` → `<cwd>/ts-dir/index.ts`, and bare `nodenext.js` → `<cwd>/nodenext.ts`, all three loading correctly.

One qualification on "otherwise preserving the cwd-anchored path": the jiti step stays on the **bare** candidate as well, because it is what resolves a bare TypeScript subpath inside `node_modules`, which `require.resolve` refuses. Swapping it for the cwd path would let a local `<cwd>/my-pkg/rules.ts` shadow an installed `my-pkg/rules.ts`. The cwd path is a separate, later candidate instead.

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,549 @@
...
+function requireFrom(cwd: string): NodeJS.Require {
+  let anchored = cwdRequireCache.get(cwd);
+
+  if (anchored === undefined) {
+    anchored = createRequire(pathToFileURL(path.join(cwd, '_')));
+    cwdRequireCache.set(cwd, anchored);
+  }
+
+  return anchored;
+}
```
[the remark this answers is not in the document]
chaosmonster: Normalisation done in `148a0de1` — but the stated reason does not hold, and it seems worth being explicit about that rather than letting the commit imply otherwise.

`pathToFileURL()` does not throw on a relative path: it resolves the path against `process.cwd()`. Checked on **22.19** (this package's `engines.node` floor) and **26.7**, the ends of the CI matrix (22/24/26). So `requireFrom` never threw and the never-throws contract was not at risk.

There is a real defect behind the suggestion, which is why the change landed anyway. The anchor memo was keyed by the **raw** `cwd` string, and a relative `cwd` is interpreted at two different times: `pathToFileURL` resolves it when the anchor is *built*, while every other resolver resolves it again on each *call*. After a `process.chdir` those disagree — the memo hands back an anchor pointing at the old directory, `require.resolve` misses a package that is plainly installed, and resolution falls through to the jiti fallback. A plain-JavaScript package then pays for jiti, which contract item 4 in the file header forbids. Measured with `JITI_DEBUG=1`: jiti initialised on the second call, and not on a control call with an absolute `cwd`.

Worth noting for anyone reading the test: the returned path was **correct** either way — jiti found what the stale anchor had missed — so only the cost differed. That is why the test asserts the jiti counter rather than the resolved path; a result assertion is structurally blind to this defect.

`cwd` is now resolved once on entry to `resolveUserModule` and that value used everywhere, cache key included, so `.`, `./`, the absolute spelling and `<abs>/./` collapse onto one entry instead of one per spelling.

PR #379, opened 2026-08-24
fix(core): deserialize parameters before schema validation

matthyk:
## Problem

Query parameters, path parameters and request/response headers were validated by handing the **raw wire value** to Ajv. Wire values are always strings, so every non-string parameter in every OpenAPI document was a false positive:

```
✖ error: Request query parameters do not conform to the schema (1 issue)
      ✖ query parameter "year" must be integer
          expected: "integer"
          actual: "2026"
```

The request was correct. `2026` is exactly what the description asked for.

`deepObject` and array parameters failed twice over — the documented parameter was reported *absent* while its wire keys were reported *undocumented*.

Reported as thymianofficial/thymian-internal#624.

## Approach

OpenAPI's `style`/`explode` describe how a typed value was **serialized into** a wire string, so validating the wire form is validating the wrong thing. This adds `packages/core/src/http-testing/deserialize-parameter.ts` — the inverse of the existing `serialize-parameter.ts` — which rebuilds the typed value before validation. All four validators use it.

Two design points worth knowing before reading:

- **Object reassembly happens in the query orchestrator, *before* the missing- and additional-parameter checks.** Both key off the parsed map, so `?filter[customers_id]=1` must already be the key `filter` by the time they run. That single ordering choice is what fixes both `deepObject` false positives.
- **Coercion is schema-directed and conservative.** A value that is not an unambiguous lexical form of its declared type is returned unchanged, so `?year=abc` still fails with a real message. `007`, `+5`, `1e999` and anything past 2^53 deliberately stay strings — the fix may only turn a false failure into a pass when it is certain, never the reverse.

Schema types are resolved through `$ref`, `allOf`, `anyOf`/`oneOf`, `enum` and `const`. This matters more than it sounds: `plugin-openapi` **localizes** `$ref` rather than inlining it, so a schema factored into `components/schemas` arrives with no top-level `type`. Resolution goes through one exported accessor that every structural decision shares.

`ajv.ts` is deliberately untouched. The same instance validates request and response **bodies**, where a JSON `"5"` in an integer field is a genuine conformance defect that must keep failing. A second coercing instance was considered and rejected: it would fix scalars only, and still leave arrays, objects and `deepObject` wrong.

## Verification

Against the reporting user's real OpenAPI document and HAR capture, diffing every `✖` line before and after:

```
< ✖ error: Request query parameters do not conform to the schema (1 issue)
< ✖ query parameter "year" must be integer
```

Two violations removed, **zero added**. Nothing else moved.

- **445 tests pass** in `core` (333 baseline + 112 new)
- All **17 project suites** pass
- Lint and build clean; no new warnings
- `git diff --stat -- packages/core/src/http-testing/validate/ajv.ts` is empty

## Scope

Implemented: `form` (explode true/false), `deepObject`, `simple` — the styles the acceptance criteria name, and the only ones any spec in either repo actually uses (`deepObject` appears 17 times in the reference document; the rest appear zero times).

`spaceDelimited`, `pipeDelimited`, `label` and `matrix` report an explicit `info` rather than a wrong result, tracked as thymianofficial/thymian-internal#672 and thymianofficial/thymian-internal#673. Note #672 is **not** a standalone deserializer fix: `serializeQueryParameter` currently maps both delimited styles to the *form* template, so it already emits the wrong wire format when exploded. Fixing only the read side would make thymian reject its own generated requests.

An unsupported style only forfeits validation for **structured** values. A scalar has no structure to restore, so `maxLength`/`pattern`/`enum`/`format` still apply regardless of style — that is the second commit, correcting a coverage regression the first one introduced.

## Reviewing this

Two adversarial review rounds ran against the diff and found ~30 real defects, each verified at runtime before being fixed. Round 1 caught the version of this change that passed all its own tests while being **ineffective for any document using `$ref`** — every fixture had used inline schemas. Round 2 caught the same class reintroduced one layer up, where a local non-resolving type read meant resolution never reached the split/fold decisions.

**Where the thin ice is.** Three pieces added during round 2 never got an adversarial pass and deserve the closest reading:

- `splitHeaderList` — the RFC 9110 §5.6.1 quoted-string-aware list parser
- `schemaTypes`' `allOf` **intersection** (first-wins made results depend on member order)
- the `form`/`explode: true` object folding loop and its ambiguity guard

`rawPairs` also replaced `URLSearchParams` with a hand-rolled parser, because delimiter splitting must happen before percent-decoding and `URLSearchParams` only returns decoded values. That trade is deliberate and tested, but it is the kind of swap that can quietly lose parity.

Closing keywords do not cross repositories, so this cannot auto-close thymianofficial/thymian-internal#624.


packages/core/src/http-testing/deserialize-parameter.ts
```diff
@@ -0,0 +1,789 @@
...
+ */
+function deserializeSimple(
+  raw: string | string[],
+  schema: ThymianSchema | undefined,
+  style: SerializationStyle | undefined,
+  split: (value: string) => string[],
+): DeserializeResult {
+  const serializationStyle = style ?? SIMPLE_DEFAULT_STYLE;
+
+  if (serializationStyle.style !== 'simple') {
```
atennert: **[high — partial AC6 violation]** "A scalar has no structure to restore" is false for `label` and `matrix`: `serializePathParameter` prefixes scalars too (`{.name}` → `.42`, `{;name}` → `;id=42`), so the extracted path segment still carries the prefix. `deserializeScalar('.42', { type: 'integer' })` leaves it a string and Ajv reports `must be integer` for a conforming request — with **no diagnostic**, which is exactly the "silently wrong validation result" AC6 forbids. String-typed params fare no better: `maxLength`/`pattern` now count the `.`/`;name=` prefix.

For `spaceDelimited`/`pipeDelimited` the shortcut is sound (those styles are only defined for arrays/objects), so the fix is to exclude `label`/`matrix` from it — return `unsupported` for them (or strip the prefix, but `unsupported` matches this module's own contract).

packages/core/test/http-testing/deserialize-parameter.test.ts
```diff
@@ -0,0 +1,790 @@
...
+    ).toEqual([1, 2, 3]);
+  });
+});
+
+describe('unsupported styles only forfeit STRUCTURED values', () => {
+  // A style describes how a structured value was flattened onto the wire.
+  // A scalar has no structure to restore, so an unsupported style must not
+  // cost it the schema checks that never depended on the style.
+  it.each(['label', 'matrix'] as const)(
+    'still validates a string-typed path parameter under %s',
```
atennert, line 787: **[test — ties to the scalar-shortcut finding]** This wire form is unrealistic: a `label`/`matrix` scalar arrives *prefixed* (`.abcdef`, `;p=abcdef`), never bare — which is why this test passes while the shortcut misvalidates real traffic. Once `label`/`matrix` scalars take the unsupported path, this test should feed the realistic prefixed form and assert the honest outcome (info, not a pass).

packages/core/src/http-testing/validate/validate-request-query-parameters.ts
```diff
@@ -1,49 +1,270 @@
...
+    ([, parameter]) =>
+      parameter.style?.style === 'form' &&
+      parameter.style.explode &&
+      typesOf(parameter.schema).includes('object'),
+  );
+  const claims = new Map<string, number>();
 
-    if (!key) {
+  for (const [, parameter] of objectParameters) {
+    for (const property of Object.keys(parameter.schema?.properties ?? {})) {
```
atennert: **[high]** This claims loop — and `declaredProperties` at line 200 — reads `parameter.schema?.properties` **raw**, while the gate above uses the resolving `typesOf`. A parameter whose object schema sits behind `$ref` (the exact shape plugin-openapi emits, per this file's own top comment) passes the gate but shows `[]` properties, so it is treated as **free-form**: every unclaimed query key — genuine typos included — is folded into it, `checkForAdditionalQueryParameters` goes silent, the folded object fails Ajv bogusly, and its real properties never enter `claims`, blinding the ambiguity guard below.

The comment at the top of this file ("a second, non-resolving copy of this logic is how `$ref`ed parameters slipped through a whole review round") sits directly above these two fresh non-resolving copies — third instance of the class rounds 1–2 caught. Resolving through `flattenSchema` (exporting it from `deserialize-parameter.ts`) before reading `properties` closes it.

packages/core/src/http-testing/validate/validate-request-query-parameters.ts
```diff
@@ -1,49 +1,270 @@
...
+      brackets !== undefined &&
+      declaredParameter(request, bracketName)?.style?.style === 'deepObject'
+    ) {
+      const [, property] = SINGLE_PROPERTY.exec(brackets) ?? [];
+
+      // Nested (`filter[a][b]`) or empty (`filter[]`) property paths are not
+      // OpenAPI `deepObject`. Absorb them so they are not misreported as
+      // undocumented parameters, and mark the parameter unvalidatable.
+      if (property === undefined || property === '') {
+        nested.add(bracketName);
```
atennert: **[medium]** `nested` is dead code — written here, never read anywhere. The comment promises the parameter is "marked unvalidatable", but nothing consumes the set: `?filter[a][b]=1` alone yields `filter = {}`, which Ajv then fails with `must have required property 'a'` — a fabricated failure blaming the request for something that was never parsed, i.e. the defect class this module exists to remove. The regression test only asserts the key isn't called undocumented, so the fabrication is unexercised.

Either consume `nested` (emit an info à la `unsupportedStyleMessage` and skip schema validation for that parameter) or drop the set and correct the comment.

packages/core/src/http-testing/deserialize-parameter.ts
```diff
@@ -0,0 +1,789 @@
...
+  let quoted = false;
+  let escaped = false;
+
+  for (const char of raw) {
+    if (escaped) {
+      current += char;
+      escaped = false;
+    } else if (quoted && char === '\\') {
+      escaped = true;
+    } else if (char === '"') {
```
atennert: **[medium]** The splitter *rewrites* member values: DQUOTEs and backslash escapes are consumed rather than appended, so `If-Match: "abc", W/"d\"ef"` yields members `abc` / `W/d"ef`. Members reach Ajv unquoted/unescaped — any `items` `pattern`/`enum`/`const` written against the wire form (entity-tags **are** quoted on the wire) falsely fails, and `maxLength` is measured after stripping, so it can falsely pass. Quote state should only suppress the comma split; every character should still be appended.

Secondary defect in the same loop: an unbalanced quote (`a,"b,c`) makes every later comma data, silently changing how malformed input splits. The test asserting `'"abc"'` → `abc` codifies the dequoting rather than questioning it.

packages/core/src/http-testing/validate/validate-request-query-parameters.ts
```diff
@@ -1,49 +1,270 @@
...
       continue;
     }
 
-    const decodedKey = decodeURIComponent(key);
-    const decodedValue = value ? decodeURIComponent(value) : '';
+    const declaredProperties = Object.keys(parameter.schema?.properties ?? {});
+    // A free-form object (`additionalProperties`/`patternProperties`) declares
+    // no property names, so every unclaimed key is a candidate.
+    const candidates =
+      declaredProperties.length > 0 ? declaredProperties : [...scalars.keys()];
```
atennert, line 233: **[medium]** Free-form objects bypass the ambiguity guard, making folding declaration-order-dependent: `claims` counts only *declared* property names, so a free-form object claims nothing while its candidate set is **all** remaining scalar keys. Declare `extra` (free-form, form/explode:true) before `filter` (`properties: {role}`) and send `?role=admin`: `extra` swallows `role` and deletes it from `scalars`; `filter` is then reported "required but not included" although its property was sent. Counting free-form candidates into `claims` — so a key claimable by two objects becomes ambiguous and folds into neither — matches the guard's own stated intent.

**[question]** Related semantics gap: an object with both `properties` and `additionalProperties` never gets extra keys folded — `?role=admin&extra=x` reports `extra` as undocumented although the schema accepts it. Intended strictness, or should an `additionalProperties`/`patternProperties` object be a candidate for unclaimed keys when it is the only claimant?

packages/core/src/http-testing/deserialize-parameter.ts
```diff
@@ -0,0 +1,789 @@
...
+    // scanning the whole literal would reject the legitimate `0e5`.
+    const significand = wire.split(/[eE]/)[0] ?? wire;
+
+    // `1e999` overflows to Infinity and `1e-999` underflows to 0; past 2^53 the
+    // conversion is lossy. All three would validate against constraints the
+    // written value violates.
+    if (
+      Number.isFinite(parsed) &&
+      (parsed !== 0 || !/[1-9]/.test(significand)) &&
+      (Number.isSafeInteger(parsed) || !Number.isInteger(parsed))
```
atennert: **[question — documented trade-off, worth deciding deliberately]** Every integral value past 2^53 still false-fails on type: `?id=9007199254740993` against `{type: 'integer', format: 'int64'}` → `must be integer` for an ordinary Snowflake-style ID — the bug class this PR removes, surviving for the most common large-value case. The lossiness rationale is also uneven: every non-integral decimal is lossy too and is accepted via `!Number.isInteger(parsed)`, while the exactly-representable `1e21` (= 2^21·5^21) is rejected.

Options: (a) keep as-is and track it; (b) emit an `info` for lexically-valid-but-unconvertible magnitudes, mirroring the unsupported-style contract — "not validated" beats "falsely failed"; (c) convert when exactly representable. Unsupported *styles* get the info escape hatch; unsupported *magnitudes* currently keep silently failing, which is hard to defend by this module's own principle.

atennert (commented):
## Adversarial review — round 1 (head 6fa9bf1)

Three independent review layers ran over the full diff — blind adversarial, exhaustive edge-case walk, and an acceptance audit against thymianofficial/thymian-internal#624 — and every finding below was re-verified against the head sources before posting.

**Overall:** the mechanism is the right one. Schema-directed, conservative deserialization on the untouched strict Ajv instance is exactly what #624 asked for (verified: no `coerceTypes` anywhere, `ajv.ts` clean), the `rawPairs` split-before-decode trade is justified and directly tested, and 5 of 6 acceptance criteria hold with real test evidence. The findings that remain:

- **2 high** — scalar `label`/`matrix` path parameters silently false-fail (partial AC6 violation, `deserialize-parameter.ts:662`); the `form`/`explode: true` folding logic is `$ref`-blind (`validate-request-query-parameters.ts:182`) — the defect class your rounds 1–2 caught, alive one layer up.
- **3 medium** — the `nested` set is dead code and a nested-only deepObject fabricates a required-property failure (`validate-request-query-parameters.ts:141`); `splitHeaderList` rewrites member values while splitting (`deserialize-parameter.ts:746`); free-form exploded objects defeat the ambiguity guard and make folding declaration-order-dependent (`validate-request-query-parameters.ts:204`).
- **~12 minor / test notes**, inline.
- **2 open design questions**, asked inline: integral magnitudes past 2^53 (`deserialize-parameter.ts:453`) and folding semantics for `additionalProperties` objects (`validate-request-query-parameters.ts:204`).

One finding with no single anchor: the four validators now carry four copies of the same ~25-line deserialize → unsupported-info → compile block, and the query variant has already diverged from the other three (its three-way object branch). A shared helper taking the subject label and a `DeserializeResult` would collapse them before they rot independently — the info-message contract currently lives in four places.

**Per-AC audit (vs thymian-internal#624):**

| AC | Verdict |
|---|---|
| 1 — `year=2026` passes; `abc`/`2100` fail with accurate messages | Pass (nit: the 2100 test never asserts the `maximum` bound is named) |
| 2 — scalars + null-unions from wire form in query/path/headers | Pass |
| 3 — arrays, `form` explode true/false incl. single-element | Pass, with an end-to-end test gap (inline at the non-exploded test) |
| 4 — `deepObject` reconstructed, not reported additional/missing | Pass |
| 5 — `=` and `+` parse correctly | Pass |
| 6 — unsupported style ⇒ explicit diagnostic, never silently wrong | **Partial fail** — structural values yes; `label`/`matrix` scalars validate the raw prefixed wire form with no diagnostic |

packages/core/src/http-testing/deserialize-parameter.ts
```diff
@@ -0,0 +1,789 @@
...
+  const types = schemaTypes(schema, root ?? schema);
+
+  // An `enum` names its permitted values outright. When exactly one of them
+  // has this wire form, that member IS the value — otherwise a mixed enum such
+  // as `['auto', 10, 20]` would take the `string` branch below and `?limit=10`
+  // would fail enum membership as the string '10'.
+  const members = enumMembers(schema, root ?? schema);
+
+  if (members) {
+    const matches = members.filter((member) => String(member) === wire);
```
atennert: **[minor]** `String(member)` matches non-lexical coincidences: an array member `[1,2]` stringifies to `'1,2'` and an object member to `'[object Object]'`, so a wire item literally spelled that way deserializes into the member and *passes* validation the raw string should fail. Restricting the match to primitive members (`typeof member !== 'object'`, which also excludes `null`) keeps the header contract — "converts only when the wire form is an unambiguous lexical representation… never guesses".

packages/core/src/http-testing/validate/validate-request-query-parameters.ts
```diff
@@ -122,8 +382,13 @@ export function validateRequestQueryParameters(
...
   request: ThymianHttpRequest,
 ): HttpTestCaseResult[] {
-  const queryString = path.includes('?') ? (path.split('?')[1] ?? '') : '';
-  const queryParams = parseQueryString(queryString);
+  const separator = path.indexOf('?');
+  // Everything after the FIRST `?` is the query string — a later `?` is an
+  // ordinary character within it, not a second delimiter — and a `#` fragment
+  // is not part of the query at all.
+  const queryString =
+    separator === -1 ? '' : (path.slice(separator + 1).split('#')[0] ?? '');
```
atennert: **[minor]** When the only `?` sits inside the fragment (`/api#frag?year=2026`), `indexOf('?')` finds it and fragment content is parsed as the query string — required params falsely satisfied, phantom undocumented-parameter failures. Strip the fragment first (as `extractPathParameters` now does), then look for `?`.

packages/core/src/http-testing/validate/validate-request-query-parameters.ts
```diff
@@ -1,49 +1,270 @@
...
+  const scalars = new Map<string, string[]>();
+
+  const addProperty = (name: string, property: string, value: string) => {
+    const target = objects.get(name) ?? new Map<string, string[]>();
+    target.set(property, [...(target.get(property) ?? []), value]);
+    objects.set(name, target);
+  };
+
+  for (const [rawKey, rawValue] of rawPairs(search)) {
+    const key = decodeFormComponent(rawKey);
```
atennert: **[minor]** Bracket parsing runs on the *decoded* key — the split-before-decode discipline applied to values is violated for keys. `?filter[a%5Db]=1` (deepObject property `a]b`) decodes to `filter[a]b]`, fails `BRACKET_KEY`, and the whole string is reported as an undocumented scalar while `filter` looks absent; conversely a percent-encoded `[` in an ordinary parameter name can *become* a bracket key after decoding and be mis-folded. Match `BRACKET_KEY` on the raw key, then decode the name and each bracket token.

packages/core/src/http-testing/validate/validate-request-query-parameters.ts
```diff
@@ -1,49 +1,270 @@
...
+      !declared.style.explode &&
+      !declaredTypes.includes('string') &&
+      (declaredTypes.includes('array') || declaredTypes.includes('object'));
+
+    // Split the still-encoded value, then decode each item.
+    const values = isNonExplodedList
+      ? splitWireList(rawValue, decodeFormComponent)
+      : [decodeFormComponent(rawValue)];
+
+    scalars.set(key, [...(scalars.get(key) ?? []), ...values]);
```
atennert, line 198: **[minor]** Repeated occurrences of a `form`/`explode: false` parameter merge silently: `?ids=1,2&ids=3` becomes one clean `[1,2,3]`, where main surfaced the duplicate key as a visible failure. That contradicts the module's own stance in `deserialize-parameter.ts` that parameter pollution "must stay reported" (it is preserved for scalars). Worth either keeping the pollution visible for non-exploded lists too, or a comment documenting the merge as deliberate.

packages/core/src/http-testing/deserialize-parameter.ts
```diff
@@ -0,0 +1,789 @@
...
+  const target = resolvePointer(schema.$ref, scope);
+
+  if (!target) {
+    return undefined;
+  }
+
+  const siblings: ThymianSchema = { ...schema };
+  delete siblings.$ref;
+
+  const merged: ThymianSchema = { ...target, ...siblings };
```
atennert: **[minor]** `{ ...target, ...siblings }` resolves `$ref` siblings by *override*; JSON Schema 2020-12 applies both conjunctively. It only bites when a sibling redeclares `type`/`enum` against the target (near-contradictory schemas), but then conversion decisions and Ajv semantics diverge exactly where this module promises they don't. A comment acknowledging the limit would be enough.

packages/core/src/http-testing/deserialize-parameter.ts
```diff
@@ -0,0 +1,789 @@
...
+  delete merged.allOf;
+
+  for (const member of node.allOf) {
+    const flat = flattenSchema(member, nodeRoot, depth + 1);
+
+    if (!flat) {
+      continue;
+    }
+
+    merged.items ??= flat.items;
```
atennert: **[minor]** The `allOf` merge keeps the *first* member's schema per structural keyword (`??=`, and spread precedence for `properties`), so `allOf: [{items: {minimum: 1}}, {items: {type: 'integer'}}]` types items off the typeless arm → no conversion → avoidable `must be integer`. Top-level `type` intersects correctly via `schemaTypes`; `items`/`properties` collapse to one arm. Same-keyword members could be merged as `{allOf: [...]}` the way `propertySchema` already does for multiple `patternProperties` matches.

packages/core/src/http-testing/deserialize-parameter.ts
```diff
@@ -0,0 +1,789 @@
...
+  return deserializeItems(items, raw, schema, serializationStyle.explode);
+}
+
+export function deserializePathParameter(
+  raw: string,
+  schema: ThymianSchema | undefined,
+  serializationStyle?: SerializationStyle,
+  decode: (item: string) => string = (item) => item,
+): DeserializeResult {
+  return deserializeSimple(decode(raw), schema, serializationStyle, () =>
```
atennert: **[minor]** This closure discards the value `deserializeSimple` passes to `split(value)` and splits the captured *encoded* `raw` instead. Correct today only because a path is never `string[]` — the header branch (`raw.flatMap((line) => split(line))`) would silently split the wrong string for any future caller. The `split: (value: string) => string[]` signature is a trap; make the closure honor its argument.

packages/core/src/http-testing/deserialize-parameter.ts
```diff
@@ -0,0 +1,789 @@
...
+
+/**
+ * Split a still-encoded wire value on a delimiter, then decode each item.
+ * Splitting before decoding is what keeps `a%2Cb` one item rather than two.
+ * `trim` drops the optional whitespace RFC 9110 §5.6.1 permits in header lists.
+ */
+export function splitWireList(
+  raw: string,
+  decode: (item: string) => string,
+  { trim = false }: { trim?: boolean } = {},
```
atennert: **[minor]** No call site passes `trim: true` — headers go through `splitHeaderList`, which trims itself — so the option is dead code and its RFC 9110 justification documents a use that never happens. Drop it, or wire the header path through it.

packages/core/src/http-testing/deserialize-parameter.ts
```diff
@@ -0,0 +1,789 @@
...
+ */
+export function deserializeQueryParameter(
+  items: string[],
+  schema: ThymianSchema | undefined,
+  serializationStyle: SerializationStyle = QUERY_DEFAULT_STYLE,
+): DeserializeResult {
+  const { style, explode } = serializationStyle ?? QUERY_DEFAULT_STYLE;
+  const raw = items.length === 1 ? (items[0] as string) : items;
+
+  if (style !== 'form') {
```
atennert: **[minor]** Both arms of this branch end in the byte-identical `deserializeItems(items, raw, schema, explode)`; only the `isStructural` gate differs. The duplication invites the two paths to diverge on the next edit — `if (style !== 'form' && isStructural(schema)) return unsupported(...)` plus a single return is the honest shape.

packages/core/src/http-testing/validate/validate-request-path-parameters.ts
```diff
@@ -9,9 +13,13 @@ function extractPathParameters(
...
 ): Record<string, string> | undefined {
-  const pathWithoutQuery = actualPath.split('?')[0] ?? actualPath;
+  // Neither the query string nor a `#` fragment is part of the path.
+  const pathWithoutQuery =
+    actualPath.split('?')[0]?.split('#')[0] ?? actualPath;
   const normalizedTemplate = templatePath.replaceAll(/{([^}]+)}/gi, ':$1');
-  const matchFn = match(normalizedTemplate);
+  // `decode: false` keeps the value percent-encoded, so a delimiter inside an
+  // item (`/users/a%2Cb`) survives splitting; each item is decoded afterwards.
+  const matchFn = match(normalizedTemplate, { decode: false });
```
atennert, line 18: **[minor — untested behavior change]** With `decode: false`, representations downstream changed: on an undecodable escape (`/users/%zz`) `decodePathComponent` falls back to the **encoded** string, which now reaches Ajv (previously `path-to-regexp` decoded during matching). Probably fine, but nothing pins it — a malformed-escape test would.

packages/core/test/http-testing/validate/validate-request-query-parameters.test.ts
```diff
@@ -59,3 +64,364 @@ describe('validateExistingQueryParameter', () => {
...
+
+    const results = validateRequestQueryParameters(
+      '/api?role=admin&level=3',
+      request,
+    );
+
+    expect(results.filter((r) => r.type === 'assertion-failure')).toEqual([]);
+  });
+
+  it('splits a non-exploded array before decoding', () => {
```
atennert, line 391: **[test — AC3 gap]** This is the only end-to-end `form`/`explode: false` test, and it can't fail: `?ids=a%2Cb` with `maxItems: 1` passes identically whether or not `parseQueryParameters` splits at all (one item either way). The unit test covers typing of a *pre-split* list, so the split itself is untested where it lives. Add a literal-comma case (`?ids=1,2` against `items: {type: 'integer'}` → zero failures) and a single-element non-exploded case.

packages/core/test/http-testing/validate/validate-request-query-parameters.test.ts
```diff
@@ -59,3 +64,364 @@ describe('validateExistingQueryParameter', () => {
...
+        actual: 'abc',
+      }),
+    );
+  });
+
+  it('still enforces numeric bounds after deserialization', () => {
+    const request = requestWithQueryParameters({ year: yearParameter });
+
+    const results = validateRequestQueryParameters(
+      '/api/userreports?year=2100',
```
atennert, line 142: **[nit — AC1 says "accurate messages"]** The `?year=2100` case asserts the failure names the parameter but never that it names the `maximum`/2027 bound — the assertion would pass just as happily on a wrong-type message. One `expect(...).toContain('maximum')`-style line closes it.

atennert (approved):
LGTM

packages/core/src/http-testing/deserialize/style/form.ts
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * `deepObject` is handled by `deserializeDeepObject`, because reconstructing it
+ * needs the `name[prop]` keys the caller alone can see.
+ */
+export function deserializeQueryParameter(
+  items: string[],
+  schema: ThymianSchema | undefined,
+  serializationStyle: SerializationStyle = QUERY_DEFAULT_STYLE,
+): DeserializeResult {
+  const { style, explode } = serializationStyle ?? QUERY_DEFAULT_STYLE;
```
atennert, line 35: The default is already set on the function parameter. Why is it set here again?

packages/core/src/http-testing/deserialize/style/form.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Reconstruct an object parameter the caller has already gathered from its
+ * constituent query keys — `name[prop]=v` for `deepObject`, or bare `prop=v`
+ * keys for `form`/`explode: true`. Both are only defined when exploded.
+ */
+export function deserializeObjectParameter(
+  properties: [string, string | string[]][],
+  schema: ThymianSchema | undefined,
+  serializationStyle: SerializationStyle = QUERY_DEFAULT_STYLE,
+): DeserializeResult {
+  const { style, explode } = serializationStyle ?? QUERY_DEFAULT_STYLE;
```
atennert, line 59: The default is already set on the function parameter. Why is it set here again?

atennert (approved):

packages/core/src/http-testing/deserialize/style/form.ts
BaggersIO: We'll check the entire code base against code smells in a later step.

PR #380, opened 2026-08-25
feat(core): load TypeScript rules and rule sets directly via jiti

chaosmonster:
## Summary

Adds the shared resolver/loader **seam** (`resolveUserModule` / `loadUserModule`) and wires the rule loader onto it, so a TypeScript rule or single-file rule set loads **directly** from `thymian.config.yaml` and `--rule-set` — no build step, no user Node flags — while the JavaScript path (every built-in rule, every run) stays byte-identical and **never instantiates jiti**.

Implements the normative §4 resolution contract of the epic:

- **Bare specifier → installed package, JS only.** Two `require.resolve` anchors (user project → core install), `exports`/`main` then a conditions-aware resolver (`resolve.exports`, `['node','import','default']`) so ESM-only/dual packages find their JS entry. A package resolving to `.ts`/`.mts`/`.cts`, or to a Node builtin, is declined with a framed error; jiti is never reached.
- **Local specifier → explicit path + explicit extension.** `.ts` → jiti; `.js`/`.mjs`/`.cjs` → native `import()`. `.mts`/`.cts` unsupported; `.d.ts` never loadable regardless of contents. No extension guessing, no `<cwd>/<specifier>` fallback.
- `resolveUserModule` returns a discriminated `{ ok }` result and **never throws**; a path is canonicalised once via `realpathSync.native`; module bodies execute exactly once per canonical path under concurrency, with no bespoke cycle detection.
- The shared `unloadableReason` / `miscasedExtension` predicates are intra-package only — not re-exported from the public `@thymian/core` surface.

The jiti transpile-cache location is pinned to a predictable `os.tmpdir()` path rather than inheriting jiti's environment-dependent default.

## Testing

- **Unit + mutation coverage** for the seam and the rule loader — each acceptance criterion was verified to fail under its named mutation (jiti-spy tests catch the "green against the wrong module" trap that a plain "does it load" test misses inside vitest).
- **External built-loader harness** in `e2e-tests`: installs the built `@thymian/core` from a local Verdaccio registry into a fresh `node_modules` outside the monorepo and loads a `.ts` rule through it in a plain `node` subprocess (no vitest transform); deleting the built output fails the harness.
- Bare-specifier fixtures are built at runtime in a temp `node_modules` (no committed `node_modules`).
- `nx build core`, `nx test core` (all passing), `nx lint core` (0 errors), and `nx e2e e2e-tests` all green.

## Notes for reviewers

- Diff is contained to `packages/core/src/rules/**` + `e2e-tests/**` (+ core `package.json`/lockfile). The plugin load site (`common-cli`) and rule-set glob semantics are separate follow-up stories.
- Adds `jiti` and `resolve.exports` as direct `@thymian/core` dependencies; `engines.node` stays `>=22`.

Story: thymianofficial/thymian-internal#729


chaosmonster:
@claude review

matthyk (commented):
Review of the resolver rewrite. I read the full diff, then verified the resolver's behaviour empirically by running this branch's `load-user-module.ts` against purpose-built package fixtures in a throwaway worktree.

**Headline:** replacing `require.resolve` with a hand-rolled resolver drops four resolution behaviours that worked before. Three are reproduced with fixtures (CJS-only packages, deep imports into non-`exports` packages, extensionless/directory `main`); the fourth (Yarn PnP / resolver hooks) is read from the implementation and flagged inline as unverified. Separately, the committed symlink fixture looks like it will turn the Windows CI job red, and the jiti cache location is a shared-host security concern.

Inline comments below. Two minor notes that didn't warrant their own threads:

- **`load-user-module.ts:63`** — `miscasedExtension` is unreachable for the case it documents. `unloadableReason` rejects any non-lowercase extension first, so requesting `./ts-rule.rule.TS` (real file `.ts`) reports *"does not have a loadable extension ... (found \".TS\")"* rather than the casing message. Only the inverse (request `.ts`, on-disk `.TS`) can reach it. The unit tests call `miscasedExtension` directly, so they never exercise the reachable path and the asymmetry stays invisible.
- **`load-user-module.ts:283`** — the comment says passing `conditions` explicitly pins the set against future resolve.exports default drift, but resolve.exports treats `conditions` as purely **additive**: it builds `new Set(['default', ...opts.conditions])` and then unconditionally adds `import`/`require` and `node`/`browser` unless `unsafe: true`. So `['node','import','default']` adds exactly zero conditions over the default, and a future added default *would* be inherited silently — the opposite of what the comment promises. Actually pinning needs `unsafe: true`.

And two test-hygiene points, raised here rather than inline since they're arguably repo-convention calls:

- **`test/rules/load-user-module-surface.test.ts`** imports `@thymian/core`, which resolves to `dist/index.js`. `nx.json`'s `targetDefaults.test` has no `dependsOn: ['build']` and there's no vitest alias, so on a fresh clone or worktree `nx test core` fails with `ERR_MODULE_NOT_FOUND` on `dist` — a failure that looks like a regression in the change under test. Worse, a *stale* `dist` makes the "public surface does not export `unloadableReason`" assertion pass for the wrong reason, so it can't catch the leak it exists to catch.
- **`e2e-tests/src/user-module-loader-external.test.ts`** — the second test `rmSync`s `node_modules/@thymian/core/dist`, destroying the fixture built once in `beforeAll`, so the file only works if the two `it`s run in declaration order (any shuffle, `.concurrent`, or a retry of the first test breaks it). Also `execFileSync('npm', ...)` has no `npm.cmd` branch, unlike `global.setup.ts` which explicitly does `const npmCmd = isWindows ? 'npm.cmd' : 'npm'`.

packages/core/test/rules/fixtures/rules/counter-link.rule.mjs
```diff
@@ -0,0 +1 @@
+counter.rule.mjs
```
matthyk: **This will fail the Windows CI job.**

`.github/workflows/ci.yaml`'s `main` matrix includes `windows-2022` and runs `npx nx affected:test`. On Windows runners git defaults to `core.symlinks=false`, so this is checked out as a 16-byte **text file** containing `counter.rule.mjs`. `realpathSync.native` then returns the link file itself and `expect(viaSymlink.path).toBe(direct.path)` fails; if it got past that, importing the file evaluates `counter.rule.mjs` as a bare expression -> `ReferenceError`.

`git ls-tree -r` says this is the repo's first and only mode-120000 entry, so there's no existing precedent showing Windows CI tolerates symlinks here.

Creating the symlink in a temp dir during test setup — and skipping when `symlinkSync` throws `EPERM` — keeps AC6 covered without committing a symlink.

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,458 @@
...
+export type UserModuleResolution =
+  { ok: true; path: string } | { ok: false; reason?: string };
+
+const LOADABLE_EXTENSIONS = new Set(['.ts', '.js', '.mjs', '.cjs']);
+const JS_EXTENSIONS = new Set(['.js', '.mjs', '.cjs']);
+const BUILTIN_MODULES = new Set(builtinModules);
+
+/**
+ * The one shared loadable-extension / declaration-file predicate (§4.6).
+ * Consumers (the rule-set glob filter, the plugin loader) call this rather
```
matthyk: **The docstring names two consumers that don't call this — and the rule-set glob in particular still has the bug this seam was meant to fix.**

Neither named consumer calls `unloadableReason`:

- **`rule-loader.ts:249`** — `loadRuleSet` globs and calls `loadRules` on every match with no filter. So a rule set whose `pattern` matches a generated `.d.ts` sitting next to its `.js` still aborts the **entire** rule-set load with a `RuleLoadError` rather than skipping the file. That's the concrete user-facing behaviour this seam was introduced to fix, and it's unchanged by this PR.
- **`common-cli/src/base-cli-run-command.ts:519`** — the plugin loader still does `require.resolve(location)` + `import(pathToFileURL(...))`. Plugins get none of the guards and none of the `.ts`/jiti support, so there are now two divergent resolution contracts in the codebase while this comment asserts they've been unified.

Either wire the consumers up or scope the docstring to what it does today — as written it will mislead the next person who assumes the check is centralized.

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,458 @@
...
+    return { ok: false };
+  }
+
+  const casingReason = miscasedExtension(requestedAbsolute, canonical);
+
+  if (casingReason) {
+    return { ok: false, reason: casingReason };
+  }
+
+  return { ok: true, path: canonical };
```
matthyk, line 161: **`unloadableReason` is applied to the requested path but never to the realpath, so `ok: true` can name an unloadable target.**

`resolveLocal` guards the requested path, then resolves the symlink and returns `canonical` without re-running the guard on it. Verified against this branch:

- `./aliasd.ts` (symlink -> `decl.d.ts`) -> `{ ok: true, path: '.../decl.d.ts' }`
- `./alias.ts` (symlink -> `real.mts`) -> `{ ok: true, path: '.../real.mts' }`
- a **directory** named `dirnamed.js` -> `{ ok: true, path: '.../dirnamed.js' }`

The first two are then caught downstream by `loadUserModule` — but as `UserModuleLoadError`, a different error name and `ref` from every other rule-load failure, so the same user mistake reports inconsistently depending on whether a symlink was involved. The directory case passes **both** guards and surfaces a raw `ERR_UNSUPPORTED_DIR_IMPORT` / `MODULE_NOT_FOUND` — exactly the unframed surprise this seam exists to prevent.

It also affects any consumer that trusts `ok: true` (the docstring names the glob filter): it will happily include unloadable paths.

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,458 @@
...
+// "Not installed" (not-found) stays distinct from "installed but broken"
+// (present, but package.json is not valid JSON) — the latter is reported, not
+// silently treated as "not found" (§4.4).
+function locatePackage(
+  anchoredRequire: NodeJS.Require,
+  packageName: string,
+): LocatePackageResult {
+  // resolve.paths returns the node_modules dirs to search, nearest first
+  // (null only for builtins/relative/absolute, which never reach here).
+  const searchPaths = anchoredRequire.resolve.paths(packageName) ?? [];
```
matthyk: **Same root cause as the `exports`/`main` comments, one more consequence: this bypasses every Node resolver hook.**

`locatePackage` walks `resolve.paths()` and does `existsSync(<nodeModulesDir>/<name>/package.json)`. That only finds packages laid out in a real `node_modules` tree. Yarn PnP has no such tree — it works by patching `Module._resolveFilename`, which the old `require.resolve` honoured and a raw `existsSync` cannot. A PnP project with `ruleSets: ['@thymian/rules-rfc-9110']` would get `Cannot resolve rule source @thymian/rules-rfc-9110.` for a package that is genuinely installed. Same class of breakage for any custom loader or resolver hook.

**Caveat:** unlike the other resolver findings on this PR, I did not stand up an actual PnP project to confirm this end-to-end — it's read off `locatePackage`'s implementation. Worth confirming before deciding whether PnP is in scope at all.

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,458 @@
...
+    // `exports` may be a string, an object, or an array of fallbacks — all
+    // three are valid and all three are handled by resolve.exports (an
+    // object-only guard would silently skip the array form).
+    try {
+      // Conditions are passed explicitly to pin the normative §4.1 set
+      // (`['node','import','default']`) rather than leaning on the library's
+      // current default, so a future resolve.exports default change cannot
+      // silently drift the contract.
+      const matches = resolveExports.exports(pkg, entry, {
+        conditions: ['node', 'import', 'default'],
```
matthyk: **Regression: CJS-only rule packages no longer load.**

The condition set omits `require`, so a package whose `exports` map only offers a `require` condition is rejected outright. Reproduced against this branch with a fixture `cjs-exports-pkg` (`"exports": { ".": { "require": "./index.cjs" } }`):

> Cannot load rule source cjs-exports-pkg: "cjs-exports-pkg" is installed but its "exports" map does not provide ".": No known conditions for "." specifier

Before this PR, `require.resolve` resolved through the `require` condition and the result was then loaded with `import(pathToFileURL(...))` — bypassing the `exports` map entirely — so CJS rule packages worked. Every published CJS-only rule package breaks on upgrade.

Restoring prior behaviour means accepting `require` as a fallback condition, which is what the old `require.resolve` path effectively did. That's a deliberate deviation from pure ESM resolution semantics and worth an explicit decision either way — but dropping support for every CJS rule package shouldn't happen by accident.

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,458 @@
...
+    } catch (error) {
+      return {
+        ok: false,
+        reason: `"${specifier}" is installed but its "exports" map does not provide "${entry}": ${(error as Error).message}`,
+      };
+    }
+  }
+
+  if (!relativeTarget) {
+    if (subpath) {
```
matthyk: **Regression: deep imports into a package without an `exports` map are now unresolvable.**

This gives up whenever the package has no `exports` map, but `require.resolve` resolved those subpaths fine. Reproduced with `deep-pkg` (`main: "./index.js"`, real file at `node_modules/deep-pkg/rules/foo.js`):

- **before:** `loadRules('deep-pkg/rules/foo.js')` loaded the rule
- **now:** `Cannot resolve rule source deep-pkg/rules/foo.js.` — and since `reason` is unset on this path, the user gets no explanation at all

Pointing a `ruleSets` entry at a single rule file inside a non-`exports` package is a reasonable thing to want, and it's now impossible. Legacy resolution here should join `root` + `subpath` and stat it, the way Node does.

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,458 @@
...
+      return { ok: false };
+    }
+
+    relativeTarget =
+      typeof pkg.main === 'string' && pkg.main.length > 0
+        ? pkg.main
+        : 'index.js';
+  }
+
+  const absoluteTarget = path.resolve(root, relativeTarget);
```
matthyk: **Regression: the hand-rolled `main` resolution drops extension guessing and directory-index resolution.**

`path.resolve(root, relativeTarget)` treats `main` as a literal file path, but Node's legacy resolution tries `<main>`, then `<main>.js`, then `<main>/index.js`. Both of the following are valid, `require.resolve`-loadable packages that this branch declines — reproduced:

- `main: "./index"` -> `"legacy-main-pkg" is installed but its resolved entry "./index" does not exist on disk.`
- `main: "./lib"` with `lib/index.js` present -> `"dir-main-pkg" resolved to "lib", which is not a loadable JavaScript file`

Both shapes are common in older packages.

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,458 @@
...
+}
+
+// jiti transpile-cache location (epic #725 §7 — a required, documented
+// decision, not an inherited default): jiti's own default prefers
+// `<cwd>/node_modules/.cache/jiti` when that directory already exists and
+// falls back to the OS tmp dir otherwise — an environment-dependent split.
+// We pin it to a single, predictable location under the OS tmp dir instead,
+// so transpiled user rule/plugin source never lands inside the user's own
+// project tree regardless of what happens to exist on disk.
+const JITI_FS_CACHE_DIR = path.join(tmpdir(), 'thymian-jiti-cache');
```
matthyk: **Pinning the jiti cache to a fixed path in the shared OS tmp dir is a cache-poisoning vector on multi-user hosts.**

On Linux this is `/tmp/thymian-jiti-cache` — world-writable and shared between users. jiti's cache filename is `<parentDirName>-<basenameNoExt>.<hash(filename)>.mjs`, and it accepts cached content whenever it ends with ` /* v<jitiVersion>-<hash(source,16)> */\n` (both in `node_modules/jiti/dist/jiti.cjs`). Both components are derivable by anyone who knows the victim's rule path and its source — which for a published rule package is public. An attacker with an account on a shared build host pre-creates the directory and drops a poisoned entry; the next `thymian lint` executes their JS as the victim. A pre-planted symlink at that filename also just gets overwritten by jiti's `writeFileSync`.

The comment above justifies this as avoiding an environment-dependent split, but jiti's default *prefers* `<project>/node_modules/.cache/jiti` — inside the user's own permission domain — and only falls back to tmp. This override removes that preference unconditionally, trading a predictable location for a shared-writable one.

If a fixed location is the goal, a per-user path keeps the predictability without the shared-write exposure (e.g. under `os.homedir()`, or tmp plus uid created with `mode: 0o700`).

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,458 @@
...
+export async function loadUserModule(canonicalPath: string): Promise<unknown> {
+  const reason = unloadableReason(canonicalPath);
+
+  if (reason) {
+    throw new ThymianBaseError(reason, {
+      name: 'UserModuleLoadError',
+      suggestions: [
+        'Rules, rule sets, and plugins must be a .ts, .js, .mjs, or .cjs file — not .d.ts, .mts, or .cts.',
+      ],
+      ref: 'https://thymian.dev/references/errors/user-module-load-error/',
```
matthyk: This link 404s. `astro-docs/src/content/docs/references/Errors/` contains no `user-module-load-error.md` and this PR doesn't add one. Every other error name in the repo (`rule-load-error`, `invalid-rule-error`, ...) has a matching page.

Either add the page or reuse `rule-load-error` here — a user hitting a `.d.ts`/`.mts` rule follows this link and lands on a 404.

packages/core/src/rules/rule-loader.ts
```diff
@@ -293,29 +289,39 @@ export async function loadRules(
     ).flat();
```
matthyk, line 296: **Drive-by, pre-existing — not introduced by this PR.** (The line in question is 286, which falls outside this diff, so I'm anchoring to the nearest line in the hunk.)

The array branch recurses with five of `loadRules`' six parameters:

```ts
input.map((entry) =>
  loadRules(entry, ruleFilter, options, cwd, ruleProfiles),
),
```

`profileConfig` is dropped and defaults back to `{}`. Any caller reaching this branch with a non-empty `profileConfig` silently loses every profile-level severity/type override for those entries — `applyProfileThenConfig` then applies an empty map.

Latent today because only the top level passes arrays, but the omission is invisible and will bite the first caller that threads profiles through an array input. Since this PR is already rewriting the body of this function, might be worth fixing in passing.

chaosmonster:
Thanks for the thorough review — this was exactly the kind of empirical checking that caught the real problems. Addressed in 54579730.

The root cause behind most of the resolver findings was that the seam **reimplemented package resolution** (`resolve.paths` + `existsSync` + `resolve.exports`) instead of using Node's own. We've reverted to **native `require.resolve`** for bare specifiers (two anchors: user project → core), layering only the loadable-set policy on top (resolved file must be built `.js`/`.mjs`/`.cjs`; unbuilt `.ts` source declined). `resolve.exports` is dropped as a dependency. This restores native behaviour and fixes several findings at once:

- **CJS-only packages** (require-condition `exports`) load again — `require.resolve` uses the `require` condition.
- **Legacy `main`** extension/index resolution (`main:"./index"`, `main:"./lib"`) and **deep subpaths** into non-`exports` packages resolve the way Node does.
- **Yarn PnP / resolver hooks** are honoured again (`require.resolve` goes through `Module._resolveFilename`, not a raw `existsSync` walk). *(I didn't stand up a full PnP project either, but native resolution is the mechanism PnP patches, so it's back in the path.)*
- The restricted-`exports` (no `./package.json`) and array-`exports` cases stay fixed — `require.resolve` handles both natively.

Other findings:

- **`resolveLocal` now guards the realpath, not just the requested path**, and rejects directories — a symlink to a `.d.ts`/`.mts`, or a directory named `x.js`, is declined at resolve time (consistently framed) instead of surfacing a raw `ERR_UNSUPPORTED_DIR_IMPORT` or a differently-named error downstream.
- **jiti cache** moved off the world-writable OS tmp dir to a per-user `homedir()/.cache/thymian/jiti` — closes the cache-poisoning vector on shared hosts while keeping the path predictable.
- **`user-module-load-error` 404** — the error now points at the existing `rule-load-error` page; the dedicated page ships with the docs work (725.5).
- **`profileConfig` dropped in the array branch** — now threaded through.
- **Surface test** asserts against the `src/index.js` barrel instead of the built `@thymian/core`, so it no longer needs a prior build.
- **Committed symlink** — the AC6 canonicalisation symlink is now created at runtime (skipped where the platform refuses), and the committed `counter*` fixtures are removed. New regression tests cover CJS-only, legacy `main`, directory-named-`*.js`, and symlink-to-`.d.ts`.

Intentionally **out of scope** for this story (flagging rather than silently ignoring):

- **Rule-set glob still aborts on a non-loadable match** rather than skipping it — the "skip non-loadable, framed per-match" glob semantics are Story 725.3 (§5); this PR is scoped to the single-file path. The seam's shared predicate is deliberately exported so 725.3 wires the glob filter onto it (docstring corrected to say "will", not "does").
- **Plugin loader** (`common-cli`) still uses the old pair — wiring it onto the seam is Story 725.2. (The docstring no longer claims it's already unified.)
- **Import-only ESM packages** (exposing *only* an `import` condition, no `require`/`default`) don't resolve via `require.resolve` — this matches pre-PR behaviour. If import-only ESM rule packages need to load, that's a deliberate follow-up (an `import.meta.resolve` fallback), worth its own decision.

All 371 core unit tests pass; lint clean.

e2e-tests/src/user-module-loader-external.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    execFileSync(
+      'npm',
+      [
+        'install',
+        `@thymian/core@${version}`,
+        '--registry',
+        process.env.npm_config_registry ?? '',
+        '--no-save',
+      ],
```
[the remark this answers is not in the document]
chaosmonster: Fixed in 954703ed — the harness now fails fast in `beforeAll` when either `THYMIAN_E2E_VERSION` or `npm_config_registry` is unset, and passes the resolved `registry` value rather than a possibly-empty string.

packages/core/src/rules/rule-loader.ts
```diff
@@ -293,29 +289,36 @@ export async function loadRules(
...
-    resolved = require.resolve(location);
-  } catch {
-    throw new ThymianBaseError(`Cannot resolve rule source ${input}.`, {
-      name: 'RuleLoadError',
-      ref: 'https://thymian.dev/references/errors/rule-load-error/',
-    });
+  if (!resolution.ok) {
+    throw resolution.reason
+      ? new ThymianBaseError(
+          `Cannot load rule source ${input}: ${resolution.reason}.`,
```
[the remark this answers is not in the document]
chaosmonster: Fixed in 954703ed — the wrapper now strips a trailing period from the seam's `reason` before appending its own, and a regression test asserts the message contains no `..`.

packages/core/src/load-user-module.ts
```diff
@@ -0,0 +1,458 @@
...
+      // silently drift the contract.
+      const matches = resolveExports.exports(pkg, entry, {
+        conditions: ['node', 'import', 'default'],
+      });
+
+      relativeTarget = matches?.[0];
+    } catch (error) {
+      return {
+        ok: false,
+        reason: `"${specifier}" is installed but its "exports" map does not provide "${entry}": ${(error as Error).message}`,
```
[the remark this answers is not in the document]
chaosmonster: Resolved in 54579730 — the bare-specifier path was rewritten to use native `require.resolve` and no longer calls `resolve.exports` at all, so this `(error as Error).message` formatting path is gone (and `resolve.exports` is dropped as a dependency).

packages/core/test/rules/load-user-module-surface.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    // a relative path — that is the entire point of this test, which checks
+    // what an external consumer sees through `dist/index.js`, not what the
+    // internal module graph exposes.
+    // eslint-disable-next-line @nx/enforce-module-boundaries
+    const importPublicCore = () => import('@thymian/core');
+    const publicSurface: Record<string, unknown> = await importPublicCore();
+
+    expect(publicSurface).not.toHaveProperty('unloadableReason');
+    expect(publicSurface).not.toHaveProperty('miscasedExtension');
+  });
```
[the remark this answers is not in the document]
chaosmonster: Fixed in 54579730 — the surface test now imports the `src/index.js` barrel instead of `@thymian/core`, so it no longer depends on a prebuilt `dist/`. Since the public surface is exactly what `src/index.js` re-exports, the source-barrel check is equivalent and build-free.

packages/core/test/rules/fixtures/rules/counter-link.rule.mjs
chaosmonster: Fixed in 54579730 — the committed `counter.rule.mjs` / `counter-link.rule.mjs` (the repo's only mode-120000 entry) are removed. The AC6 canonicalisation test now writes the file and creates the symlink in a temp dir at runtime, and skips when `symlinkSync` throws `EPERM`/`ENOSYS`/`EACCES` (Windows without the privilege). Good catch on why Windows passed by luck before.

packages/core/src/load-user-module.ts
chaosmonster: Fixed the docstring in 54579730 — it no longer claims the consumers call the predicate today; it now says the rule-set glob filter (725.3) and plugin loader (725.2) *will* share it, neither being wired here. Those two consumers are deliberately out of scope for 725.1: the rule-set glob's skip-non-loadable-match semantics are Story 725.3 (§5), and the plugin loader in `common-cli` is Story 725.2. So the divergence is real but tracked, not asserted-away.

packages/core/src/load-user-module.ts
chaosmonster: Fixed in 54579730 — `resolveLocal` now re-runs `unloadableReason` on the realpath and rejects non-files. So `./alias.ts`->`real.d.ts`, `./alias.ts`->`real.mts`, and a directory named `x.js` are all declined at resolve time with the consistent framed reason (surfacing as `RuleLoadError` via the caller), not as a later `UserModuleLoadError` or a raw `ERR_UNSUPPORTED_DIR_IMPORT`. New regression tests cover the directory and symlink-to-`.d.ts` cases.

packages/core/src/load-user-module.ts
chaosmonster: Fixed in 54579730 — bare-specifier resolution is back on native `require.resolve` (which goes through `Module._resolveFilename`, the hook PnP patches), so PnP and other resolver hooks are honoured again; the `existsSync`-walk is gone. Same caveat as yours: I didn't stand up a real PnP project, but the mechanism PnP relies on is back in the path.

packages/core/src/load-user-module.ts
chaosmonster: Fixed in 54579730 — native `require.resolve` resolves through the `require` condition, so CJS-only packages load again. Added a `cjs-only-pkg` fixture (`exports: { ".": { require: "./index.cjs" } }`) with a regression test asserting it resolves.

packages/core/src/load-user-module.ts
chaosmonster: Fixed in 54579730 — with native `require.resolve` back, `deep-pkg/rules/foo.js` resolves exactly as it did before (Node's legacy subpath resolution for a non-`exports` package), and the `reason`-less silent failure is gone. Note on the agreed rule set: it lists "no deep imports" for node_modules, but rather than add code to actively *block* deep subpaths I've kept it to plain native behaviour (deep files resolve for non-`exports` packages, and are gated by `exports` for modern ones). If you want deep imports hard-rejected, that's an extra guard — happy to add it, but flagging it as a deliberate call rather than doing it silently.

packages/core/src/load-user-module.ts
chaosmonster: Fixed in 54579730 — native `require.resolve` does Node's own `main` resolution, so `main:"./index"` -> `index.js` and `main:"./lib"` -> `lib/index.js` both load again. Added a `legacy-main-pkg` (`main:"./index"`) fixture + regression test.

packages/core/src/load-user-module.ts
chaosmonster: Fixed in 54579730 — the cache is moved off the world-writable OS tmp dir to a per-user `homedir()/.cache/thymian/jiti`. That keeps the location predictable (the reason it was pinned) while staying inside the user's own permission domain, so another user can't pre-seed a poisoned entry. Thanks for the concrete exploit write-up — it's what made the severity clear.

packages/core/src/load-user-module.ts
chaosmonster: Fixed in 54579730 — `UserModuleLoadError` now points at the existing `rule-load-error` page instead of the missing `user-module-load-error` one, so the link resolves. The dedicated page is part of the docs/error-reference work in Story 725.5.

packages/core/src/rules/rule-loader.ts
chaosmonster: Fixed in passing in 54579730 — the array branch now threads `profileConfig` through the recursive `loadRules` call, so profile-level overrides are no longer dropped for array inputs.

packages/core/test/rules/load-rules.test.ts
```diff
@@ -3,12 +3,70 @@ import { join } from 'node:path';
...
+        /Cannot load rule source unbuilt-ts-pkg: "unbuilt-ts-pkg" ships unbuilt TypeScript source/,
+      );
+      // The reason already ends in a period; the wrapper must not double it up.
+      expect(error.message).not.toMatch(/\.\./);
+    } finally {
+      fixtures.cleanup();
+    }
+  });
+
+  it('AC3: reports a .d.ts specifier as unloadable, never as a missing default export', async () => {
```
matthyk:
```suggestion
  it('reports a .d.ts specifier as unloadable, never as a missing default export', async () => {
```

matthyk (changes requested):
Please remove all references to specific ACs, stories, etc.

matthyk (approved):

matthyk (approved):

PR #381, opened 2026-08-25
fix(plugin-openapi): build req.path with a URL joiner instead of node:path

matthyk:
Fixes thymianofficial/thymian-internal#621.

## The bug

`openapi.processor.ts` built `req.path` with `join` from `node:path`. Both operands are **URL** path segments — `basePath` comes from `new URL(...).pathname` — so on Windows this produced `\v1\pets`.

It surfaces now because epic 575 commits a generated `request-types.d.ts` whose `Path` union is built from these strings, and `sampler sync --check` is a CI gate. A checked-in artifact that differs by platform is not a gate. CI's matrix includes `windows-2022`.

## The real defect was two joiners for one job

The package computed "basePath + path" two different ways — `node:path`'s `join` for the value that goes on the graph, and a private `joinUrls` for filter matching. They disagreed in three ways:

| input | `join` (posix) | `joinUrls` |
|---|---|---|
| `/v1` + `/a/../b` | `/v1/b` (resolves) | `/v1/a/../b` |
| `/v1` + `/pets` on win32 | `\v1\pets` | `/v1/pets` |
| `a` + `///b` | `a/b` | `a//b` (regex doesn't rescan) |

So this adds one `joinUrlPath` helper used at **both** call sites and deletes `joinUrls`. `.` and `..` are deliberately **not** resolved: a spec path is an operation template, not a file location, and rewriting `/a/../b` would change which operation a selector names.

The two genuine filesystem joins (`search-for-openapi-files.ts`, `load-openapi.ts`) are untouched.

## Testing a platform bug from POSIX

On POSIX `join` and `posix.join` are identical, so no local test can fail before the fix. Three layers instead:

1. The extracted helper is unit-tested directly — platform-independent by construction. Run against the two implementations it replaces **and** against `win32.join` as a stand-in: 35 failures across the three legacy joiners, 24/24 passing on the new one.
2. An ESLint `no-restricted-imports` rule blocks `node:path` under `src/processors/**` and in `url-path.ts` itself. Verified it fires.
3. A no-backslash assertion on a produced `req.path` — inert on POSIX, earns its keep on the `windows-2022` job.

## Note

One deliberate canonicalisation: `('', 'pets')` now yields `/pets` where both old joiners returned `pets`. `req.path` is an origin-form request target and now a checked-in identity key, so it should always be rooted; OpenAPI requires `paths` keys to begin with `/`, so no conforming document is affected.

## Verification

`plugin-openapi:test` 119 · `run-many -t test` 17 projects · `run-many -t lint` 18 projects, 0 errors. `packages/core` and `packages/plugin-sampler` untouched.

### DIFF
--- a/packages/plugin-openapi/eslint.config.mjs
+++ b/packages/plugin-openapi/eslint.config.mjs
@@ -1,7 +1,24 @@
 import baseConfig from '../../eslint.config.mjs';
 
+const noNodePathMessage =
+  "Only URL paths are handled here, and node:path is platform dependent (it joins with '\\' on Windows) and resolves '.'/'..' segments. Use joinUrlPath from src/url-path.ts instead (see thymian-internal#621).";
+
 export default [
   ...baseConfig,
+  {
+    files: ['**/src/processors/**/*.ts', '**/src/url-path.ts'],
+    rules: {
+      'no-restricted-imports': [
+        'error',
+        {
+          paths: [
+            { name: 'node:path', message: noNodePathMessage },
+            { name: 'path', message: noNodePathMessage },
+          ],
+        },
+      ],
+    },
+  },
   {
     files: ['**/*.json'],
     rules: {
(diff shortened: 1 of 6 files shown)


atennert (approved):
LGTM

PR #382, opened 2026-08-25
fix(plugin-openapi): skip x- extensions and reject non-integer status keys

matthyk:
Fixes thymianofficial/thymian-internal#634.

## The bug

OpenAPI 3.x permits Specification Extensions in the Responses Object, so this is a **conforming** document:

```yaml
responses:
  '200': { description: OK, content: {...} }
  x-internal-note: reviewed by the API guild
```

`responses-object.processor.ts` treated every non-`default`, non-range key as a status code:

```ts
const statusCodeNumber = +statusCode;                     // +'x-internal-note' → NaN
if (statusCodeNumber < 100 || statusCodeNumber > 599) { throw ... }
```

The guard never fires, because **both comparisons are false for `NaN`**. A response node is created with `statusCode: NaN`.

Pre-existing: `sampler init` used to write a directory literally named `@GET/NaN` and exit 0. Silent garbage. Epic 575 renders every transaction as a selector, so the same document now aborts the load — a louder and better failure, but it makes the underlying bug worth fixing rather than tolerating. The user-visible symptom is a workspace that `thymian lint`s clean (it loads with `emitFormat: false`) and dies on `thymian test`.

## The fix

1. **Skip `^x-` keys**, with the `continue` placed *before* reference resolution. That also removes a second failure mode on conforming input: `x-linked-response: {$ref: '#/components/responses/DoesNotExist'}` previously threw "Could not resolve internal response reference" — an extension value is not a Response Object and should never have been dereferenced.
2. **`!Number.isInteger(statusCodeNumber)`** added to the range guard, so nothing non-integral reaches a response node whatever the key shape. This also closes `'200.5'`, which passed the old `100..599` check.
3. A mis-cased range key (`2xx`) gets an appended corrective sentence rather than silently becoming `NaN`.

## On lowercase `2xx`: reject, and `packages/core` is untouched

`isHttpStatusCodeRange` stays `/^[12345]XX$/`. Thymian's own bundled `@scalar/openapi-parser` already rejects `2xx` — the OAS 3.1 meta-schema matches response keys as `^[1-5](?:[0-9]{2}|XX)$` plus extensions. Accepting it in the transform would make the transform **more permissive than the validator in the same binary**, which is the same split-brain this issue complains about, pointed the other way. The same schema explicitly allows `x-` keys, so "accept `x-`, reject `2xx`" is one rule — follow the OAS schema — not two ad-hoc choices.

## Verification

New `responses-object.processor.test.ts` — **6 of 10 cases fail against the pre-fix source**, 10/10 after. End-to-end: the `x-internal-note` document went from 3 nodes (one with `statusCode: NaN`, serialising as `"statusCode":null`) to 2.

`plugin-openapi:test` 101 · `run-many -t test` 17 projects / 1440 tests · `run-many -t lint` 18 projects, 0 errors.

Two files changed; `packages/core` and `packages/plugin-sampler` untouched.

### DIFF
--- a/packages/plugin-openapi/src/processors/responses-object.processor.ts
+++ b/packages/plugin-openapi/src/processors/responses-object.processor.ts
@@ -9,6 +9,22 @@ import {
 } from './response-object.processor.js';
 import type { Parameters } from './utils.js';
 
+/**
+ * OpenAPI Specification Extension field names. Extensions are explicitly
+ * allowed in the Responses Object and carry metadata, not responses, so they
+ * must never be interpreted as status codes.
+ *
+ * @see https://spec.openapis.org/oas/v3.1.0#specification-extensions
+ */
+const specificationExtensionPattern = /^x-/;
+
+/**
+ * A status code range written in a case other than the uppercase form the
+ * OpenAPI Specification requires (e.g. `2xx` instead of `2XX`). Only used to
+ * produce a better error message — such keys are rejected, not accepted.
+ */
+const misCasedStatusCodeRangePattern = /^[1-5]xx$/i;
+
 export function processResponsesObject(
   responsesObject: OpenApiV31.ResponsesObject | undefined,
   parameters: Parameters,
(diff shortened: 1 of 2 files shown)


atennert (approved):
LGTM

PR #385, opened 2026-08-26
feat(core): flatten, bound, and frame rule-set glob loading

chaosmonster:
## Summary

Story 725.3 (epic thymianofficial/thymian-internal#725 §5), stacked on #383 (Story 725.2).

Rewrites the `ruleSet.pattern` glob branch of `loadRuleSet` in
`packages/core/src/rules/rule-loader.ts` so that a glob rule set is flat, bounded, and
framed:

- `node_modules` is excluded from every glob, and a rule set's own file is excluded from
  its own matches (AC1).
- A glob match that is itself a rule set is a framed error instead of nesting — recursion
  is impossible by construction, no visited-set (AC2).
- A non-loadable-kind match (wrong extension, `.d.ts`, broken symlink, non-regular file)
  is skipped with a framed reason (via `process.emitWarning`) while the rest of the set
  still loads (AC3).
- A match that fails while loading (syntax error, vanished between glob and read) fails
  the whole set with a framed, file-named error (AC4).
- A glob that matched files but produced zero rules throws, instead of silently returning
  an empty set; matches still load in deterministic sort order (AC5).

Extracts `resolveEligibleRule` as the single per-rule pipeline now shared by the inline
`ruleSet.rules` branch, the glob branch, and the top-level single-file `loadRules` path
(previously duplicated three ways).

Diff containment per the story: only `packages/core/src/rules/rule-loader.ts` +
`packages/core/test/rules/**`, plus a 2-line `eslint.config.mjs` exclusion for one
fixture file that is genuinely-invalid JS (needed for a real AC4 syntax-error test — an
ESLint-parseable "syntax error" fixture would be vacuous per epic §9).

## Test plan

- [x] `nx build core`
- [x] `nx test core` — 382/382 passing (39 files)
- [x] `nx lint core` — 0 errors (14 pre-existing warnings, unrelated)
- [x] `nx typecheck core` — clean
- [x] All 5 ACs mutation-tested by hand (no Stryker configured): 9 named mutations across
      the 5 ACs, each confirmed red, reverted, confirmed green — logged in the story's Dev
      Agent Record.
- [x] `git diff` against base stays within the stated containment (verified via
      `git diff origin/story/725-2...HEAD --stat`).

### DIFF
--- a/packages/core/src/rules/rule-loader.ts
+++ b/packages/core/src/rules/rule-loader.ts
@@ -257,10 +257,10 @@ function describeError(error: unknown): string {
 // The glob filter's own guard, on top of the seam's `unloadableReason`
 // (extension/declaration-file predicate — reused, never hand-copied):
 // classifies a match by filesystem kind so a directory/FIFO/socket/broken
-// symlink is *skipped* (AC3) rather than attempted. A path that is not on
-// disk at all is deliberately left unclassified here (`undefined`) — that is
-// not "the wrong kind", it is a match that vanished between the glob call and
-// the load attempt below, which must fail the whole set (AC4), not skip.
+// symlink is *skipped* rather than attempted. A path that is not on disk at
+// all is deliberately left unclassified here (`undefined`) — that is not
+// "the wrong kind", it is a match that vanished between the glob call and
+// the load attempt below, which must fail the whole set, not be skipped.
 function nonLoadableGlobMatchReason(resolved: string): string | undefined {
   const extensionReason = unloadableReason(resolved);
 
@@ -371,8 +371,8 @@ async function loadRuleSet(
           canonical = undefined;
         }
 
-        // AC1: the rule set's own file is excluded from its own matches
-        // (trivial self-match), and never counted toward "anything matched".
+        // The rule set's own file is excluded from its own matches (trivial
+        // self-match), and never counted toward "anything matched".
         if (canonical !== undefined && canonical === basePath) {
           continue;
         }
--- a/packages/core/src/rules/rule-loader.ts
+++ b/packages/core/src/rules/rule-loader.ts
@@ -261,11 +261,27 @@ function describeError(error: unknown): string {
 // all is deliberately left unclassified here (`undefined`) — that is not
 // "the wrong kind", it is a match that vanished between the glob call and
 // the load attempt below, which must fail the whole set, not be skipped.
-function nonLoadableGlobMatchReason(resolved: string): string | undefined {
+function nonLoadableGlobMatchReason(
+  resolved: string,
+  canonical: string | undefined,
+): string | undefined {
   const extensionReason = unloadableReason(resolved);
 
   if (extensionReason) {
     return extensionReason;
+  }
+
+  // A symlink can have a loadable *spelling* (e.g. `alias.rule.ts`) while its
+  // realpath target is an unloadable kind (`real.d.ts`, `.mts`/`.cts`). We load
+  // through the canonical path, and `loadUserModule` would throw on it — which
+  // would fail the whole set. Classify it as a skip here instead, per AC3, so
+  // it is warned-and-skipped like a directly-matched unloadable file.
+  if (canonical !== undefined) {
+    const canonicalExtensionReason = unloadableReason(canonical);
+
+    if (canonicalExtensionReason) {
+      return canonicalExtensionReason;
+    }
   }
 
   let entryStat;
@@ -386,7 +402,7 @@ async function loadRuleSet(
 
         anyMatched = true;
 
-        const skipReason = nonLoadableGlobMatchReason(resolved);
+        const skipReason = nonLoadableGlobMatchReason(resolved, canonical);
 
         if (skipReason) {
           warnSkippedGlobMatch(resolved, skipReason, ruleSet.name);
--- a/packages/core/src/rules/rule-loader.ts
+++ b/packages/core/src/rules/rule-loader.ts
@@ -1,8 +1,13 @@
+import { lstatSync, realpathSync, statSync } from 'node:fs';
 import * as path from 'node:path';
 
 import { glob } from 'tinyglobby';
 
-import { loadUserModule, resolveUserModule } from '../load-user-module.js';
+import {
+  loadUserModule,
+  resolveUserModule,
+  unloadableReason,
+} from '../load-user-module.js';
 import { ThymianBaseError } from '../thymian.error.js';
 import { isRecord } from '../utils.js';
 import { validate } from './ajv-validate.js';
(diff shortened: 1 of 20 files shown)


packages/core/src/rules/rule-loader.ts
```diff
@@ -199,6 +204,116 @@ function applyProfileThenConfig(
...
+      `Skipping "${resolved}" matched by rule set "${ruleSetName}": ${reason}`,
+      {
+        name: 'RuleLoadError',
+        suggestions: [
+          'Remove or fix the file so it matches a loadable rule, or narrow the glob pattern to exclude it.',
+        ],
+        ref: 'https://thymian.dev/references/errors/rule-load-error/',
+      },
+    ),
+  );
```
matthyk: What's the reason for the `process.emitWarning` call here?

packages/core/src/rules/rule-loader.ts
```diff
@@ -237,28 +348,109 @@ async function loadRuleSet(
...
+        `Rule set "${ruleSet.name}" pattern matched files but produced no loadable rules.`,
+        {
+          name: 'RuleLoadError',
+          suggestions: [
+            'Check that the glob pattern matches rule files whose default export is a rule.',
+          ],
+          ref: 'https://thymian.dev/references/errors/rule-load-error/',
+        },
+      );
+    }
```
matthyk, line 490: the rule loading logic is duplicated to prevent nested rule sets, right?

packages/core/src/rules/rule-loader.ts
```diff
@@ -237,28 +348,109 @@ async function loadRuleSet(
...
+
+        if (skipReason) {
+          warnSkippedGlobMatch(resolved, skipReason, ruleSet.name);
+          continue;
+        }
+
+        let rawModule: unknown;
+
+        try {
+          rawModule = await loadUserModule(resolved);
```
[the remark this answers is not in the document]
chaosmonster: Good catch — fixed in 14ff6a11. The per-match pipeline now loads through the canonical path (`const loadPath = canonical ?? resolved; await loadUserModule(loadPath)`), so a symlinked spelling no longer executes the file twice or bypasses the seam's exactly-once cache. User-facing messages stay on the non-canonical `resolved`, and the `canonical === undefined` case (realpath failed = vanished between glob and read) still loads `resolved` and fails the whole set as before. Added a regression test that loads the same `.ts` rule via a symlink and its target and asserts the module executes once (non-vacuous — verified red without the fix; native ESM canonicalizes symlinks on its own, jiti does not).

packages/core/src/rules/rule-loader.ts
```diff
@@ -237,28 +348,109 @@ async function loadRuleSet(
...
+      throw new ThymianBaseError(
+        `Rule set "${ruleSet.name}" pattern matched files but produced no loadable rules.`,
+        {
+          name: 'RuleLoadError',
+          suggestions: [
+            'Check that the glob pattern matches rule files whose default export is a rule.',
+          ],
+          ref: 'https://thymian.dev/references/errors/rule-load-error/',
+        },
+      );
```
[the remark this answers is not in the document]
chaosmonster: Fixed in 14ff6a11. Added an `anyRuleLoaded` flag set whenever a matched module's default export `isRule(candidate)` — i.e. before `ruleFilter` runs — and the throw is now `anyMatched && !anyRuleLoaded` instead of `rules.length === 0`. A glob set whose rules are all excluded by the caller's filter (e.g. `rules list` with a strict severity threshold) now returns empty, matching the inline `ruleSet.rules` branch, while a glob that matched only non-rule files still throws. Regression test added.

packages/core/src/rules/rule-loader.ts
chaosmonster: It surfaces AC3's non-fatal diagnostic: a match of the wrong filesystem kind (`.d.ts`, directory, broken symlink, non-regular file) is skipped while the rest of the set still loads, so it must be a warning, not a thrown error. `process.emitWarning` rather than `logger.warn` because `loadRules`/`loadRuleSet` are free functions in the loader seam with no logger threaded through them — they're called from multiple entry points (including direct plugin calls), before any Thymian/logger instance exists. It's the standard Node mechanism for a non-fatal, structured diagnostic without coupling the loader to a logger or writing to console directly, and the tests observe it via `process.on('warning')`. Happy to thread a logger through instead if you'd prefer the codebase's `logger.warn` convention here — that'd be a broader signature change across the loader entry points, so I left it as-is.

packages/core/src/rules/rule-loader.ts
chaosmonster: Right. The glob branch deliberately does *not* recurse through `loadRules`/`loadRuleSet` for each match — that would let a matched rule set load as a nested set. Instead it inlines a flattened single-rule load (`loadUserModule` → `assertHasDefaultExport` → `isRuleSet` ⇒ framed throw / `isRule` ⇒ `resolveEligibleRule`), so a glob-matched rule set is a framed error rather than nesting. That's AC2: recursion is impossible by construction, no visited-set needed. The genuinely shared per-rule logic (config/profile overrides → type-declaration check → filter → execution-invariant check) is already factored out into `resolveEligibleRule`, which the inline, glob, and single-file paths all call; what looks duplicated is just the load-and-classify framing that has to differ because the glob path forbids nesting.

chaosmonster:
All review comments have been addressed in 14ff6a11 (pushed).

**Copilot findings (fixed + resolved):**
- **Canonical glob load** — each glob match now loads through its canonical realpath (`canonical ?? resolved`), so a symlinked spelling no longer re-executes the rule file or bypasses `loadUserModule`'s exactly-once cache. The realpath-failed (vanished) case still fails the whole set. User-facing messages stay on `resolved`.
- **Filter-aware zero-rules throw** — the "matched files but produced no loadable rules" throw now keys off `anyRuleLoaded` (whether any match was a loadable `Rule`, before `ruleFilter`) rather than post-filter `rules.length`, so a glob set whose rules are all excluded by the caller's filter returns empty — matching the inline `ruleSet.rules` branch.

Both got regression tests (the symlink test is a `.ts` rule so it's non-vacuous — verified red without the fix).

**@matthyk questions** — replied inline on both (rationale for `process.emitWarning`, and confirming the glob path inlines a flattened load to forbid nested rule sets by construction). Left those threads open for you.

Gates: `nx build/test/lint/typecheck core` all green — 384 tests pass, 0 lint errors.

packages/core/test/rules/load-rules-glob.test.ts
```diff
@@ -0,0 +1,204 @@
...
+      expect(rules[0]?.meta.name).toBe('glob-skip-kinds-valid');
+
+      // Give the emitted warning a tick to be delivered.
+      await new Promise((resolve) => setImmediate(resolve));
+
+      expect(warnings).toHaveLength(1);
+      expect(warnings[0]).toMatchObject({
+        name: 'RuleLoadError',
+        message: expect.stringMatching(/declaration\.rule\.d\.ts/),
+      });
```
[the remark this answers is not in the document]
chaosmonster: Good point — fixed in c09a07d5. The test now filters `warnings` down to the `RuleLoadError` whose message names `declaration.rule.d.ts` and asserts that subset has length 1, instead of asserting on the global warning count. An unrelated process warning during the run no longer affects it.

packages/core/src/rules/rule-loader.ts
```diff
@@ -237,28 +348,128 @@ async function loadRuleSet(
...
+        }
+
+        anyMatched = true;
+
+        const skipReason = nonLoadableGlobMatchReason(resolved);
+
+        if (skipReason) {
+          warnSkippedGlobMatch(resolved, skipReason, ruleSet.name);
+          continue;
+        }
```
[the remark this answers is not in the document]
chaosmonster: Good catch — real AC3 gap. Fixed in c9a06dca. `nonLoadableGlobMatchReason` now also applies `unloadableReason` to the canonical path (when realpath succeeded): a symlink with a loadable spelling (`alias.rule.ts`) pointing at an unloadable target (`real.d.ts` / `.mts`/`.cts`) is now warned-and-skipped rather than attempted-and-fatal. Added a regression test (symlink → `.d.ts`, alongside a valid rule) asserting the set still loads the valid rule and emits exactly one skip warning naming the target — verified red without the fix.

packages/core/src/rules/rule-loader.ts
```diff
@@ -199,6 +204,132 @@ function applyProfileThenConfig(
...
+  // it is warned-and-skipped like a directly-matched unloadable file.
+  if (canonical !== undefined) {
+    const canonicalExtensionReason = unloadableReason(canonical);
+
+    if (canonicalExtensionReason) {
+      return canonicalExtensionReason;
+    }
+  }
+
+  let entryStat;
```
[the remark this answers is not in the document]
chaosmonster: Applied in ca49f3a7 — `entryStat` is now `let entryStat: ReturnType<typeof lstatSync>`.

One clarification on the premise: this did not actually fail typechecking. `let entryStat;` followed by a definite assignment is an "evolving `any`", which `noImplicitAny` permits (it errors on implicit-any parameters and never-assigned locals, not on evolving-let) — `nx typecheck core` was clean before and after. The explicit annotation is still worth it for readability/intent, so applied regardless.

packages/core/src/rules/rule-loader.ts
matthyk: I think we should avoid using it. It could confuse the users. We should use the logger for it.

matthyk (changes requested):
Besides the `process.emitWarning`, the PR looks good to me.

packages/core/src/rules/rule-loader.ts
chaosmonster: Agreed — done in e995b35f. Skipped glob matches now warn through the caller's `logger.warn` instead of `process.emitWarning` (which surfaced as a raw Node warning with a stack trace and reads as an internal fault). I threaded an optional `logger?: Logger` through `loadRules` → `loadRuleSet` → `warnSkippedGlobMatch` (following the existing optional-logger pattern in `finding-render.ts`) and pass `this.logger` from the three `Thymian` call sites. The message stays self-contained (reason + remedy + docs reference) so it keeps the same framing the thrown errors have, just without the stack noise. Tests now observe the skip via a capturing `Logger` rather than `process.on('warning')`. (This does reach slightly beyond the story's stated diff containment into `thymian.ts`, which is unavoidable to thread the logger to the call sites.)

matthyk (approved):

PR #387, opened 2026-08-27
feat: `thymian report diff` — compare two Thymian reports and write out the difference

atennert:
Implements **thymianofficial/thymian-internal#743** (story 502.1 of epic thymianofficial/thymian-internal#502): a `thymian report diff` command that compares two persisted Thymian JSON reports — a base and a head — and writes out a structured diff document, so regressions **and improvements** between two runs are visible (before/after a change, or in CI gates).

> **Stacked PR** on #362 (`story/507-report-merge`) — the base branch of this PR. The diff command builds on the `thymian:` report loader, `core.report.convert` claim path, and ADR-0020 introduced there. Once #362 merges, this branch gets rebased onto `main` and the PR retargeted. Until then, this PR's diff shows only the three report-diff commits.

```sh
thymian report diff \
  --base thymian:./report-before.json \
  --head thymian:./report-after.json \
  --fail-on regression
```

## What's in here

- **core**: `ReportDiff` document model + Ajv schema (`report-diff.ts` — `diffId`, `createdAt`, base/head identity, `changes` as a discriminated union of run-result / specification / rule changes); pure diff modules `report-diff-compute.ts` (run-result identity = run type + tool name + rule id + location + failure details, set semantics per side; rule/ruleset union diff) and `report-diff-endpoints.ts` (format-graph endpoint pairing by method+path natural key, slot-based response pairing, changed-aspect deltas); `Thymian.reportDiff()` loads each side via **one `core.report.convert` emission per side** (per-side loading keeps attribution and dodges the cross-input `runId` dedup, which is wrong for diff) and never routes through report emission — a diff must not side-effect-write report files.
- **common-cli**: the first-colon `<type>:<location>` parser extracted into shared `parseTypedInput()` (`report-flag`/`spec-flag`/`traffic-flag` were three verbatim copies — deferred-work item picked up here); `--fail-on` gate classification (`report-diff-gate.ts`) and the human summary renderer (`report-diff-summary.ts`, terminal-rendering ownership rule).
- **plugin-reporter**: the fragment `report` tag carries a `reportCount`, closing the run-less-sibling loophole in the exactly-one-report-per-side gate.
- **thymian CLI**: `report diff` command — required `--base`/`--head` (Thymian-only in v1, usage error names the restriction), `--fail-on none|regression|error|any-change`, JSON document emission plus the compact summary.
- **e2e**: `cli-report-diff.test.ts` — identical inputs → empty diff, improvements-only wording, regression informational by default vs. exit 1 under opted-in `--fail-on regression`, **two inputs with different `thymianFormat` hashes** producing specification changes (the case #507 never exercised e2e), unsupported input type → exit 2.
- **docs**: "Diffing reports" tutorial + cross-link from the merge tutorial; **ADR-0021** (diff verb placement: per-side loading via `core.report.convert`, diff computed in core, diff document as a non-`Report` output, `--fail-on` gates diff changes — not ADR-0015's report classification).

## Semantics (review-ratified 2026-08-26)

- **`--fail-on` defaults to `none`** — the diff is informational; gating is opt-in. `regression` fails on any added run-result, `error` only on added run-results with resolved severity `error`, `any-change` on any change at all. Improvements never fail the regression/error gates.
- **`runType` and `run.tool.name` are deliberate identity components** — a failure moving between modes, or the same failure reported by an additional tool, is a distinct fact.
- **Set semantics per side** — the diff compares distinct findings, not occurrence counts (documented v1 semantics; count-based identity deferred).
- No fuzzy endpoint matching — a path rename is removed+added.

## Verification

- Adversarial code review round 1 (2026-08-26): all findings addressed in `f4927b41` — incl. the HIGH slot-based response/transaction pairing fix, per-side-consistent endpoint key scheme, per-graph request scoping, gate-aware improvement verdict, and the `reportCount` gate; two items deferred as tracked issues (thymianofficial/thymian-internal#744, thymianofficial/thymian-internal#745).
- Unit: all 17 project test targets green.
- E2E: **92/92** (full suite, Verdaccio-published CLI).


matthyk:
Closed and rework with the new workspace setup

PR #388, opened 2026-08-27
docs(repo): documentation, error-reference pages, and migration notes for TS loading (725.5)

chaosmonster:
## Summary

- Adds the normative `Loading Rules and Plugins` reference page (loadable set, bare-vs-local specifier contract, what isn't checked at load, flat rule sets, glob behaviour, jiti cache location) with its embedded `## Migration` section.
- Resolves the `.rule.js`/`.rule.ts` contradiction in `how-to-use-rules.md` without a blanket replace, and reframes the `tsc --noEmit` troubleshooting step now that loading and type-checking are independent.
- Updates `creating-new-rules.md` and `plugin-developers-guide.md` to lead with TypeScript and document the no-build-step / `erasableSyntaxOnly` behaviour.
- Adds `plugin-load-error.md` and `user-module-load-error.md`, extends `rule-load-error.md` with the multi-cause convention, and repoints `UserModuleLoadError`'s `ref` (the one permitted `packages/**` change — a single line plus its stale comment in `load-user-module.ts`).
- Adds `error-ref-pages.test.ts`, a durability guard asserting every `ref` a `packages/**` source names has a matching `Errors/<slug>.md` page.

**Base branch note:** this PR targets `story/725-4-the-generator-emits-only-the-loadable-set`, not `main` — as of this story's creation, 725.2 (#383), 725.3 (#385), and 725.4 (#386) are all still open and stacked. Documenting from `main` alone would describe behaviour (seam-routed plugins, flat/framed globs, generator extension steering) that doesn't exist there yet. Rebase onto `main` once the stack merges.

Closes half 1 of #633 (the `plugin-load-error` reference page — half 2, the `ref` on the throw, already landed in 725.2).

## Test plan

- [x] `nx run-many -t build` — 17/17 projects green
- [x] `nx run-many -t test` — 17/17 projects green (core 384, common-cli 293, thymian 101, astro 179)
- [x] `nx build astro` — `starlightLinksValidator` passes, all internal links resolve
- [x] `error-ref-pages.test.ts` verified red when either new error page is deleted, green on restore
- [x] `git diff <base> --stat` — touches only `astro-docs/**` plus the single `ref` line in `packages/core/src/load-user-module.ts`
- [x] Every quoted decline message in the story's Dev Notes was grepped against actual source on this branch and matches verbatim


astro-docs/test/error-ref-pages.test.ts
```diff
@@ -0,0 +1,83 @@
...
+const PRE_EXISTING_GAPS = new Set([
+  'action-timeout-error',
+  'invalid-action-input-error',
+]);
+
+const REF_PATTERN =
+  /https:\/\/thymian\.dev\/references\/errors\/([a-z0-9-]+)\//g;
+
+const PACKAGES_ROOT = join(__dirname, '../../packages');
+const ERRORS_DIR = join(__dirname, '../src/content/docs/references/Errors');
```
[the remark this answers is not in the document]
chaosmonster: Fixed in b52d6e9 (test now derives `__dirname` from `import.meta.url` + `fileURLToPath`/`dirname`).

Worth noting the test was actually green before this — Vitest v4 injects a `__dirname`/`__filename` shim into ESM test files, so it never threw at runtime. But relying on that shim in a `type: "module"` package is exactly the fragility you flagged, so the `import.meta.url` form is the right call regardless.

astro-docs/src/content/docs/references/loading-rules-and-plugins.md
```diff
@@ -0,0 +1,100 @@
...
+- **`.ts`** — loaded via [jiti](https://github.com/unjs/jiti).
+- **`.js` / `.mjs` / `.cjs`** — loaded via native `import()`.
+
+**`.mts` and `.cts` are not supported.** `.d.ts` is **never** loadable, whatever it contains —
+a declaration file is not runnable source, regardless of its extension matching otherwise.
+
+## Bare vs local specifiers
+
+Every specifier you pass to `--rule-set`, `--plugin`, a `ruleSets:` entry, or a plugin `path:`
+is exactly one of two kinds, decided syntactically:
```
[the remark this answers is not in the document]
chaosmonster: Fixed in b52d6e9. You caught a real contradiction, though the resolution turned out to be the reverse of the assumed one: a `--rule-set`/`ruleSets:` value is **never** a glob. It always resolves to a single rule-set module as a bare or local specifier (see `toRuleSetInputs` in `merge-inputs.ts` → `resolveUserModule`, and the `ruleSets` config schema, which only accepts a package name or `{ name, profile }`).

Globs live one level down: the `pattern:` field a rule-set module *exports* (`RuleSet.pattern` in `rule-set.ts`), matched against its member rule files in `loadRuleSet`. So the "exactly one of two kinds" statement was correct; the **Globs** section was the imprecise one for attributing globs to the top-level value.

I reconciled both sections instead of adding an exception: the Globs heading now attributes the glob to the rule set's `pattern:` field, and bare-vs-local gained a short "Globs are not a third specifier kind" note cross-linking the two.

astro-docs/src/content/docs/references/Errors/rule-load-error.md
```diff
@@ -4,8 +4,46 @@ title: 'RuleLoadError'
...
+### A rule module does not export its rule using a default export
+
 A rule module was loaded but does not export its rule using a default export. Thymian rules and rule sets must be exported using `export default` (ES modules) or `module.exports =` (CommonJS).
 
+### The rule source specifier could not be resolved
+
+`Cannot resolve rule source <input>.`
+
+The specifier passed to `--rule-set` (or a `ruleSets:` entry) is neither an installed package
+nor a local file Thymian could find.
```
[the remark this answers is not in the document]
chaosmonster: Valid — fixed in 30f95785. The resolve-failure paragraph now separates the two: a `ruleSets:` config entry is a rule-set **package** name (per the config schema), while a `--rule-set` value is a package **or** a local path with an explicit extension. The "neither an installed package nor a file on disk" wording now applies only to the local `--rule-set` case.

astro-docs/src/content/docs/guides/HTTP rules/how-to-use-rules.md
```diff
@@ -17,7 +17,7 @@ Create a `thymian.config.yaml` file in your project root:
 
 ```yaml
 ruleSets:
-  - './rules/**/*.rule.js' # Local rules
+  - './rules/**/*.rule.ts' # Local rules
   - '@thymian/rules-rfc-9110' # Npm package
```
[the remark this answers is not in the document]
chaosmonster: Valid — fixed in 30f95785. The `thymian.config.yaml` example now uses a package name (`@thymian/rules-rfc-9110`) for `ruleSets:`; local rules are shown via `--rule-set ./rules/my-api.rule.ts`, and the twin `rules list --rule-set ./rules/**/*.rule.ts` glob example was corrected to a single explicit-extension path. Added a note that a bare glob is not a valid `ruleSets:` entry — globbing is a rule set's own `pattern:` field.

astro-docs/src/content/docs/guides/HTTP rules/creating-new-rules.md
```diff
@@ -32,6 +32,20 @@ import { httpRule } from '@thymian/core';
...
 
+Point the CLI at the generated `.ts` file and it runs — **no build, no bundler, no Node
+flags.** `thymian generate rule` only ever emits a file that loads: `.ts` by default, `.cjs`
+under `--cjs`, and **never** `.mts`/`.cts`. If `--output` names an extension that conflicts
+with the mode (e.g. `--cjs --output foo.js`), the command declines rather than silently
+rewriting your path.
+
+:::note
+`erasableSyntaxOnly` is **not** required of your rule code. That restriction belongs to Node's
+own type-stripping feature, which Thymian's loader does not use — loading goes through jiti
```
matthyk:
```suggestion
own type-stripping feature, which Thymian's loader does not use — loading goes through [jiti](https://github.com/unjs/jiti)
```

astro-docs/src/content/docs/guides/HTTP rules/how-to-use-rules.md
```diff
@@ -17,17 +17,25 @@ Create a `thymian.config.yaml` file in your project root:
...
 
 plugins:
   '@thymian/plugin-http-linter': {}
 ```
 
+`ruleSets:` entries are rule-set **package names**. To lint against rules that live in your own
+repository, pass them to `--rule-set` (below) with an explicit-extension path, or bundle them
+into a rule set that globs them through its [`pattern:` field](#rule-set-entry-point) — a bare
+glob is not a valid `ruleSets:` entry.
+
```
matthyk: Is this really true? I thought that I can also pass a relative path in the thymian config?

astro-docs/test/error-ref-pages.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    expect(missing).toEqual([]);
+  });
+
+  it('names plugin-load-error and user-module-load-error, so this test goes red if either page is removed', () => {
+    const referenced = findReferencedSlugs();
+
+    expect(referenced.has('plugin-load-error')).toBe(true);
+    expect(referenced.has('user-module-load-error')).toBe(true);
+  });
+});
```
matthyk: Why?

astro-docs/src/content/docs/guides/HTTP rules/how-to-use-rules.md
chaosmonster: You're right — corrected in 4b262aaa. I over-corrected in the previous round.

A `ruleSets:` entry flows through the same seam as `--rule-set` (`normalizeRuleSetEntry` → `loadRules` → `resolveUserModule`), so `isLocalSpecifier` accepts `./`, `../`, and absolute paths and a **local path with an explicit extension resolves and loads** — e.g. `./rules/my-api.rule.ts`. So a relative path is valid, exactly as you said.

What genuinely does *not* work as a `ruleSets:` value is a **glob** (`./rules/**/*.rule.ts`): `resolveUserModule` treats it as a literal path, the file doesn't exist, resolution fails. Globbing is a rule set's own `pattern:` field. The old example was a glob, which is why it was broken — but the fix should have kept "or a local path", not narrowed it to packages.

The config example now shows a local rule file alongside the package, and the prose across `how-to-use-rules.md`, `rule-load-error.md`, and `loading-rules-and-plugins.md` now reads package-**or**-local-path, just not a glob.

astro-docs/test/error-ref-pages.test.ts
chaosmonster: Removed the test in 9fe242dc, per our offline discussion. The astro suite is green without it (10 files, 177 tests).

For the record on the "why" it existed: it was AC8 of story 725.5 — a durability guard asserting that every error `ref: https://thymian.dev/references/errors/<slug>/` a `packages/**` source names has a matching `Errors/<slug>.md` page, so a thrown error's doc link never 404s. Dropping it here as agreed; we can revisit that coverage separately if we want it.

matthyk (approved):

PR #389, opened 2026-08-27
feat(plugin-reporter)!: write every report into its own run directory

matthyk:
## ⚠️ Breaking change — read this first

The per-formatter **`path` option is removed.** A config that still sets `formatters.<name>.path` now fails config validation at plugin registration and exits `2` — it is never accepted and ignored.

**Migration**

```yaml
# before
'@thymian/plugin-reporter':
  options:
    formatters:
      markdown: { path: ./report.md }

# after
'@thymian/plugin-reporter':
  options:
    reportsDir: build/reports    # optional; defaults to .thymian/reports
    formatters:
      markdown: {}
```

Where a pipeline read a fixed file, glob `<reportsDir>/*/report.<ext>` instead. **There is no longer a way to pin an exact output file name** — that is a deliberate, discussed trade for a uniform layout.

## The problem

Each formatter hardcoded a constant default path (`.thymian/reports/report.{md,csv,json}`), so every run without an explicit `path` **overwrote the previous run's report**. Output location was also a per-formatter concern, so one run's artifacts could scatter.

## The change

One plugin-level `reportsDir` base. Every report gets its own directory beneath it, named `<createdAt>-<reportId prefix>`, with stable basenames inside:

```
.thymian/reports/2026-08-26T14-16-17-451Z-de3a3f0e/report.md
                                                  /report.csv
                                                  /report.json
```

A run's formats stay together by containment, and `:` never appears in a directory name so it is valid on Windows.

### Fixes a `serve` bug along the way

The destination is resolved on **every** `core.report` rather than pinned to the first. Previously `this.outputPath ??=` / `this.opening ??=` bound a whole session to its first report — so in `thymian serve`, where one plugin instance serves every workflow and flushes only at `core.close`, a second workflow's report was written into the **first** workflow's directory while the rendered document reported the second workflow's timestamp.

### Write lifecycle

Formatters write during `core.report`, which the emitter dispatches **without awaiting**. Each formatter therefore serializes its writes on a queue that `flush()` drains, closing two windows: a report could still be in flight when `core.close` returned and `serve` called `process.exit`, and two overlapping reports could clobber a shared stream handle leaving one file unclosed. Rendering now runs inside the same guarded region as the write, so a malformed report is logged and skipped rather than aborting the run, and the derived directory name is memoized per report object so every formatter agrees even on the fallback path.

### Failure handling, split by when it is knowable

- **Precondition** — when at least one formatter is configured, the `reportsDir` base is created once at registration. An unusable base (read-only mount, EACCES, a path shadowed by a file) raises `UnusableReportsDirectoryError` and exits `2` **before any workflow runs**, instead of logging one line and exiting `0` with no report.
- **In-flight** — failures that can only appear mid-write still degrade to a logged error, because by then the findings exist and aborting would destroy useful output.

There is a comment on the asymmetry so it is not "fixed" into consistency later.

## Verification

| Gate | Result |
|---|---|
| Unit tests | 17/17 projects |
| `plugin-reporter` | 12 files / 117 tests |
| e2e (published CLI via Verdaccio) | 10 files / 70 tests |
| Lint | 0 errors |
| Typecheck | clean |

Also exercised against the real CLI: two runs → two run directories; custom `reportsDir` honoured with nothing written to `.thymian`; unwritable base → exit 2 with findings **not** rendered; legacy `path` → exit 2, nothing written.

## Reviewer notes

- Windows CI is the first real test of the no-colon invariant — all local verification ran on macOS.
- `Plugins/Reporter/options.md` will pick up the new `reportsDir` description on the next docs build; the committed stub is a placeholder and was deliberately not hand-edited.
- Known follow-ups, deliberately out of scope: `.thymian/reports` has no retention policy, and the resolved path is logged at `debug` while the CLI defaults to `warn`, so it is not discoverable without `--verbose`.

### DIFF
--- a/astro-docs/src/content/docs/references/configuration.mdx
+++ b/astro-docs/src/content/docs/references/configuration.mdx
@@ -166,25 +166,89 @@ plugins:
 
 ## Report formatters
 
-`@thymian/plugin-reporter` writes one file per configured formatter. Each formatter accepts an
-optional `path`: a relative path is resolved against `--cwd`, and an absolute path is used as-is.
+`@thymian/plugin-reporter` writes one file per configured formatter. Where those files land is a
+plugin-level setting, not a per-formatter one: `reportsDir` names the base directory — a relative
+path is resolved against `--cwd`, an absolute path is used as-is — and it defaults to
+`.thymian/reports`. Leave it out to take the default; an empty `reportsDir` fails validation
+rather than scattering run directories through the working directory.
+
+Every report gets its own directory beneath that base, and the files inside keep a stable name, so
+consecutive runs never overwrite each other:
+
+```
+<reportsDir>/<created-at>-<report-id>/report.<ext>
+.thymian/reports/2026-08-25T10-30-00-123Z-a1b2c3d4/report.md
+```
+
+`<created-at>` is the report's creation timestamp with `:` and `.` replaced by `-` (no colons, so
+the name is valid on Windows too), and `<report-id>` is the first eight characters of the report
+id. Every formatter in the same run sees the same report, so the markdown, CSV, and JSON output of
+one run lands side by side in that one directory as `report.md`, `report.csv` and `report.json` —
+correlating them is opening the directory.
+
+The rule holds everywhere, including `thymian serve`: one reporter instance serves the whole
+session, but each workflow emits its own report, so each workflow gets its own run directory.
+
+```yaml
+plugins:
+  '@thymian/plugin-reporter':
+    options:
+      # Optional. Defaults to .thymian/reports.
+      reportsDir: build/reports
+      formatters:
+        markdown: {} # build/reports/<created-at>-<report-id>/report.md
+        csv: {} # build/reports/<created-at>-<report-id>/report.csv
+        json: {} # build/reports/<created-at>-<report-id>/report.json
+```
+
+| Formatter  | Output file                                         | Use it for                                             |
+| ---------- | --------------------------------------------------- | ------------------------------------------------------ |
+| `markdown` | `<reportsDir>/<created-at>-<report-id>/report.md`   | Human review, PR comments, CI job summaries            |
+| `csv`      | `<reportsDir>/<created-at>-<report-id>/report.csv`  | Tabular analysis in spreadsheets or BI tools           |
+| `json`     | `<reportsDir>/<created-at>-<report-id>/report.json` | Automation and post-processing of the full report data |
+
+Formatters take no options of their own — `markdown: {}` selects the formatter, nothing more. There
+is no way to pin one fixed output file, because a fixed file is exactly what makes a second run
+destroy the first one's report.
+
+A downstream step therefore matches a pattern rather than a path: a CI job that uploads or reads
+the output should glob `<reportsDir>/*/report.<ext>` — for example `.thymian/reports/*/report.md`.
+When a step needs exactly one match, point `reportsDir` at a directory that starts the run empty
+(`reportsDir: build/reports` in a fresh workspace, or a step that clears it first).
+
+### Migrating from the per-formatter `path`
+
+The per-formatter `path` option was removed. A formatter that still sets it fails config
+validation, because formatter options take no properties at all. To migrate:
+
+1. Delete the `path` key — `markdown: {}` is the whole formatter configuration now.
+2. If it pointed somewhere other than the default, set the plugin-level `reportsDir` to that
+   directory instead. It is one base for every formatter, not a per-formatter file.
+3. Where a pipeline read the fixed file, glob `<reportsDir>/*/report.<ext>` instead.
 
 ```yaml
 plugins:
   '@thymian/plugin-reporter':
     options:
       formatters:
-        markdown: {} # .thymian/reports/report.md
-        csv:
-          path: reports/findings.csv
-        json: {} # .thymian/reports/report.json
+        markdown:
+          path: build/reports/report.md # ❌ before
 ```
 
-| Formatter  | Default path                   | Use it for                                             |
-| ---------- | ------------------------------ | ------------------------------------------------------ |
-| `markdown` | `.thymian/reports/report.md`   | Human review, PR comments, CI job summaries            |
-| `csv`      | `.thymian/reports/report.csv`  | Tabular analysis in spreadsheets or BI tools           |
-| `json`     | `.thymian/reports/report.json` | Automation and post-processing of the full report data |
+```yaml
+plugins:
+  '@thymian/plugin-reporter':
+    options:
+      reportsDir: build/reports # ✅ after
+      formatters:
+        markdown: {} # build/reports/<created-at>-<report-id>/report.md
+```
+
+The reporter also checks that `reportsDir` can be created before the run starts, whenever at least
+one formatter is configured. A base directory it cannot create fails the run with
+[`UnusableReportsDirectoryError`](/references/errors/unusable-reports-directory-error/) rather than
+letting a green run produce no files. With `formatters: {}` nothing is checked and no directory is
+created.
 
 ### When to use the Markdown formatter
 
@@ -230,8 +294,8 @@ report:
 
 The file contains the canonical report payload exactly as Thymian emitted it on the `core.report`
 event, with no presentation transforms applied. The top level is always a **JSON array** of report
-objects, so a session that produces several reports stays complete. A single-workflow run yields a
-one-element array.
+objects, holding the one report that run directory belongs to — a session that produces several
+reports produces several directories, each with its own one-element array.
 
 Locations are stored as **references**, not rendered endpoint strings: a `thymianFormat` location
 carries `elementId`/`elementType` that you resolve against the run's entry in the report's
--- a/astro-docs/src/content/docs/references/configuration.mdx
+++ b/astro-docs/src/content/docs/references/configuration.mdx
@@ -250,6 +250,10 @@ one formatter is configured. A base directory it cannot create fails the run wit
 letting a green run produce no files. With `formatters: {}` nothing is checked and no directory is
 created.
 
+A failure _while_ a report is being written is deliberately different: it is logged as an error on
+stderr and leaves that one file missing, but it does not fail the run. The findings already exist by
+then, so an incomplete report beats losing them.
+
 ### When to use the Markdown formatter
 
 Pick `markdown` when a person or AI agent reads the result — in a pull request, a CI job summary, or locally:
--- a/astro-docs/src/content/docs/references/configuration.mdx
+++ b/astro-docs/src/content/docs/references/configuration.mdx
@@ -213,8 +213,26 @@ destroy the first one's report.
 
 A downstream step therefore matches a pattern rather than a path: a CI job that uploads or reads
 the output should glob `<reportsDir>/*/report.<ext>` — for example `.thymian/reports/*/report.md`.
-When a step needs exactly one match, point `reportsDir` at a directory that starts the run empty
-(`reportsDir: build/reports` in a fresh workspace, or a step that clears it first).
+
+```yaml
+# Upload every run directory the job produced.
+- uses: actions/upload-artifact@v7
+  with:
+    name: thymian-reports
+    path: .thymian/reports/*/report.*
+```
+
+A step that needs exactly _one_ file — posting the Markdown as a PR comment, piping the JSON into
+a script — has two options. Either take the newest run: a run directory name starts with the
+report's creation timestamp, so the names sort chronologically and the last one is the latest run.
+
+```bash
+report="$(ls -1d .thymian/reports/*/ | tail -1)report.md"
+```
+
+Or point `reportsDir` at a directory that is empty when the run starts (`reportsDir: build/reports`
+in a fresh workspace, or a step that clears it first), so the glob can only ever match once. In a
+CI job with a cached or reused workspace, clearing it is a step of its own.
 
 ### Migrating from the per-formatter `path`
 
--- a/packages/plugin-reporter/src/formatter.ts
+++ b/packages/plugin-reporter/src/formatter.ts
@@ -78,6 +78,34 @@ export function analyze(reports: Report[], logger?: Logger): ReportAnalysis {
   };
 }
 
+/**
+ * User-facing configuration shared by every file formatter.
+ *
+ * Deliberately empty: where a formatter writes is not a per-formatter concern
+ * any more. The plugin-level `reportsDir` option picks the base directory and
+ * every report gets its own directory beneath it, so all formats of one run
+ * stay together. This mirrors the reporter plugin's options schema one-to-one —
+ * that schema is `additionalProperties: false` at every level, so nothing may be
+ * added here that a user cannot set, and values the plugin injects belong in
+ * {@link FormatterRuntimeOptions}.
+ */
+export type FileFormatterOptions = Record<never, never>;
+
+/**
+ * Values the reporter plugin injects into a formatter's `init`. Deliberately not
+ * part of {@link FileFormatterOptions}: they are runtime context, not config,
+ * and must never widen the user-facing options schema.
+ */
+export type FormatterRuntimeOptions = {
+  /** Run directory a relative `reportsDir` is anchored to. */
+  cwd?: string;
+  /**
+   * Base directory each report's own run directory is created under. Relative
+   * to `cwd`; unset means `.thymian/reports`.
+   */
+  reportsDir?: string;
+};
+
 export interface Formatter<Options = Record<PropertyKey, unknown>> {
   options: Options;
 
--- a/packages/plugin-reporter/src/formatters/csv.ts
+++ b/packages/plugin-reporter/src/formatters/csv.ts
@@ -55,6 +55,12 @@ export class CsvFormatter implements Formatter<CsvFormatterOptions> {
   /**
    * Content of the most recently written report, handed back by {@link flush}
    * so a caller that drives a single report still gets the rendered output.
+   *
+   * No production consumer: the reporter plugin discards `flush()`'s return
+   * value. It exists for the {@link Formatter} contract and for callers — tests
+   * today — that drive one report and assert on the rows. Bounded to one
+   * document on purpose; retaining every report is what this formatter used to
+   * do to stream one session-level file.
    */
   private lastOutput: string | undefined;
 
@@ -150,7 +156,7 @@ export class CsvFormatter implements Formatter<CsvFormatterOptions> {
     // only reliable once the stream has been ended: an async write error (a
     // failed header, say) surfaces after `stream.write` has already returned.
     if (written !== undefined && !open.failed) {
-      this.logger.debug(`Wrote CSV report to ${open.outputPath}`);
+      this.logger.info(`Wrote CSV report to ${open.outputPath}`);
       this.lastOutput = `${CSV_HEADER}${written}`;
     }
   }
--- a/packages/plugin-reporter/src/formatters/csv.ts
+++ b/packages/plugin-reporter/src/formatters/csv.ts
@@ -12,14 +12,17 @@ import {
   walkExecutions,
 } from '@thymian/core';
 
-import type { Formatter } from '../formatter.js';
+import type {
+  FileFormatterOptions,
+  Formatter,
+  FormatterRuntimeOptions,
+} from '../formatter.js';
+import { resolveReportPath } from '../report-file-name.js';
 
 const CSV_HEADER =
   'run_id,run_type,tool,rule_id,location,row_type,status,severity,finding_kind,finding_id,title,message,detail\n';
 
-export type CsvFormatterOptions = {
-  path: string;
-};
+export type CsvFormatterOptions = FileFormatterOptions;
 
 function executionLabel(
   execution: Execution,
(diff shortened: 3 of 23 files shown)


Markus-Ende (commented):
Reviewed against `main` with the pre-change code side by side, and verified the schema behaviour by running the reporter options schema through the repo's own ajv instance.

The write-lifecycle work here is the strongest part and stands on its own: `core.report` is dispatched without being awaited, so serializing writes on a queue that `flush()` drains closes a real window before `serve`'s `process.exit`. The fail-fast/degrade split — hard failure at registration, logged degradation once findings exist — is the right asymmetry, and the comment explaining it should keep it from being "fixed" into consistency later. Naming coverage is thorough (colon-free for Windows, `NAME_MAX` clamping, no trailing dash, `WeakMap` memoization so all three formatters agree on the fallback path).

Two questions and a handful of findings.

## Question: is losing the custom file name intended?

As of this branch there is no way to control the output file name at all — not just no way to pin an exact path. `REPORT_BASENAME` is hardcoded, the directory is derived from the report, and formatter options are the empty type `Record<never, never>` behind `additionalProperties: false` with zero declared properties, so `path`, `fileName`, anything, fails validation. There is no CLI substitute either (no `--report-file`/`--report-out`).

The PR describes this as "a deliberate, discussed trade for a uniform layout", so this may be exactly the intent — but it is worth stating explicitly, because it removes a capability rather than relocating it. If the intent was only to stop runs from overwriting each other, that lives entirely in the **default**: every formatter defaulted to a constant `.thymian/reports/report.{md,csv,json}`, while a user who set `path` had explicitly opted into a fixed name. Making the default a per-run directory fixes the bug without removing the option — the branch is one line in `resolveReportPath` (`options.path ? resolve(cwd, options.path) : <run-directory layout>`).

The strongest argument *for* removal is that in `serve` a fixed path becomes last-write-wins, plus one layout to document instead of two. That is a legitimate product call, and at `0.2.x` canary it is semver-cheap. But if the capability should survive in some form, the cheapest shapes are: keep `path` as a documented opt-in override, or add a `fileName` template with `{createdAt}`/`{reportId}` placeholders.

Related: I could not find a tracker issue specifying this removal (searched both repos), so as it stands the decision was made inside the PR. If there is prior discussion, linking it in the description would settle the question permanently.

## Question: was the aggregate report meant to go away?

The description's "Fixes a `serve` bug along the way" section attributes the old behaviour to `this.outputPath ??=` / `this.opening ??=`. That code does not exist on `main` — `git grep '??=' origin/main -- packages/plugin-reporter` returns nothing.

What `main` actually does: each formatter **accumulates** reports and writes **one aggregate** at `flush` — markdown with a combined rollup (`across N runs`, `formatGeneratedDate` taking the most recent `createdAt`), JSON as an array of *all* reports, CSV as a single stream with one header. So a second workflow's report was never written into the first workflow's directory; it was merged into one document by design.

That matters twice. The justification given for that part of the change does not hold as written, and the actual breaking change is larger than the description states: **the session-level aggregate report is gone** (per-report files, one-element JSON arrays, per-report CSV headers). Right now that only surfaces as an implication in a docs paragraph; it belongs in the breaking-change section next to the `path` removal, since anyone consuming the aggregate markdown or the multi-report JSON array from a `serve` session is affected. The same non-existent baseline is described in the new `write-lifecycle.test.ts` comment ("While the stream lived in an instance field…") — the test is valuable, the history in its comment is not.

## Blocker: `reportsDir: null` crashes registration with a raw `TypeError`

`packages/plugin-reporter/src/index.ts` (schema) + `packages/plugin-reporter/src/report-file-name.ts` (`resolveReportsBaseDirectory`).

`nullable: true` makes `{ reportsDir: null }` **valid** — ajv skips `minLength` for `null` — and then:

```
reportsDir !== undefined && reportsDir.trim() !== ''
// TypeError: Cannot read properties of null (reading 'trim')
```

Verified against this repo's ajv config (`Ajv2020`, `allowUnionTypes`, `allErrors`, `verbose`): `{reportsDir: null}` → VALID, `{reportsDir: ''}` → rejected, and `null.trim()` throws.

The throw happens *outside* `ensureReportsDirectory`'s `try`, so instead of `UnusableReportsDirectoryError` and a clean exit 2 the user gets an unhandled `TypeError`. It is reachable from ordinary YAML: `reportsDir:` with nothing after it parses as `null`.

Suggested fix: drop `nullable: true` from `reportsDir` so validation rejects `null` outright. That also matches the newer house convention documented in `packages/core/src/actions/core-report-convert.action.ts` ("Optional properties are NOT `nullable`: the type never allows `null` … rather than the older `nullable: true` idiom"). Belt-and-braces: widen `resolveReportsBaseDirectory` to accept `string | null | undefined`, since it is also reachable from a directly-constructed formatter. `options-schema.test.ts` covers `''` — worth adding the `null` case beside it.

## Discoverability: users can no longer find their report

Output paths are generated now, and the only place the resolved path appears is `logger.debug`, while the config default is `logLevel: warn` (`packages/common-cli/src/thymian-config-schema.ts`). Before this PR the path was knowable because it was configured; now it is neither configurable nor visible without `--verbose`. The description lists this as an out-of-scope follow-up, but it is created by this change and is roughly one `info` line (or a closing summary line) away from fixed. Given the `path` removal, I would rather not ship the two together.

## Smaller items

- **Docs/code mismatch** (also flagged inline): `unusable-reports-directory-error.md` says an in-flight write failure stays a logged *warning*; all three formatters log via `logger.error`. Same wording drift in the description.
- **Stale example:** `packages/common-cli/test/flags/option-flag.test.ts` still parses `@thymian/plugin-reporter.formatters.markdown.path=report.md`. Harmless to the parser test, but it is the last reference to a removed option — the doc comments in `option-flag.ts` were updated, this one was missed.
- **`flush()`'s return value is now test-only:** `index.ts` discards it, yet every formatter retains `lastOutput` — a full rendered document/payload — for the lifetime of the session purely to serve it. Worth a comment if it stays, since the natural reading is that something in production consumes it.
- **Registration-time side effect:** `ensureReportsDirectory` creates the base directory even for a run that never emits a report, so an opted-in config leaves an empty `.thymian/reports/` behind after an early failure. Does not bite the default config (`formatters: {}` → no check, no directory), so this is cosmetic.

## Suggested path forward

1. Fix the `reportsDir: null` hole (blocking) and add the test.
2. Answer the two questions above; if the file-name removal is intended, say so in the description and drop the `??=` narrative, and move the loss of the session aggregate into the breaking-change section.
3. Fold the `info`-level "wrote report to …" line into this PR rather than a follow-up.
4. Sweep the warning/error wording and the stale `--option` example.

packages/plugin-reporter/src/formatter.ts
```diff
@@ -78,6 +78,34 @@ export function analyze(reports: Report[], logger?: Logger): ReportAnalysis {
...
+ *
+ * Deliberately empty: where a formatter writes is not a per-formatter concern
+ * any more. The plugin-level `reportsDir` option picks the base directory and
+ * every report gets its own directory beneath it, so all formats of one run
+ * stay together. This mirrors the reporter plugin's options schema one-to-one —
+ * that schema is `additionalProperties: false` at every level, so nothing may be
+ * added here that a user cannot set, and values the plugin injects belong in
+ * {@link FormatterRuntimeOptions}.
+ */
+export type FileFormatterOptions = Record<never, never>;
```
Markus-Ende, line 92: **Blocking question.** With this empty and `additionalProperties: false` at every level, plus a hardcoded `REPORT_BASENAME`, there is no longer any way for a user to influence the output file name — not just no way to pin an exact path. Is that the intent, or was the goal only to stop consecutive runs from overwriting each other?

The overwriting bug was in the *default* path, not in the `path` option: a user who set `path` had explicitly chosen a fixed name. Keeping it as an opt-in override is one branch in `resolveReportPath`:

```ts
options.path
  ? resolve(cwd, options.path)
  : resolveReportPath(cwd, reportsDir, report, ext)
```

A `fileName` template (`{createdAt}` / `{reportId}`) would be the other cheap shape. If removal really is the decision, that is defensible at `0.2.x` — please just say so explicitly in the description so the trade is recorded where users will look for it.

packages/plugin-reporter/src/index.ts
```diff
@@ -90,20 +88,31 @@ export const reporterPlugin: ThymianPlugin<ReporterPluginOptions> = {
...
       Object.entries({
         ...(userFormatters ?? {}),
       }).filter(([, options]) => options != null),
     ) as Formatters;
 
+    // `null` is what a YAML `reportsDir:` with no value parses to, and the
+    // schema has to admit it (an optional property must be `nullable: true`).
+    // It means "unset" — normalized once, here, so nothing downstream has to
+    // know that the option can arrive as anything but a string.
+    const reportsBase = reportsDir ?? undefined;
```
Markus-Ende, line 111: This is the right fix for the `null` crash — thanks. Normalizing at the plugin's own boundary keeps `null` from leaking into anything downstream, and the `typeof` guard in `resolveReportsBaseDirectory` still covers direct callers. I verified the premise in the comment: `ThymianPlugin.options` is `JSONSchemaType<Options>` (`packages/core/src/thymian-plugin.ts`), and Ajv's `JSONSchemaType` does require `nullable: true` on optional properties, so the schema genuinely cannot reject `null` up front. Resolved from my side.

packages/plugin-reporter/src/formatters/markdown.ts
```diff
@@ -639,48 +645,116 @@ function buildTestSection(
...
+
+    try {
+      // Rendering is inside the guard too: a malformed report (an execution
+      // shape `analyze` chokes on, a throwing accessor) must degrade exactly
+      // like an unwritable destination rather than reject out of `report()`.
+      const output = this.render(report);
+
+      await mkdir(path.dirname(outputPath), { recursive: true });
+      await writeFile(outputPath, output, 'utf-8');
+      this.logger.debug(`Wrote Markdown report to ${outputPath}.`);
```
Markus-Ende: **Blocking (small).** This is now the only place the output path is ever stated, and the CLI default is `logLevel: warn` (`packages/common-cli/src/thymian-config-schema.ts`), so by default a user cannot discover where their report went. That was fine while the path was configured; it is not once the path is generated and the `path` option is gone.

The description parks this as a follow-up, but this PR is what creates it. One `info`-level line per written report — or one summary line at `flush` — closes it, and keeps the `path` removal from landing together with the invisibility. Same applies to the equivalent lines in `json.ts` and `csv.ts`.

packages/plugin-reporter/test/write-lifecycle.test.ts
```diff
@@ -0,0 +1,256 @@
...
+
+describe('CsvFormatter with two reports in flight at once', () => {
+  it('writes both files completely and ends both streams', async () => {
+    const cwd = await freshCwd('lifecycle-csv-overlap');
+    const logger = new NoopLogger();
+    const debugSpy = vitest.spyOn(logger, 'debug');
+    const formatter = new CsvFormatter(logger);
+    formatter.init({ cwd });
+
+    // Both reports are handed over before either finishes. While the stream
```
Markus-Ende: The test is worth keeping, but this comment describes a state that never existed on `main`: there was no per-report stream in an instance field and no `finally` closing it — `main` opened one stream in `init()` and closed it in `flush()`, so "the first call's `finally` closed the second call's stream" cannot have happened there. Reads like it is documenting an intermediate revision of this branch.

Same root issue as the `??=` narrative in the PR description. Suggest restating it against what the code guarantees now ("two overlapping reports must each get their own stream, both fully written and both ended") rather than against a bug history a future reader cannot find in the log.

packages/plugin-reporter/src/get-formatters.ts
```diff
@@ -22,75 +23,135 @@ export type Formatters = {
...
+  reportsDir?: string,
 ): Promise<Formatter[]> {
+  const entries = Object.entries(config);
+
+  // The precondition runs before anything is constructed: a base directory
+  // that cannot be created is a run whose findings would never reach disk.
+  if (entries.length > 0) {
+    // Once for the shared base, not once per formatter — and only when a
+    // formatter is actually configured, so `formatters: {}` creates nothing.
+    await ensureReportsDirectory(cwd, reportsDir);
```
Markus-Ende, line 150: Non-blocking. Because this runs at registration, a configured formatter now creates the base directory even for a run that dies before emitting any report — leaving an empty `.thymian/reports/` behind. Harmless, and it does not touch the shipped default (`formatters: {}` skips the check entirely), so I would not change the behaviour: the fail-fast guarantee is worth more than the stray directory. Noting it only so it is not mistaken for a bug later.

packages/plugin-reporter/src/formatters/markdown.ts
```diff
@@ -639,48 +645,116 @@ function buildTestSection(
...
-   * config schema stays `{ path }`; defaults to `endpoint`.
+   * config schema stays empty; defaults to `endpoint`.
    */
   private sortReportsBy: SortReportsBy = 'endpoint';
 
+  /**
+   * Document of the most recently written report, handed back by {@link flush}
+   * so a caller that drives a single report still gets the rendered output.
+   */
+  private lastOutput: string | undefined;
```
Markus-Ende, line 676: Non-blocking. `flush()`'s return value has no production consumer left — `index.ts` discards it — yet all three formatters now retain a full rendered document/payload for the lifetime of the session to serve it, which in a long `serve` session is memory held for tests only. (Still strictly better than `main`, which held every report.) Worth a line saying it exists for the `Formatter` contract and its test callers, so the next reader does not go looking for who consumes it.

astro-docs/src/content/docs/references/configuration.mdx
```diff
@@ -166,25 +166,93 @@ plugins:
...
+| `csv`      | `<reportsDir>/<created-at>-<report-id>/report.csv`  | Tabular analysis in spreadsheets or BI tools           |
+| `json`     | `<reportsDir>/<created-at>-<report-id>/report.json` | Automation and post-processing of the full report data |
+
+Formatters take no options of their own — `markdown: {}` selects the formatter, nothing more. There
+is no way to pin one fixed output file, because a fixed file is exactly what makes a second run
+destroy the first one's report.
+
+A downstream step therefore matches a pattern rather than a path: a CI job that uploads or reads
+the output should glob `<reportsDir>/*/report.<ext>` — for example `.thymian/reports/*/report.md`.
+When a step needs exactly one match, point `reportsDir` at a directory that starts the run empty
```
Markus-Ende: This sentence is where the cost of removing `path` lands on the user: a CI step that needs exactly one file now has to guarantee the directory starts empty, or sort the glob by name and take the last. Good that it is stated plainly — but it is worth a concrete two-line snippet here (the glob for upload-artifact, and a `ls -1d "$dir"/*/ | tail -1` for the single-match case), since "point `reportsDir` at a directory that starts the run empty" is advice a user has to translate into a step themselves. This is the same trade as my top-level question 1: if `path` stays as an override, most of this paragraph becomes unnecessary.

Markus-Ende (changes requested):
Re-reviewed at `7fa002fe`. The `reportsDir: null` blocker I raised is **fixed, and fixed well** — normalizing at the plugin boundary (`reportsDir ?? undefined`) plus a `typeof === 'string'` guard in the total helper, with schema, registration, and unit tests around it. The justification in the new comments also checks out: `ThymianPlugin.options` is `JSONSchemaType<Options>` (`packages/core/src/thymian-plugin.ts`), and Ajv's `JSONSchemaType` does force `nullable: true` on optional properties, so admitting `null` at the schema and normalizing it away in code is the right resolution rather than a workaround. The warning/error wording drift in both docs is fixed too.

Requesting changes on what is still open — none of it is code correctness, and all four are small.

## 1. The question I most want answered: is the loss of any file-name control intended?

`REPORT_BASENAME` is hardcoded, the directory is derived from the report, and formatter options are `Record<never, never>` behind `additionalProperties: false` with zero declared properties — so `path`, `fileName`, anything, fails validation, and there is no CLI substitute. This is not "no way to pin an exact path", it is no influence over the name at all.

The description calls it "a deliberate, discussed trade", so this may be precisely the intent — but the overwriting bug it is traded against lives entirely in the **default**: every formatter defaulted to a constant `.thymian/reports/report.{md,csv,json}`, while anyone who set `path` had explicitly opted into a fixed name. Making the default a per-run directory fixes the bug without removing the option (one branch in `resolveReportPath`). The real argument for removal is uniformity plus last-write-wins under `serve` — legitimate, and semver-cheap at `0.2.x` canary, but it is a product decision presented as a technical necessity, and I could not find a tracker issue behind it in either repo.

Either answer is fine; I would just like it settled in the description (and linked to prior discussion if there is any). If the capability should survive in some form, the cheapest shapes are keeping `path` as a documented opt-in override, or a `fileName` template with `{createdAt}`/`{reportId}` placeholders.

## 2. The description's baseline is wrong, and it hides half the breaking change

"Fixes a `serve` bug along the way" attributes the old behaviour to `this.outputPath ??=` / `this.opening ??=`. That code is not on `main` — `git grep '??=' origin/main -- packages/plugin-reporter` returns nothing. What `main` actually does is **accumulate** reports and write **one aggregate** at `flush`: markdown with a combined rollup (`across N runs`, `formatGeneratedDate` taking the most recent `createdAt`), JSON as an array of *all* reports, CSV as a single stream with one header. A second workflow's report was never written into the first workflow's directory — it was merged into one document by design.

So the stated justification for that part of the change does not hold, and the actual breaking change is larger than documented: **the session-level aggregate report is gone.** Anyone consuming the aggregate markdown or the multi-report JSON array from a `serve` session is affected, and right now that only surfaces as an implication in a docs paragraph. It belongs in the breaking-change section beside the `path` removal, since that section is what becomes the migration note.

## 3. Report paths are now undiscoverable

Generated paths, logged only at `debug`, against a default `logLevel: warn` (`packages/common-cli/src/thymian-config-schema.ts`). Before this PR the path was knowable because you configured it; now it is neither configurable nor visible without `--verbose`. Listed as an out-of-scope follow-up, but this change is what created it, and it is about one `info` line away. I would rather not ship the `path` removal and the invisibility together.

## 4. Stale reference

`packages/common-cli/test/flags/option-flag.test.ts:448` still parses `@thymian/plugin-reporter.formatters.markdown.path=report.md` — harmless to the parser test, but it is the last mention of a removed option, and the doc comments in `option-flag.ts` were updated around it.

---

To flip this to an approve: answer 1 in the description, correct the baseline in 2 and move the aggregate loss into the breaking-change list, and fold 3 and 4 in. Two inline notes below are non-blocking observations, marked as such.

The write-lifecycle work remains the strongest part of the PR and is sound independent of all of the above.

packages/plugin-reporter/src/formatter.ts
matthyk: **Answering the blocking question: yes, it is intended.** Losing *any* influence over the output file name is the decision, not a side effect of fixing the default.

You're right that this is a product call and not a technical necessity — an opt-in `path` really is one branch in `resolveReportPath`, and I don't want that framed as impossible. The reasons we're taking it anyway:

- **One layout, one glob.** A consumer learns `<reportsDir>/*/report.<ext>` once and it holds for every format, every run, every command. A per-formatter override means every downstream step has to handle both shapes.
- **Under `serve` a fixed name is last-write-wins.** One reporter instance serves the whole session and each workflow emits its own report, so a pinned `path` silently overwrites itself once per workflow. That's a footgun that only fires in the multi-report case — the case where losing output hurts most — and keeping the option means shipping it knowingly.
- **Cheap to reverse.** We're at `0.2.x`; if the demand shows up, adding it back is additive and non-breaking, whereas removing it later would not be.

Both shapes you proposed are recorded as the follow-up, and we'll open a ticket for them rather than lose them in this thread: `path` as a documented opt-in override, or `fileName` with `{createdAt}`/`{reportId}` placeholders — the latter being the better fit, since it keeps uniqueness by construction instead of trading it away.

packages/plugin-reporter/src/formatters/markdown.ts
matthyk: **Fixed** in bb13eb5f. The per-report success line is now `info` in all three formatters (`markdown.ts`, `json.ts`, `csv.ts`), so the path shows up at the CLI's default `logLevel: warn`... at `info`, which `warn` does not suppress — i.e. visible without `--verbose`.

The three spies that asserted on the level moved with it: `write-lifecycle.test.ts` (the two-in-flight positive assertion and the render-throws negative one) and `csv-write-failure.test.ts`. Agreed on not shipping the `path` removal and the invisibility together.

matthyk: **Correction to my reply above: `info` does not clear the default threshold.** I claimed the line is now visible without a flag — it isn't.

`shouldLog` is `logLevelValues[messageLevel] >= logLevelValues[configuredLevel]` (`packages/core/src/logger/log-level.ts:32`), and `info` is `2` against `warn`'s `3`. At the default an `info` line is still suppressed, so promoting `debug` → `info` does not by itself make the path discoverable.

What it does buy, per `resolveLogLevelWithConfig` (`packages/common-cli/src/base-cli-run-command.ts:475`) — `--log-level` wins, else `--debug` → `debug`, else `--verbose` → `info`, else config, else `warn`:

- Before: the path was at `debug`, so it needed `--debug` / `--log-level debug`. `--verbose` (which resolves to `info`) did **not** show it — meaning the PR description's own "not discoverable without `--verbose`" was already inaccurate.
- After: it shows with `--verbose`, one notch cheaper, and the description's claim becomes true.

So your finding stands on the part that matters: **at the default, the path is still invisible.** None of the one-word fixes are good, though:

- `warn` is visible by default, but a successful write is not a warning and it would land on stderr.
- Flipping the CLI's default `logLevel` to `info` closes it, but also surfaces `Configuration loaded.`, `Resolved N specification(s)`, and every other plugin's `info` output — a CLI-wide UX change this PR has no business making.
- What I'd actually want: print the written paths into the stdout summary `handleWorkflowOutcome` already emits unconditionally (`packages/common-cli/src/workflow-outcome.ts:88`). Log-level independent, and it would give `flush()`'s return value the production consumer it currently lacks — which folds in your `lastOutput` note on the same PR. It needs the paths plumbed from the plugin back to the command, so it's a design change rather than a review fix.

Which is why I'd rather take the third option as the follow-up than bolt one of the first two on here. Your call, and if you want default visibility inside this PR I'll take it — just tell me which shape you prefer.

packages/plugin-reporter/test/write-lifecycle.test.ts
matthyk: **Fixed** in bb13eb5f, and you were more right than the comment let on. It isn't just absent from `main` — it was never in *any* committed revision of this branch. The branch's first commit already had the stream as a local, with a docblock saying so verbatim:

```
git show f2457898:packages/plugin-reporter/src/formatters/csv.ts
  * The stream is a local, not instance state: {@link report} serializes calls,
  * so exactly one write is in flight and nothing else can close or overwrite
  * this stream mid-flight.
```

So the comment documented an uncommitted working state — a bug that exists in no reachable commit, which is the worst kind of comment to leave for someone bisecting. Restated against the invariant instead:

```ts
// Both reports are handed over before either finishes. Each write must own
// its own stream: two complete files, each with its own header and its own
// row, both streams ended, and each success line naming its own path.
// `write()` keeps the stream in a local for exactly this reason — a shared
// instance field would let one write close and clear the other's.
```

packages/plugin-reporter/src/get-formatters.ts
matthyk: Agreed, and behaviour unchanged — thanks for writing it down rather than filing it. The trade is the one you describe: an unusable `reportsDir` has to fail while the plugin is still registering, and the price is a stray empty base directory on a run that dies before emitting anything. Worth it, and confined to configured formatters (`formatters: {}` still creates nothing).

packages/plugin-reporter/src/formatters/markdown.ts
matthyk: **Added** in bb13eb5f, on all three `lastOutput` fields:

```ts
/**
 * ...
 * No production consumer: the reporter plugin discards `flush()`'s return
 * value. It exists for the {@link Formatter} contract and for callers — tests
 * today — that drive one report and assert on the rendering. Bounded to one
 * document on purpose; retaining every report is what this formatter used to
 * do to build a session-level aggregate.
 */
```

Worth noting these two threads point at the same fix: giving `flush()` a real consumer — the CLI printing written paths into its stdout summary — would close the discoverability note as well, which is why I'd rather do that as one follow-up than patch either symptom here.

astro-docs/src/content/docs/references/configuration.mdx
matthyk: **Added** in 829b116f. Both cases now carry the step instead of the advice:

```yaml
# Upload every run directory the job produced.
- uses: actions/upload-artifact@v7
  with:
    name: thymian-reports
    path: .thymian/reports/*/report.*
```

```bash
report="$(ls -1d .thymian/reports/*/ | tail -1)report.md"
```

Three notes on the shape:

- **`tail -1` is stated as the first option, not the empty-directory trick.** You had it as the fallback; I inverted them, because "point `reportsDir` at a directory that starts the run empty" is a *requirement* dressed as advice — in a job with a cached or reused workspace, clearing it is a step of its own. The docs now say that explicitly instead of leaving the reader to discover it.
- **`tail -1` is sound, verified rather than assumed.** `defaultRunDirectoryName` is `<stamp>-<shortId>` where the stamp is `createdAt` through `sanitize` — fixed-width, so lexicographic order is chronological order. It holds on the fallback path too: an unusable `createdAt` falls back to `new Date().toISOString()`, still an ISO stamp, so the ordering survives a malformed report.
- **`upload-artifact` is pinned to `v7`, not the `v4` in your snippet.** `actions/upload-artifact` latest is `v7.0.1`; we have a standing rule against copying an action version from an existing example precisely because that is how a repo ends up three majors behind.

Your closing point stands recorded: if `path` ever returns as an override, this paragraph mostly collapses — noted on the follow-up ticket alongside the `fileName` template.

Markus-Ende:
@matthyk — I picked this up for a re-review and went looking for the originating ticket first. I can't find one anywhere.

**Where I looked.** Every issue, open and closed, matching on title *and* body (GitHub's search index doesn't return results for our private trackers, so this was a full fetch-and-scan): both internal trackers (~750 issues), this repo, and the demo and IntelliJ plugin repos.

Terms tried: `reportsDir`, "reports directory" / "report directory" / "run directory", `overwrit*`, `.thymian/reports`, `report.{md,csv,json}`, `plugin-reporter`, the per-formatter `path` option, and `serve` combined with report.

I also checked this PR's own timeline (no cross-references, no linked issues), both commit messages (no `Closes #…`), and the changed-file list for a committed spec or story doc. Nothing.

The closest neighbours I found — none of which cover this work — are the markdown-formatter finding-flattening issue, the reporter/report-model cleanup batch, switchable reporter verbosity, and "no reporter output when an error is thrown".

**Two questions:**

1. **Is there a ticket I've missed?** Possibly under wording I didn't guess, or somewhere outside the trackers above. If this was genuinely unticketed that's fine — I'll write one retroactively so the review has something to check the behaviour against.

2. **Would an ADR make sense here?** Several things in this PR read as decisions rather than implementation:
   - one plugin-level `reportsDir` replacing the per-formatter `path`, with **no way to pin an exact output file name** — the PR body calls this "a deliberate, discussed trade", and I can't find where that discussion is recorded;
   - `<createdAt>-<reportId prefix>` as the directory identity, and the no-colon invariant that makes it valid on Windows;
   - the deliberate asymmetry between failing fast at registration (exit 2) and degrading to a logged error mid-write — there's a code comment guarding it against being "fixed" into consistency later, which is usually the signal that the reasoning belongs somewhere more durable than a comment.

   Those are the parts a future reader will want the *why* for, and they're hard to reverse once pipelines depend on the layout.

matthyk:
@Markus-Ende On your second question: yes, an ADR makes sense. I've added [ADR-0022](https://github.com/thymianofficial/thymian/blob/feat/reporter-run-directories/docs/arc42/adr/0022-reports-land-in-per-report-run-directories.md) to this branch in a1077e07, status **Accepted**. It records the decisions you listed plus two more:

1. One plugin-level `reportsDir` replaces the per-formatter `path`.
2. Run directories are named `<createdAt>-<reportId prefix>`. The names contain no `:`, so they're valid on Windows, and sorting them as text also sorts them by time.
3. There is no control over the output file name. The ADR says plainly that this is a product decision, not a technical necessity, and gives the reasons.
4. Each report gets its own files, so the `serve` session aggregate is gone. Combining reports is now `report merge`'s job (ADR-0020).
5. An unusable `reportsDir` exits `2` at registration, while a failure during a write only logs an error.

The drawbacks are listed as well: both breaking changes, no retention policy yet, and report paths only visible with `--verbose`.

packages/plugin-reporter/src/index.ts
```diff
@@ -153,8 +162,14 @@ export const reporterPlugin: ThymianPlugin<ReporterPluginOptions> = {
     });
 
     emitter.onAction('core.close', async (_event, ctx) => {
-      await flushReporters();
-      ctx.reply();
+      try {
+        await flushReporters();
+      } finally {
+        // The action must be answered even when a formatter cannot write its
+        // file, otherwise `core.close` waits on a reply that never comes. The
```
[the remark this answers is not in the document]
matthyk: **Fixed** in 9bb555ad, plus a related gap. Reports emitted *before* `core.close` were already covered: the emitter's `Subject` calls listeners synchronously and `report()` extends the queue synchronously, so they are queued by the time `flush()` looks. The real hole was a report queued *after* flush started, e.g. a workflow finishing while `serve` shuts down.

- **`flush()`** in all three formatters now waits until the queue stops growing, not just for the tail it saw on entry.
- **`flushReporters`** now keeps the running flush as a promise instead of a `hasFlushed` flag. In `serve`, SIGINT and `core.exit` can both close the session, and a second close used to be answered immediately while the first flush was still writing. Only the first close reports a failure; a later one just waits.

Tests: `write-lifecycle.test.ts` (a late report, per formatter) and `reporter-close.test.ts` (a concurrent second close). Both fail against the previous code.

packages/plugin-reporter/src/report-file-name.ts
```diff
@@ -0,0 +1,183 @@
...
+/** The naming rule itself, applied once per report by the memoizing wrapper. */
+function deriveRunDirectoryName(report: Report | undefined): string {
+  const stamp = clampStamp(
+    sanitize(report?.createdAt) || sanitize(new Date().toISOString()),
+  );
+  const shortId =
+    sanitize(report?.reportId).replaceAll('-', '').slice(0, SHORT_ID_LENGTH) ||
+    randomShortId();
+
+  return `${stamp}-${shortId}`;
```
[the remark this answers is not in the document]
matthyk: **Not changing this.** A collision needs two reports with the same `createdAt` to the millisecond *and* the same first 8 hex characters of a `randomUUID()`: a 1-in-2³² chance within one millisecond. No code path reuses identity: core's report builder gives every report a fresh `reportId` and `createdAt`, and the `thymian:` input used by `report convert` and `merge` feeds runs into a new report rather than re-emitting the stored report's identity. On the fallback path for a malformed report, the random short ID is memoized per report object, so formatters still agree. The naming rule is recorded in ADR-0022.

packages/plugin-reporter/src/formatters/csv.ts
```diff
@@ -31,118 +34,258 @@ function executionLabel(
...
+    const outputPath = resolveReportPath(
+      this.options.cwd ?? process.cwd(),
+      this.options.reportsDir,
+      report,
+      'csv',
+    );
+
+    // Open — and therefore write the header — before rendering, so a report
+    // that produces no rows still leaves a header-only file behind.
+    const open = await this.openStream(outputPath);
```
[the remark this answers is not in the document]
matthyk: **Fixed** in 9bb555ad. The CSV formatter now renders the rows before it opens the stream. A report that fails to render logs the error and leaves no file behind, and a valid zero-row report still gets its header-only file. The render-failure test in `write-lifecycle.test.ts` now also asserts that no `report.csv` exists.

packages/plugin-reporter/src/get-formatters.ts
```diff
@@ -22,75 +23,135 @@ export type Formatters = {
...
+ * up-front check fails hard.
+ */
+async function ensureReportsDirectory(
+  cwd: string,
+  reportsDir: string | undefined,
+): Promise<void> {
+  const directory = resolveReportsBaseDirectory(cwd, reportsDir);
+
+  try {
+    await mkdir(directory, { recursive: true });
```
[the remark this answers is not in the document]
matthyk: **Fixed** in 9bb555ad. After the `mkdir`, `ensureReportsDirectory` now runs `access(directory, W_OK)` inside the same guard, so an existing read-only base raises `UnusableReportsDirectoryError` and exits `2` at registration. I chose `access` over a create-and-delete probe file because it leaves nothing on disk and catches `EACCES`/`EROFS` on POSIX. The message is now "Cannot create or write to…", and the error doc and ADR-0022 say so. New test in `get-formatters.test.ts`, skipped on Windows and as root, where `chmod` can't make a directory unwritable.

packages/plugin-reporter/src/index.ts
```diff
@@ -79,6 +67,16 @@ export const reporterPlugin: ThymianPlugin<ReporterPluginOptions> = {
...
       },
+      reportsDir: {
+        description:
+          'Base directory report output is written under. A relative path resolves against the run working directory, an absolute one is used as-is. Defaults to .thymian/reports. Every report gets its own directory beneath it, named <createdAt>-<reportId prefix>, holding one report.<ext> per configured formatter — so consecutive runs never overwrite each other and every format of one run stays together. CI should glob <reportsDir>/*/report.<ext>.',
+        nullable: true,
+        type: 'string',
+        // A blank base would resolve to the run working directory itself and
+        // scatter timestamped run directories through the user's project, so
+        // it is rejected here rather than silently defaulted.
+        minLength: 1,
```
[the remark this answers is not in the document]
matthyk: **Fixed** in 9bb555ad. The `reportsDir` schema now carries `pattern: '\\S'` next to `minLength: 1`, so a whitespace-only value fails validation instead of silently falling back to `.thymian/reports`. A test in `options-schema.test.ts` covers `'   '`, and the configuration reference mentions it.

astro-docs/src/content/docs/references/configuration.mdx
```diff
@@ -166,25 +166,111 @@ plugins:
...
+
+### Migrating from the per-formatter `path`
+
+The per-formatter `path` option was removed. A formatter that still sets it fails config
+validation, because formatter options take no properties at all. To migrate:
+
+1. Delete the `path` key — `markdown: {}` is the whole formatter configuration now.
+2. If it pointed somewhere other than the default, set the plugin-level `reportsDir` to that
+   directory instead. It is one base for every formatter, not a per-formatter file.
+3. Where a pipeline read the fixed file, glob `<reportsDir>/*/report.<ext>` instead.
```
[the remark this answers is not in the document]
matthyk: **Fixed** in 9bb555ad. The `report merge` tutorial now enables formatters with `-o '@thymian/plugin-reporter.formatters.markdown={}'`, sets `reportsDir` for the base, and explains that output lands in `<reportsDir>/<run directory>/report.<ext>`, with a link to the layout in the configuration reference. It was the last reference to the removed `path` option in the docs.

PR #390, opened 2026-08-31
fix(plugin-openapi): strip x-* schema extensions; report Ajv compile failures distinctly

atennert:
Fixes thymianofficial/thymian-internal#622

## Problem

`thymian analyze` reported `Response body is not valid JSON.` for responses whose body **is** valid JSON, whenever the OpenAPI document uses Specification Extensions (`x-*`) inside schemas. Two independent defects combined:

1. **Root cause** — `json-schema.processor.ts` stripped OAS-only keywords (`keysToRemove`) but let `x-*` keys fall through into the `ThymianSchema`. `ajv.ts` builds `Ajv2020` with strict mode at its default (`true`), which **throws at compile time** on unknown keywords (`strict mode: unknown keyword: "x-enumNames"`).
2. **Symptom** — `validate-body.ts`/`validate-request-body.ts` wrapped `parse(body)`, `ajv.compile(...)` and `validate(json)` in a single `try`, discarded `err`, and unconditionally blamed the observed body.

## Changes

- **`plugin-openapi`**: `normalizeSchemaObject` now skips `x-*` keys (the existing recursion covers every depth incl. `$defs` and localized `$ref` targets); `xml` added to `keysToRemove` — the last remaining OAS-only keyword Ajv strict mode rejects. Ajv strict mode stays **on**, so genuine schema-authoring typos still surface. Follows the precedent of #634 (`2161395b`), which skips `x-*` at the Responses-Object level.
- **`core`**: the `try` is narrowed to the JSON parse. Schema compilation gets its own failure path (`assertion: 'schema-compilation'`) that carries the Ajv error text and attributes the defect to the API description document. `err` is never discarded. `validate(json)` runs outside any catch-all.
- **`rules-api-description-validation`**: when the failures stem from schema compilation, the umbrella violation message reads "… could not be validated: the schema in the API description document failed to compile" instead of "… does not conform to the schema".

## Acceptance criteria (from the issue)

1. ✅ Bodies validate against schemas carrying `x-*` at any depth (incl. `$defs`/`$ref` targets) — unit tests + Ajv integration test in `json-schema.processor.test.ts`.
2. ✅ "… is not valid JSON." is emitted only when JSON parsing actually fails.
3. ✅ Uncompilable schemas produce a distinct finding including the Ajv error text.
4. ✅ Regression tests cover parse failure, compile failure, and schema violation for both `validate-body.ts` and `validate-request-body.ts`.

## Verification

- `nx run-many -t test lint typecheck --projects=core,plugin-openapi,rules-api-description-validation` — green (388 + 105 + 12 tests).
- End-to-end against the original repro (Clockodo HAR + OpenAPI 3.1 document, 189 `x-*` occurrences): the two false `Response body is not valid JSON.` findings for `GET /api/v2/clock` and `GET /api/v2/entries` are gone (9 → 7 errors); all other findings unchanged. Counter-check with a deliberately tampered body yields a real schema violation (`property "stopped.id" must be integer`), proving body validation is now actually running.

Note (out of scope, observed during verification): Ajv `strictTypes` warnings now print raw to the console because compilation proceeds further than before — a follow-up should route them through the Thymian logger.

### DIFF
--- a/packages/core/src/http-testing/validate/validate-body.ts
+++ b/packages/core/src/http-testing/validate/validate-body.ts
@@ -1,3 +1,4 @@
+import type { ValidateFunction } from 'ajv';
 import { parse } from 'secure-json-parse';
 import { is } from 'type-is';
 
@@ -10,43 +11,11 @@ export function validateJsonBody(
   body: string,
   response: ThymianHttpResponse,
 ): HttpTestCaseResult[] {
+  let json: unknown;
+
   try {
-    const json = parse(body);
-
-    if (!response.schema) {
-      return [
-        {
-          type: 'info',
-          message: 'No response schema is provided.',
-          details: '',
-          timestamp: Date.now(),
-        },
-      ];
-    }
-    const validate = ajv.compile(response.schema);
-
-    validate(json);
-
-    if (validate.errors && validate.errors.length > 0) {
-      // Emit one assertion-failure per schema error instead of collapsing all of
-      // them into a single joined message, so each error is reported on its own.
-      return validate.errors.map((err) => ({
-        type: 'assertion-failure',
-        message: describeSchemaError(err, 'response body'),
-        ...schemaErrorDetail(err),
-        timestamp: Date.now(),
-      }));
-    }
-
-    return [
-      {
-        type: 'assertion-success',
-        message: 'Valid response body.',
-        timestamp: Date.now(),
-      },
-    ];
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  } catch (err) {
+    json = parse(body);
+  } catch {
     return [
       {
         type: 'assertion-failure',
@@ -55,6 +24,55 @@ export function validateJsonBody(
       },
     ];
   }
+
+  if (!response.schema) {
+    return [
+      {
+        type: 'info',
+        message: 'No response schema is provided.',
+        details: '',
+        timestamp: Date.now(),
+      },
+    ];
+  }
+
+  let validate: ValidateFunction;
+
+  try {
+    validate = ajv.compile(response.schema);
+  } catch (err) {
+    // A schema that does not compile is a defect of the API description
+    // document, not of the observed response body.
+    return [
+      {
+        type: 'assertion-failure',
+        assertion: 'schema-compilation',
+        message: `The response schema in the API description document could not be compiled: ${err instanceof Error ? err.message : String(err)}`,
+        timestamp: Date.now(),
+      },
+    ];
+  }
+
+  validate(json);
+
+  if (validate.errors && validate.errors.length > 0) {
+    // Emit one assertion-failure per schema error instead of collapsing all of
+    // them into a single joined message, so each error is reported on its own.
+    return validate.errors.map((err) => ({
+      type: 'assertion-failure',
+      message: describeSchemaError(err, 'response body'),
+      ...schemaErrorDetail(err),
+      timestamp: Date.now(),
+    }));
+  }
+
+  return [
+    {
+      type: 'assertion-success',
+      message: 'Valid response body.',
+      timestamp: Date.now(),
+    },
+  ];
 }
 
 export function validateBodyForResponse(
--- a/packages/core/src/http-testing/validate/validate-request-body.ts
+++ b/packages/core/src/http-testing/validate/validate-request-body.ts
@@ -1,3 +1,4 @@
+import type { ValidateFunction } from 'ajv';
 import { parse } from 'secure-json-parse';
 
 import type { ThymianHttpRequest } from '../../index.js';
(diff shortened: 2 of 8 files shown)


matthyk (commented):
What was the reason why you stripped the `x-*` properties from the schemas, instead of allowing them via Ajv config?

atennert:
> What was the reason why you stripped the `x-*` properties from the schemas, instead of allowing them via Ajv config?

Claude presented me with a choice about this and I decided for the current solution mostly based on the first and second of the following points:

1. **"Allow via Ajv config" existiert für einen Namensraum nicht.** Ajv (hier 8.20) kann unbekannte
   Keywords nur auf zwei Wegen tolerieren: `strictSchema: false | 'log'` (instanzweit) oder
   `addKeyword`/`keywords` mit **exakten** Namen (`keyword: string | string[]` — kein Präfix-/
   Pattern-Support). `x-*` ist aber ein offener Vendor-Namensraum (Clockodo allein: `x-enumNames`,
   `x-precision`, `x-error-category`, …) — exaktes Registrieren skaliert nicht.
2. **`strictSchema: false`/`'log'` kostet echte Fehlererkennung.** Strict mode ist das Einzige, was
   vertippte JSON-Schema-Keywords im Nutzer-Dokument (`requred`, `mimimum`) meldet — eine
   Defektklasse, die Thymian gut melden kann. `'log'` würde zudem roh auf die Konsole schreiben
   (die `strictTypes`-Warnung leakt jetzt schon, siehe Out-of-scope-Note im PR). Issue #622 hat
   beide Optionen explizit gegeneinander abgewogen; das war die bewusste Wahl.
3. **Layering.** `ThymianSchema` ist der normalisierte, formatagnostische Vertrag; der Normalizer
   strippt OAS-only-Keywords (`nullable`, `example`, jetzt `xml`) ohnehin schon — `x-*` ist dieselbe
   Klasse (laut OAS-Spec reine Metadaten ohne Validierungssemantik, Entfernen ändert kein
   Validierungsergebnis). Das Ajv in `core` wird von allen Description-Formaten gefüttert; eine
   OpenAPI-spezifische Toleranz dort würde die Schichtung invertieren.
4. **Präzedenz:** #634 (`2161395b`) strippt `x-*` bereits auf Responses-Object-Ebene.

matthyk:
> > What was the reason why you stripped the `x-*` properties from the schemas, instead of allowing them via Ajv config?
> 
> Claude presented me with a choice about this and I decided for the current solution mostly based on the first and second of the following points:
> 
> 1. **"Allow via Ajv config" existiert für einen Namensraum nicht.** Ajv (hier 8.20) kann unbekannte
>    Keywords nur auf zwei Wegen tolerieren: `strictSchema: false | 'log'` (instanzweit) oder
>    `addKeyword`/`keywords` mit **exakten** Namen (`keyword: string | string[]` — kein Präfix-/
>    Pattern-Support). `x-*` ist aber ein offener Vendor-Namensraum (Clockodo allein: `x-enumNames`,
>    `x-precision`, `x-error-category`, …) — exaktes Registrieren skaliert nicht.
> 2. **`strictSchema: false`/`'log'` kostet echte Fehlererkennung.** Strict mode ist das Einzige, was
>    vertippte JSON-Schema-Keywords im Nutzer-Dokument (`requred`, `mimimum`) meldet — eine
>    Defektklasse, die Thymian gut melden kann. `'log'` würde zudem roh auf die Konsole schreiben
>    (die `strictTypes`-Warnung leakt jetzt schon, siehe Out-of-scope-Note im PR). Issue #622 hat
>    beide Optionen explizit gegeneinander abgewogen; das war die bewusste Wahl.
> 3. **Layering.** `ThymianSchema` ist der normalisierte, formatagnostische Vertrag; der Normalizer
>    strippt OAS-only-Keywords (`nullable`, `example`, jetzt `xml`) ohnehin schon — `x-*` ist dieselbe
>    Klasse (laut OAS-Spec reine Metadaten ohne Validierungssemantik, Entfernen ändert kein
>    Validierungsergebnis). Das Ajv in `core` wird von allen Description-Formaten gefüttert; eine
>    OpenAPI-spezifische Toleranz dort würde die Schichtung invertieren.
> 4. **Präzedenz:** #634 (`2161395b`) strippt `x-*` bereits auf Responses-Object-Ebene.

Sounds reasonable. But please do not remove the `xml` tags. They are supported by Thymian and especially by the `plugin-sampler`. What is the reason you removed it?

atennert:
> Sounds reasonable. But please do not remove the xml tags. They are supported by Thymian and especially by the plugin-sampler. What is the reason you removed it?

The reason is Claudes reasoning ... thank you for confirming my strange feeling about that one. That needs to come back then.

atennert:
Follow-up pushed in d4201412: \`xml\` is back as a preserved field in the ThymianSchema output (plugin-sampler keeps working — it reads \`xml\` to generate XML samples).

To avoid Ajv strict mode throwing on \`xml\` (`strict mode: unknown keyword: "xml"`) without giving up typo detection, the body-validation Ajv instance (\`packages/core/src/http-testing/validate/ajv.ts\`) now registers it as a no-op annotation keyword: \`ajv.addKeyword({ keyword: 'xml', valid: true })\`. This deliberately has no \`schemaType\`, so it tolerates any \`xml\` value shape while genuinely unknown keywords (e.g. a typo like \`requred\`) still surface as a \`schema-compilation\` failure.

Added a regression test in \`validate-body.test.ts\` covering a schema containing \`xml\` validating successfully, and inverted the existing \`json-schema.processor.test.ts\` case to assert \`xml\` is preserved rather than stripped.

matthyk (approved):

PR #396, opened 2026-09-12
feat(core): closed, compile-checked concern-tag vocabulary

BaggersIO:
## Summary

Adds a **concern tag** as a first-class, closed axis on a rule — orthogonal to a rule set's
own topic organization (directories, chapters, ...) — so a rule can say *this is about
security* / *this is about privacy* in a way that's checked, not just a string nobody reads.

- `packages/core/src/rules/rule-tags.ts`: a 14+3-member vocabulary (`security:*`,
  `privacy:*`), declared once as a structured object (for the shape — a third level and a
  bare category on a rule are both unrepresentable) and once as a flat description record
  (for a readable compiler error — TypeScript lists a `keyof typeof <object literal>`
  union's members, correct spelling first, but only names an alias for a mapped type).
  `satisfies` holds the two in lockstep.
- The builder's `.tags()` and `RuleMeta.tags` narrow from `string[]` to the closed `RuleTag`
  union. Both narrowings are free: neither has an existing call site or reader.
- A segment-prefix matcher (`tagMatches`) ships now, specified and tested, even though
  nothing calls it yet — it's the guarantee that a future third tag level is additive rather
  than breaking.
- A generated reference page (`astro-docs/src/content/docs/references/concern-tags.md`),
  wired the same way the existing schema-docs pages are: an Nx target on `core`, run as part
  of the docs site build.
- A `Rule Tag` glossary entry in `CONTEXT.md`.

**Out of scope, deliberately**: nothing consumes a tag yet (no `--tag` filter), and no rule
in the existing corpus is tagged. That's the next slice — this one just makes the axis exist,
be correct, and be readable.

## Test plan

- [x] `nx run core:build`
- [x] `nx run core:typecheck` — the three refusal cases (mistyped member, third level, bare
      category on a rule) compile-fail as `@ts-expect-error`d
- [x] `nx run core:test` — 412 passed, incl. the new `rule-tags.test.ts` (refusals + the
      matcher's 4 cases: exact, category-to-member, future third level, non-segment prefix)
- [x] `nx run core:lint` — no new warnings
- [x] `nx run astro:build` — the generated `concern-tags` page builds and links validate


matthyk (approved):
LGTM. Currently, there is no runtime validation of the provided tags. Let's defer this.

PR #397, opened 2026-09-12
feat(repo): untagged-rule lint lane, warn-level

BaggersIO:
## Summary

A rule author who forgets to classify a rule (see #396 for the concern-tag vocabulary this
builds on) now sees a warning while writing the file, and an author who has considered the
question and concluded nothing fits records that by suppressing the warning — the suppression
comment **is** the "considered, nothing fits" record.

- `tools/eslint-rules/rules/require-rule-tags.js`: a new workspace ESLint rule. Fires on an
  `httpRule(...)` chain with no `.tags(...)` call, or a `.tags()` call with zero arguments.
  Reports on the `httpRule(...)` declaration (not the chain's `.done()`), so a suppression
  comment attaches to the declaration and survives reformatting. Pure syntax — no type
  information needed, so no parser-services cost. A non-literal argument (a shared tag
  constant) is accepted.
- Registered at `warn`, scoped to every `*.rule.ts` file, in the root `eslint.config.mjs`.
  This is the **expand** step: it warns, it does not fail the build. Promoting to a gate is
  the last ticket in this effort, after every rule in the corpus has actually been judged.
- `linterOptions.reportUnusedDisableDirectives: 'warn'` turned on repo-wide, so a suppression
  can't outlive its reason. Verified against the 57 pre-existing `eslint-disable` directives
  elsewhere in the repo — none go stale under it.
- `tools/eslint-rules` is a small hand-rolled Nx project (plain JS, no build step — the root
  ESLint config imports it directly, and CI's lint job never runs a build first) rather than
  the `@nx/eslint:workspace-rule` generator's default scaffold, which pulls in Jest; this repo
  tests exclusively with Vitest. The rule is unit-tested with
  [`@typescript-eslint/rule-tester`](https://typescript-eslint.io/packages/rule-tester/),
  typescript-eslint's ESLint-9-era replacement for the retired built-in `RuleTester`.

## Test plan

- [x] `nx run eslint-rules:test` — 5 cases: no-tags warns once on the declaration; a tagged
      chain is clean; `.tags()` with no arguments warns; a non-literal argument is clean; a
      suppression comment silences the warning
- [x] `nx run eslint-rules:lint`
- [x] `nx run rules-rfc-9110:lint` — 402 warnings (every untagged rule in the corpus; the
      sweep that tags them is the next slice), 0 errors
- [x] `nx run rules-api-description-validation:lint` — 6 warnings, 0 errors
- [x] `nx run-many -t lint` — 20 projects, 0 errors, no unused-disable-directive warnings
      against the pre-existing 57


matthyk (approved):
There is an issue with the eslint rules which is fixed in #398 with commit https://github.com/thymianofficial/thymian/pull/398/commits/1597dc3b9e7f05afaaa9224c39c5177404494d18. So let's merge

PR #398, opened 2026-09-12
feat(rules-rfc-9110): concern tag sweep 1/6 — identifiers, authentication, message-context

BaggersIO:
## Stacked PR — read this first

This branch is stacked on two not-yet-merged PRs: #396 (the concern-tag vocabulary in
`@thymian/core`) and #397 (the `require-rule-tags` lint lane), merged into this branch so the
work here has something to build against. **The diff below therefore includes their commits
too.** The only commit that is actually new here is the last one,
`feat(rules-rfc-9110): concern tag sweep 1/6 — ...` — that's the one worth reviewing; #396 and
#397 already have their own PRs open. Once those merge, this PR's base will move to `main` and
the diff will shrink to just this batch.

## Summary

Every rule file under `packages/rules-rfc-9110/src/rules/identifiers/`, `.../authentication/`
and `.../message-context/` (68 files — the three densest directories in the corpus, judged
first as a pilot batch) is now either classified with a concern tag from core's closed
vocabulary, or explicitly marked judged-untaggable with an
`eslint-disable-next-line thymian-internal/require-rule-tags` suppression comment.

**32 of 68 tagged**: 12/20 `identifiers`, 6/12 `authentication`, 14/36 `message-context`.

Every one of the 68 descriptions was read directly rather than filtered by keyword — a keyword
prefilter is known to miss ~25% of real matches, clustering in rules that describe a
message-framing hazard without ever using the word "smuggling." No rule in *this* batch turned
out to need the `security:request-smuggling` framing-integrity tag, so no rule here carries a
per-rule tag-reason comment; every tag is defended by the vocabulary's own description plus the
disjointness test (see `packages/core/src/rules/rule-tags.ts` and
`.github/skills/add-http-rule-set/reference/concern-tags.md`).

Some notable calls, for reviewers who want to sanity-check the judgment rather than just the
count:
- All six `Referer` rules carry `privacy:referrer`, including the two that only constrain
  *forwarding* rather than leaking (per the mechanism's-whole-surface policy).
- The `userinfo`-in-URI rules carry `security:spoofing` (phishing via an obscured authority) —
  this is the rule that motivated `spoofing` existing in the vocabulary at all.
- The TLS/certificate-verification cluster in `identifiers/authoritative-access/` is tagged
  `security:transport` as a whole surface, including the MUST-provide/MAY-disable
  configuration-setting rules (the setting's existence is treated as the audit-relevant
  surface, not just its use).
- Pure syntax/grammar/interop rules (duplicate URI hosts read as a "misrouting" case per #90,
  auth-scheme grammar requirements, User-Agent version-token formatting, the 100-continue
  timing cluster) are left untagged — topic correctness, not concern.

**Deviation from the #90 census, flagged rather than forced**: the census in
thymian-workspace#93 targets 13/20 for `identifiers` and 15/36 for `message-context`; this
lands at 12 and 14 respectively (`authentication`'s 6/12 matches exactly). Both are a single
tag short, on rules genuinely re-judged as borderline rather than a systematic gap — recorded
here rather than force-fit to hit the number.

A new `packages/rules-rfc-9110/src/tags.test.ts` asserts this batch's exact tag counts against
its own recorded rule-id list (the same drift-guard idiom `profiles.test.ts` already uses), so
it's independently verifiable ahead of the other five batches landing.

## Test plan

- [x] `nx run rules-rfc-9110:build` — every tag string is a valid, non-typo'd `RuleTag` member
- [x] `nx run rules-rfc-9110:test` — 114 passed, including the 5 new tag-sweep assertions
- [x] `nx run rules-rfc-9110:lint` — 336 warnings (0 errors): 334 remaining untagged rules
      outside this batch + 2 pre-existing unrelated warnings; **zero** `require-rule-tags`
      warnings inside `identifiers/`, `authentication/`, `message-context/`


BaggersIO:
Follow-up commit `d68f60ce`: cross-batch review against #399 (routing/representation) surfaced that the TE connection-option rule here is the same hop-by-hop-leak shape that batch 2 tags as `security:request-smuggling` (Upgrade, Keep-Alive/Proxy-Connection). Fixed for consistency — also brings `message-context` to the census's 15/36 exactly.

matthyk (approved):

PR #399, opened 2026-09-12
feat(rules-rfc-9110): concern tag sweep 2/6 — routing, representation-data-and-metadata

BaggersIO:
## Stacked PR — read this first

This branch is stacked on three not-yet-merged PRs: #396 (concern-tag vocabulary), #397
(`require-rule-tags` lint lane), and #398 (batch 1 of this sweep), merged/committed in so
this batch has something to build against. **The diff below therefore includes all of their
commits too.** The only commit that is actually new here is the last one,
`feat(rules-rfc-9110): concern tag sweep 2/6 — ...` — that's the one worth reviewing. Once
#396/#397/#398 merge in order, this PR's base will move to `main` and the diff will shrink to
just this batch.

## Summary

Every rule file under `packages/rules-rfc-9110/src/rules/routing/` and
`.../representation-data-and-metadata/` (84 files) is now either classified with a concern tag
or explicitly marked judged-untaggable with an
`eslint-disable-next-line thymian-internal/require-rule-tags` suppression comment.

**27 of 84 tagged**: 18/48 `routing`, 9/36 `representation-data-and-metadata`.

Every description was read directly, not keyword-filtered. This directory carries most of the
corpus's framing-integrity cluster — 13 of the sweep's `security:request-smuggling` tags land
here, each with a one-line reason comment, since RFC 9110 writes every one of these as message
framing and never as "smuggling":
- The `Connection` header hop-by-hop cluster (parse-and-strip, must-list, must-not-mark
  end-to-end fields as hop-by-hop, `intermediary-must-implement-connection-header`) and the
  analogous `Upgrade`-connection-option rule: a field an intermediary should have stripped or
  scoped to one hop, left to leak to the next, is a two-hops-disagree-about-the-message shape
  regardless of which specific header carries it.
- The `Content-Length` forwarding/mismatch cluster, including the CONNECT-tunnel
  Content-Length rule named explicitly on the parent ticket (thymian-workspace#90) as one of
  the four canonical shapes for this tag.

Other notable calls:
- `intermediary-must-not-forward-message-to-itself` and the whole `Max-Forwards` cluster are
  tagged `security:dos` — an unbounded forwarding loop is a resource-exhaustion hazard, not a
  framing one.
- The `Via` header's pseudonym/comment-stripping rules (firewall internal-host handling,
  replacing a host with a pseudonym, removing comments before forwarding, generating
  identifying comments) are tagged `security:disclosure`; the base "you must send Via at all"
  requirement rules are left untagged — the disclosure surface is what goes *into* Via, not
  the requirement that it exists.
- The two `origin-server-must-reject-*` scheme/certificate rules are `security:transport`.
- `recipient-may-assume-media-type-or-determine-its-type` is the MIME-sniffing permission
  named on #90 as the one case where the permission itself is the hazard —
  `security:content-type`, alongside its mirror-image rule encouraging `Content-Type`
  presence to avoid needing to sniff at all.

**A cross-batch consistency question for the reviewer**: batch 1 (#398,
`sender-must-send-te-connection-option-with-te-header`) left the TE-header analogue of this
batch's `sender-must-send-upgrade-connection-option` rule untagged, reasoning it was
negotiation-shaped rather than framing-shaped. Having now read this batch's parent-ticket text
naming "hop-by-hop connection options" as this directory's cluster, I tagged the Upgrade
version — the same shape, same reasoning, opposite conclusion. Worth deciding one way and
reconciling the other before all six batches land.

**Deviation from the #90 census, flagged rather than forced**: `representation-data-and-metadata`
targets 10/36; this lands at 9. Every content-length/framing case that explicitly names or
implies smuggling risk is tagged; the ones left untagged (`origin-server-should-send-content-length-when-size-known`,
`user-agent-should-send-content-length-for-request-with-defined-content`, the two 304/HEAD
`may-send` rules) are "establishes the mechanism exists" or cache-semantic-accuracy rules
rather than framing-boundary hazards — recorded here rather than stretched to hit the number.

Extends `packages/rules-rfc-9110/src/tags.test.ts` (introduced in #398) with this batch's own
exact-count drift guard, following the same shape.

## Test plan

- [x] `nx run rules-rfc-9110:build` — every tag string is a valid `RuleTag` member
- [x] `nx run rules-rfc-9110:test` — 118 passed (114 from batch 1 + 4 new)
- [x] `nx run rules-rfc-9110:lint` — 0 errors; zero `require-rule-tags` warnings inside
      `routing/` and `representation-data-and-metadata/`; no unused-disable-directive warnings


BaggersIO:
Resolved the TE/Upgrade consistency question raised above: fixed in #398 (commit d68f60ce, now merged into this branch too) — TE now tags `security:request-smuggling` for the same reason Upgrade does here. No other change needed on this PR's side.

matthyk (approved):

PR #61, opened 2025-11-24
Restructure project and module boundaries

Markus-Ende:
Summary of changes:

- Merged `libs/http-status-codes` into `@thymian/core` (`core/src/http-status-codes`).
- updated nx to v22
- fixed watch issue with vite plugin, causing tests to stall
- revised nx module boundaries according to thymianofficial/thymian-internal#68

PR #406, opened 2026-09-14
feat(core): the impossibility vocabulary — closed, compile-checked, exported

BaggersIO:
## Summary

Adds a closed, core-owned vocabulary for *why* an HTTP rule asserts nothing — the reason a
rule is `informational` — and makes `.type('informational', reason, note)` actually compile.
Today `.type()` only takes validation-context names; this is the first half of closing that
gap (see [ADR-0021 §4](https://github.com/thymianofficial/thymian/blob/main/docs/arc42/adr/0021-http-security-rule-sets.md)).

Two tiers, both `as const satisfies` records mirroring the existing concern-tag vocabulary
(`rule-tags.ts`):

- **Tier 1** (`WorldReason`) — makes a rule `informational`, lives in the rule file:
  `only-origin-knows`, `peer-not-observable`,
  `nothing-to-check`, `tool-limitation`.
- **Tier 2** (`ContextReason`) — records why *one* context is missing from an executable
  rule, for a future coverage record; exported now with no caller yet, same as `tagMatches`
  shipped ahead of the tag filter: `participant-not-reachable`, `condition-not-producible`,
  `requires-controlled-input`, `not-representable`, `tool-limitation`.

`tool-limitation` is the only code in both tiers and requires a cited tracker issue — an
uncited one does not compile.

## What changed

- **`rule-impossibility.ts`** (new): the vocabulary, `IssueReference` (a shape guard —
  `` `${string}#${number}` ``), and `RuleImpossibility`, a discriminated union on `reason`
  carried on `RuleMeta`.
- **`rule-builder.ts`**: `.type()` gains overloads for `('informational', reason, note)` and
  `('informational', 'tool-limitation', issue, note)`. The bare `.type('informational')` call
  **stays callable** in this PR — the existing 231 informational rules in `rules-rfc-9110`
  keep compiling unchanged while a separate sweep transcribes their prose justifications into
  this vocabulary. It closes (becomes a compile error) in a follow-up PR once that sweep
  lands.
- **`rule-execution-invariant.ts`**: the runtime invariant gains a violation for a
  hand-constructed rule that carries an impossibility reason on an executable type
  declaration — the type system's guarantee, backstopped for rules built without the
  builder.
- Tests beside the existing `rule-tags.test.ts`/`rule-execution-invariant.test.ts` covering
  every refusal (`@ts-expect-error`) and the metadata a reasoned rule carries.

## Design notes for reviewers

- The `reason` parameter on the non-tool-limitation overload is typed
  `Exclude<WorldReason, 'tool-limitation'>`, not the full `WorldReason` — otherwise a
  `tool-limitation` call missing its required issue would silently match that overload
  instead of being refused. Same reasoning shapes `RuleImpossibility`'s first branch.
- A dedicated overload for the bare `.type('informational')` call is required, not
  incidental: without it, overload resolution falls through to the executable-types overload
  and reports a confusing "'informational' is not assignable to 'ExecutableRuleType'".

## Test plan

- [x] `nx test core` — 586/586 passing
- [x] `nx run core:typecheck` — green
- [x] `nx run core:lint` — no new errors/warnings
- [x] `nx test rules-rfc-9110` — unaffected, 138/138 passing (no rule sets this vocabulary yet)

## Review response (@matthyk)

Two commits added on top, and the rename propagates through the rest of the stack (#407–#414).

**1. The suggestion on the `nothing-to-check` claim — applied.** The claim reads `A statement
that asserts nothing to check.` The `MAY` half was a routing hint rather than part of the
definition, so it moved to `executability-gate.md`, where an author actually picks a code.

**2. The naming — taken, and all three tier-1 codes renamed.** No length limit exists; they are
object keys and union members, so nothing constrains them. But length was not the defect. Tier 2
names *what is missing* (`participant-not-reachable`, `condition-not-producible`,
`not-representable`) while tier 1 named a topic area and left the impossibility to be inferred —
which is why `tool-limitation`, the one tier-1 code that already named its reason, did not read
as complicated. The rule now is that a code compresses its own claim and carries the word that
makes it impossible:

| Was | Now |
| --- | --- |
| `origin-internal-ground-truth` | `only-origin-knows` |
| `peer-internal-behaviour` | `peer-not-observable` |
| `permission-or-statement-of-fact` | `nothing-to-check` |

On whether users see these: they should not. The vocabulary is `code → written claim` precisely
so the coverage README renders the *sentence*. `coverage.ts` and its generator do not exist yet,
so that stays a decision rather than a fact — captured on #61.

**3. ADR-0021 §4 also corrected** (separate commit). It said `recommended` "excludes or
downgrades" a heuristic rule and `strict` "includes it", which reads as a leniency ordering. §3
already says otherwise: `recommended` *promotes*, gated on a concern tag AND non-heuristic
status, and may run a tagged, exactly-observable `SHOULD` at `error` where `strict` holds it at
the source's `warn` — `Vary` is the worked case. #54's constraint 10, which was the source of
that sentence, is amended to match.

Tracker updated: #54, #100, #101, #108 corrected in place; #59 (closed) carries a superseding
note rather than a rewrite.


packages/core/src/rules/rule-impossibility.ts
```diff
@@ -0,0 +1,60 @@
...
+// it is written. See
+// `.github/skills/add-http-rule-set/reference/executability-gate.md` for how
+// to choose a code; adding one later is a free union widening.
+export const tier1ImpossibilityReasons = {
+  'origin-internal-ground-truth':
+    'Only the origin knows the fact the statement is about.',
+  'peer-internal-behaviour':
+    'The obligation is on a peer whose internals are not observable.',
+  'permission-or-statement-of-fact':
+    "A 'MAY', or a statement that asserts nothing to check.",
```
matthyk:
```suggestion
  'permission-or-statement-of-fact': 'A statement that asserts nothing to check.",
```

matthyk (commented):
The existing reasons are fine (except for that one comment). Some ideas for more reasons:

* too many false positive. If a rule is worth to be created but its actual implementation would create a lot of noise by producing a lot of false positives


In general we should maybe think about the naming. The names are pretty hard to understand. So they will be even harder for users to understand. Are we limited by any length? Especially `origin-internal-ground-truth`, `peer-internal-behaviour` and `permission-or-statement-of-fact` sound very complicated.

packages/core/src/rules/rule-impossibility.ts
BaggersIO: Applied — thanks. Landed as:

```ts
'nothing-to-check': 'A statement that asserts nothing to check.',
```

(Your snippet opened with `'` and closed with `"`, so it would not have compiled; with the inner `'MAY'` gone the string no longer needs double quotes at all.)

The `MAY` half was doing routing work rather than definitional work, so it moved to `executability-gate.md` under the tier-1 table — that doc is the decision tree an author picks a code from, and losing "a `MAY` lands here" would have been a regression. The claim string stays a definition.

This also made the `permission-or-` half of the old code name orphaned, which is what tipped the naming question below into a rename. See the review response in the PR description.

BaggersIO:
@matthyk — both points taken, and the second one turned into a rename across the whole stack. Full write-up is in the PR description; the short version:

**"Are we limited by any length?"** No. They are object keys in an `as const satisfies Record<string, string>` and members of a `keyof typeof` union — nothing constrains them, not the type system and not Prettier (the key never wraps, only the claim does). So "shorter because we must" was never on the table.

**But length was not the defect.** Look at the two tiers side by side:

| Tier 2 | Tier 1 (was) |
| --- | --- |
| `participant-not-reachable` | `origin-internal-ground-truth` |
| `condition-not-producible` | `peer-internal-behaviour` |
| `not-representable` | `permission-or-statement-of-fact` |

Every tier-2 code contains the word that says *why it cannot be done*. The tier-1 codes were noun piles naming a **topic area**, leaving the impossibility to be inferred — `origin-internal-ground-truth` never says we cannot see it. Which is also why `tool-limitation` did not draw your complaint despite sitting in the same table: it is the one tier-1 code that already named its reason.

So the rule is now **a code compresses its own claim and carries the word that makes it impossible**, which is what tier 2 was already doing:

| Was | Now | Its claim |
| --- | --- | --- |
| `origin-internal-ground-truth` | `only-origin-knows` | Only the origin knows the fact the statement is about. |
| `peer-internal-behaviour` | `peer-not-observable` | The obligation is on a peer whose internals are not observable. |
| `permission-or-statement-of-fact` | `nothing-to-check` | A statement that asserts nothing to check. |

All three are shorter too (17/19/16 vs 28/23/31), but that is a side-effect.

**On "even harder for users to understand" — users should not see these at all.** The vocabulary is `code → written claim` precisely so the coverage README renders the *sentence*. `coverage.ts` and its generator do not exist yet, so that is still a decision rather than a fact; I have noted it on #61 so the generator renders claims, never codes.

**Timing.** Renaming was free exactly once and this was the window: no rule carried a reason and no `coverage.ts` existed. It is now 225 call sites across nine PRs, and after the sweep merges it would have been a codemod plus a re-review of every one. Good catch, and worth the churn now rather than later.

**One thing your comment led me to that is not a naming issue.** Chasing where these render, I found ADR-0021 §4 claiming `recommended` "excludes or downgrades" a heuristic rule and `strict` "includes it" — which reads as a leniency ordering. §3 already says otherwise: `recommended` **promotes**, gated on a concern tag AND non-heuristic status. `Vary` is the case that makes it concrete — `warn` by keyword, `error` under `recommended` because a missing `Vary` poisons a shared cache. Corrected in its own commit, and #54's constraint 10 (the source of that sentence) is amended to match.

**Still open, and I would value your read:** if `Vary`'s assertion counts as *heuristic* — it infers "the response was tailored" from negotiation request headers being present — then §3's promotion gate forbids exactly the promotion we want, since non-heuristic is a hard condition there. Either heuristic-ness attaches to the assertion and not the selector (so the promotion is legal as written), or the gate needs a third condition for rules where the security consequence justifies `error` despite the guess. Nothing has classified it yet, so it is free to settle — and `Vary` is the right case to settle it on.

**On the "too many false positives" reason you floated:** I would leave it out, and the promotion gate is the reason. Its non-heuristic half *is* a false-positive control — a rule that guesses may never reach `error`, whatever tag it carries. Tier 1 also requires a permanent claim about the world that no release can falsify, and "too noisy" is falsifiable by a narrower scope or a better matcher, which makes it `tool-limitation` (cite an issue) rather than a world-claim. And `value-not-pinned-in-description` was already rejected on an argument that transfers exactly: a document-dependent verdict would make the README lie for half its readers, and false-positive rate is document-dependent the same way. If you have a concrete statement in mind, send it over — it routes to heuristic-at-`hint`, `tool-limitation` + issue, or a runtime `rule-skip`, and if it is really "not worth writing at all" then that is a coverage-record question rather than a vocabulary one.

packages/core/src/rules/rule-impossibility.ts
```diff
@@ -0,0 +1,58 @@
...
+
+// The glossary term: both tiers together. `tool-limitation` is deliberately
+// the only code in both — the distinction it draws (us versus the world) is
+// orthogonal to the distinction between one context and all of them.
+export type ImpossibilityReason = WorldReason | ContextReason;
+
+// A shape guard, not a parser: `#123` and `org/repo#123` are both legal,
+// `soon` is not. `tool-limitation`'s issue is a required argument rather than
+// an optional field precisely so this type has a call site to guard.
+export type IssueReference = `${string}#${number}`;
```
matthyk, line 48: this issue reference type is very github specific, isn't it?

matthyk (approved):
We can also defer the change to the `IssueReference` type. It is non-blocking for me.

PR #407, opened 2026-09-14
feat(thymian): the impossibility reference page, the scaffolder, and the glossary entry

BaggersIO:
## Summary

Stacked on #406 (the impossibility vocabulary core module). Brings the three author-facing
surfaces that pointed at that vocabulary before it had a published form into agreement with
it:

- **The scaffolder.** `thymian generate rule` now prompts for a reason (and, for
  `tool-limitation`, a tracker issue) when the author picks `informational`, and emits
  `.type('informational', reason, note)` — so a generated file compiles once the bare
  `.type('informational')` form stops being callable.
- **The reference page.** The vocabulary renders to the docs site from
  `rule-impossibility.ts` via a new `generate-impossibility-docs` core target, on the exact
  precedent of the existing `generate-rule-tags-docs` target — so the page can't drift from
  the source.
- **The skill.** `reference/executability-gate.md` keeps its procedure and claim tables (an
  author still reads those while judging) and gains a pointer to the generated page as the
  canonical list.

Also lands the `CONTEXT.md` `Impossibility Reason` glossary entry, matching the shape of the
existing `Rule Tag` entry.

## Test plan

- [x] `nx test thymian` — 115/115 passing (2 new: informational+reason, tool-limitation+issue)
- [x] `nx run-many -t build` — green
- [x] `nx run thymian:lint` / `nx run core:lint` — no new errors or warnings
- [x] Generated `astro-docs/src/content/docs/references/impossibility-reasons.md` manually
      inspected for content


CONTEXT.md
```diff
@@ -93,6 +93,17 @@ against tags may be partial (a bare category, or a future third level), which is
...
+The closed vocabulary of _why_ a rule, or one context of an otherwise-executable rule,
+asserts nothing. Two tiers: a **world-claim**, whole-rule and permanent, which is what makes
+a rule `informational`; and a **per-context claim**, recorded in a coverage record, which
+does not affect what the rule executes. `tool-limitation` is the only code in both tiers —
+the one claim about Thymian's own implementation rather than about the world — and it must
+cite a tracker issue, so a fixable gap cannot be written off forever by omission. Owned by
+core with the same closure discipline as `Rule Tag`: adding a code is a union widening.
+_In code_: `WorldReason` (tier 1, narrows `.type('informational', reason, note)`) and
+`ContextReason` (tier 2), both in `packages/core/src/rules/rule-impossibility.ts`.
+
```
matthyk, line 106: We already discussed about it. Let's add this section for now. We should do an overall run about the context.md to remove all direct code referneces.

matthyk (approved):

PR #142, opened 2026-02-16
chore(repo): remove unused dependency

atennert:

PR #115, opened 2026-01-28
chore(repo): read commands from source

matthyk:

PR #408, opened 2026-09-14
feat(rules-rfc-9110): impossibility sweep 1/6: status-codes (45 of 45)

BaggersIO:
## Summary

Batch 1 of 6 of the impossibility-reason sweep. Stacked on #407 (the impossibility
vocabulary + author-facing surfaces). Independent of the other five batches — they touch
disjoint rule directories and can be reviewed in any order.

Transcribes every informational rule in `status-codes` (45 of 402 rules in the package) from
a prose comment into `.type('informational', code, note)`, using the closed vocabulary core
now owns.

## Per-code tally

- `peer-not-observable`: 15
- `nothing-to-check`: 22
- `tool-limitation`: 8 — 3 citing a new issue for the 1xx-interim-response capture gap
  (thymianofficial/thymian-workspace#111), 5 citing a new issue for the multipart/byteranges
  body-parsing gap (thymianofficial/thymian-workspace#112)

Count matches the ticket's predicted 45 exactly; no mis-declaration findings.

## Test plan


packages/rules-rfc-9110/src/rules/status-codes/informational-1xx/proxy-must-forward-1xx-responses.rule.ts
```diff
@@ -3,10 +3,12 @@ import { httpRule } from '@thymian/core';
...
 export default httpRule('rfc9110/proxy-must-forward-1xx-responses')
   .severity('error')
-  // Interim 1xx responses are not represented as discrete transactions in
-  // captured traffic (HAR records the final response), so a proxy's forwarding
-  // of them cannot be observed.
-  .type('informational')
+  .type(
+    'informational',
+    'tool-limitation',
+    'thymianofficial/thymian-workspace#111',
```
matthyk, line 9: Should we add internal references here?

matthyk (changes requested):
I read a lot of something like "no non-conformant condition exists to detect". This does not align with our definition of the severity `may`. If a rule has severity `hint` a violation of this rule does not imply a violation to HTTP but a hint that the protocol provides a mechanism that COULD be used.

BaggersIO:
> I read a lot of something like "no non-conformant condition exists to detect". This does not align with our definition of the severity `may`. If a rule has severity `hint` a violation of this rule does not imply a violation to HTTP but a hint that the protocol provides a mechanism that COULD be used.

Agreed, and the fix belongs in the gate's definition rather than in this batch. The phrasing you are seeing is not the sweep's own idea: thymianofficial/thymian-workspace#59 routed every BCP 14 `MAY` to `permission-or-statement-of-fact` ("nothing is non-conformant either way, so there is no violation to detect"), and after the rename on #406 that routing survived as the sentence in `executability-gate.md`: "a BCP 14 `MAY` is the common case for `nothing-to-check`". The sweep followed it literally, 22 times in this PR and 66 times across the stack.

That routing contradicts our own definition of `hint` (docs: "a suggestion for a truly optional mechanism of HTTP") and constraint 1 of the route (`informational` only where no check is possible in any context). The corpus already disagrees with it too: 29 `hint` rules on `main` are executable `MAY`s that fire when the optional mechanism is absent, e.g. `origin-server-may-generate-server-header-field` with `not(responseHeader('server'))` in `static`, `analytics` and `test`.

**Proposed definition, for discussion:**

1. **A `MAY` is checkable.** The assertion is "the mechanism is used", the finding at `hint` is "the protocol offers this mechanism and it is not being used". A `MAY` rule declares every context in which use / non-use of the mechanism is wire-observable, at `hint`, like any other rule.
2. **A `MAY` goes `informational` only under the same world-claims as everything else.** `peer-not-observable` where the permitted choice is internal to a peer (e.g. `user-agent-may-select-most-appropriate-choice-for-406-response`), `only-origin-knows`, or `tool-limitation` with a cited issue. The keyword is never the reason.
3. **`nothing-to-check` shrinks to bare statements of fact**: definitions and descriptions that name no actor obligation and no actor permission.

**Consequence for this stack**, without adding implementations here: the 66 `hint` rules currently at `nothing-to-check` get reclassified. `peer-not-observable` where the choice is genuinely peer-internal; `tool-limitation` citing one new tracker issue ("MAY mechanisms as `hint` checks") where use of the mechanism is wire-observable and the check is simply not written yet (retry after 413, `Connection: close` on 413, POST→GET after 301, coalesced ranges, ...). Notes drop the "non-conformant" wording. The 231-informational count that #414's gate asserts stays intact, and the implementations follow as their own slice.

**Record to update if we agree**: the routing sentence in `executability-gate.md`, ADR-0021's tier-1 table row for `nothing-to-check`, the same table on thymianofficial/thymian-workspace#100, and a correction note on thymianofficial/thymian-workspace#59.

matthyk:
> > I read a lot of something like "no non-conformant condition exists to detect". This does not align with our definition of the severity `may`. If a rule has severity `hint` a violation of this rule does not imply a violation to HTTP but a hint that the protocol provides a mechanism that COULD be used.
> 
> Agreed, and the fix belongs in the gate's definition rather than in this batch. The phrasing you are seeing is not the sweep's own idea: [thymianofficial/thymian-workspace#59](https://github.com/thymianofficial/thymian-workspace/issues/59) routed every BCP 14 `MAY` to `permission-or-statement-of-fact` ("nothing is non-conformant either way, so there is no violation to detect"), and after the rename on #406 that routing survived as the sentence in `executability-gate.md`: "a BCP 14 `MAY` is the common case for `nothing-to-check`". The sweep followed it literally, 22 times in this PR and 66 times across the stack.
> 
> That routing contradicts our own definition of `hint` (docs: "a suggestion for a truly optional mechanism of HTTP") and constraint 1 of the route (`informational` only where no check is possible in any context). The corpus already disagrees with it too: 29 `hint` rules on `main` are executable `MAY`s that fire when the optional mechanism is absent, e.g. `origin-server-may-generate-server-header-field` with `not(responseHeader('server'))` in `static`, `analytics` and `test`.
> 
> **Proposed definition, for discussion:**
> 
> 1. **A `MAY` is checkable.** The assertion is "the mechanism is used", the finding at `hint` is "the protocol offers this mechanism and it is not being used". A `MAY` rule declares every context in which use / non-use of the mechanism is wire-observable, at `hint`, like any other rule.
> 2. **A `MAY` goes `informational` only under the same world-claims as everything else.** `peer-not-observable` where the permitted choice is internal to a peer (e.g. `user-agent-may-select-most-appropriate-choice-for-406-response`), `only-origin-knows`, or `tool-limitation` with a cited issue. The keyword is never the reason.
> 3. **`nothing-to-check` shrinks to bare statements of fact**: definitions and descriptions that name no actor obligation and no actor permission.
> 
> **Consequence for this stack**, without adding implementations here: the 66 `hint` rules currently at `nothing-to-check` get reclassified. `peer-not-observable` where the choice is genuinely peer-internal; `tool-limitation` citing one new tracker issue ("MAY mechanisms as `hint` checks") where use of the mechanism is wire-observable and the check is simply not written yet (retry after 413, `Connection: close` on 413, POST→GET after 301, coalesced ranges, ...). Notes drop the "non-conformant" wording. The 231-informational count that #414's gate asserts stays intact, and the implementations follow as their own slice.
> 
> **Record to update if we agree**: the routing sentence in `executability-gate.md`, ADR-0021's tier-1 table row for `nothing-to-check`, the same table on [thymianofficial/thymian-workspace#100](https://github.com/thymianofficial/thymian-workspace/issues/100), and a correction note on [thymianofficial/thymian-workspace#59](https://github.com/thymianofficial/thymian-workspace/issues/59).

Let's do this!

packages/rules-rfc-9110/src/rules/status-codes/informational-1xx/proxy-must-forward-1xx-responses.rule.ts
BaggersIO: Good catch, and it is about to get bigger, so I filed it rather than answering it myself: **thymianofficial/thymian-workspace#144**.

`thymianofficial/thymian` is public and `thymianofficial/thymian-workspace` is private, so every one of these citations points a reader of the public source at an issue they cannot open. There are 8 today; the `MAY` reclassification we just agreed adds roughly 50 more.

Nothing is broken yet — the impossibility note and its issue reference are not rendered in reports or on the docs site, so no dead link is actually presented to a user. The cost today is someone reading the source and following a 404.

The three options are in #144: keep private refs as an internal work-tracking device, file capability gaps in the public repo and cite those, or file privately and mirror a public tracking issue. Whichever way it goes it is a one-string-per-file rewrite, so I have kept this stack on the existing convention rather than splitting it across two.

I also noted there your point from #406 that `IssueReference` is very GitHub-specific; the two decisions belong together.

BaggersIO:
Done — this PR and the five behind it are rebased and force-pushed. Ready for another look.

**The fix went into the definition, not just the notes.** `pm/103` now carries a first commit that withdraws the routing which produced the wording you objected to:

- `packages/core/src/rules/rule-impossibility.ts` — `nothing-to-check` now reads *"A statement no HTTP message can conform to or violate. Not a `MAY`: a permission is checkable at `hint`."* This is the record the reference page generates from, so the page is regenerated with it.
- `.github/skills/add-http-rule-set/reference/executability-gate.md` — the sentence "a BCP 14 `MAY` is the common case for `nothing-to-check`" is gone, replaced by the test an author should actually apply: *if the actor exercised or declined this permission, would Thymian see a difference in the messages available to it?*
- `docs/arc42/adr/0021-http-security-rule-sets.md` — the tier-1 row, a paragraph in §4, and a Status History "Amended" row.
- Two test fixtures stopped teaching it (`'A MAY with nothing to check.'`).

**Then the six sweep batches were reclassified.** 78 rules sat at `nothing-to-check`; **62 changed, 16 stayed.**

| Code | Rules |
| --- | --- |
| `tool-limitation` → thymian-workspace#141 (observable, check unwritten) | 41 |
| `peer-not-observable` | 8 |
| `tool-limitation` → #115 / #111 / #142 / #112 / #120 | 3 / 3 / 2 / 2 / 1 |
| `only-origin-knows` | 2 |
| stayed `nothing-to-check` | 16 |

Every "no non-conformant condition" phrasing is gone from the reclassified rules. In this PR specifically, all 22 changed and none stayed.

**One thing your review changed beyond the wording.** Working through the rules showed `nothing-to-check` also has to hold requirements addressed to *whoever writes a specification* — "specifications for newly defined fields SHOULD limit their values to visible US-ASCII", "trailer fields that might be generated more than once MUST be defined as list-based". No HTTP message can conform to or violate those, so they are not `MAY`s and not impossible-to-observe either. That is why 16 stayed rather than the handful I predicted.

**Nothing changed executability**, so #414's gate still asserts 231 informational rules. At the stack tip: 402 rule files, 230 with a code plus the one named #116 exception, `nx test rules-rfc-9110` 144/144, `nx test core` 586/586, `nx test thymian` 115/115, `nx run-many -t build` green.

**Two things I did not change, and filed instead:**

- **thymian-workspace#146** — three rules in `redirection-3xx/` say a redirect follow *cannot* be attributed from captured traffic, which is the opposite of what eight rules in the same directory now say. Attribution is heuristic: fine at `warn`/`hint` per constraint 10, not at `error`, and one of the three is an `error` rule. Worth settling before anyone writes the checks.
- **thymian-workspace#147** — `cache-may-use-responses-to-head-...` carries the GET description and the GET RFC anchor.

Your inline question about internal references is answered above and filed as thymian-workspace#144.

matthyk:
Once we've agreed on https://github.com/thymianofficial/thymian-workspace/issues/144, we can merge it.

BaggersIO:
**Agreed on thymian-workspace#144, so the gate you set is clear.**

Peter and you settled it away from all three options I listed: no public issues, no citation changes. The `tool-limitation` references stay on the private tracker, and the user-facing answer becomes a catalog of packages and rules on the docs page once every slice of the rule-set route is done. Recorded and closed on that issue.

**Nothing in this stack changed for it.** Since your review, the only commit added anywhere is `54c51906` on `pm/104` (PR #409), fixing a HEAD rule that carried the GET description and GET anchor — thymian-workspace#147, found by this batch rather than caused by it.

All nine PRs are green, seventeen checks each.

matthyk (approved):

PR #114, opened 2026-01-28
Mk/feat/analytics context

matthyk:
## Summary

Refactor and enhance HTTP linter architecture with improved API context implementations, database abstraction, and visitor pattern-based filter compilation.

### Database Abstraction Layer
- **Introduced `HttpTransactionRepository` interface** to support multiple database backends beyond SQLite
  - Extracted SQLite-specific implementation into `SqliteHttpTransactionRepository`
  - Moved SQL compilation logic to `packages/http-linter/src/db/sqlite/` directory
  - Created database-agnostic transaction filtering methods
  - Added support for role-based filtering via `HttpParticipantRole[]`

### AnalyticsApiContext Refactoring
- Simplified implementation by using the new repository interface instead of direct SQL compilation
- Removed complex SQL compiler utilities from api-context layer (moved to db/sqlite layer)
- Now leverages `HttpTransactionRepository.readTransactionsByHttpFilter()` for filtering
- Added `httpTransactionToLabel()` utility in `@thymian/core` for consistent transaction labeling

### API Context Architecture Improvements
- Extracted common types into `common-types.ts` (`CommonHttpRequest`, `CommonHttpResponse`)
- Added `RuleViolationLocation` to common types for better error reporting
- Simplified `ApiContext` interface by removing helper methods
- Updated `LiveApiContext` to work with new repository interface
- Refactored `StaticApiContext` to use visitor pattern for filter compilation

### Visitor Pattern Implementation
- **Applied visitor pattern across all context types** for filter expression compilation:
  - Created `http-filter-expression-to-filter.ts` for dynamic filtering
  - Created `http-filter-to-static-by-fn.ts` for groupBy operations
  - Updated `http-filter-to-transaction-validation-fn.ts` for validation logic
  - Moved SQL-specific visitors to `db/sqlite/` directory
- Enables cleaner separation between filter logic and execution context

### HttpTestApiContext Enhancements
- Migrated to flatter directory structure
- Improved validation with better type safety and location tracking
- Better integration with `@thymian/http-testing` pipeline
- Uses visitor pattern for filter/groupBy operations

### Core Utilities
- Added `httpTransactionToLabel()` in `@thymian/core` for consistent transaction formatting
- Created conversion utilities:
  - `utils/http-to-common.ts` - transforms HTTP objects to common types
  - `utils/thymian-to-common.ts` - transforms Thymian format to common types
- Removed old `utils.ts` with mixed responsibilities

### RFC-9110 Rules Adaptation
[...]

PR #409, opened 2026-09-14
feat(rules-rfc-9110): impossibility sweep 2/6: methods, fields (50 of 50)

BaggersIO:
## Summary

Batch 2 of 6 of the impossibility-reason sweep. Stacked on the status-codes batch. Independent
of the other five batches — they touch disjoint rule directories and can be reviewed in any
order.

Transcribes every informational rule in `methods` and `fields` (50 of 402 rules) into
`.type('informational', code, note)`.

## Per-code tally

- `peer-not-observable`: 17 (10 fields, 7 methods)
- `only-origin-knows`: 7 (methods)
- `nothing-to-check`: 11 (6 fields, 5 methods)
- `tool-limitation`: 14 (9 fields, 5 methods) — citing thymianofficial/thymian-workspace#110
  (fuzzing sampler / sending gaps, 4), #113 (new: no CONNECT-tunnel modeling, 2), #114 (new:
  no full-message comparison for TRACE echo validation, 1), #115 (new: no per-field
  grammar/type registry, 7)

Count matches the predicted 50 exactly. One rule's prose maps to no code and is left
unchanged pending the separate mis-declaration audit ADR-0021 records: filed as
thymianofficial/thymian-workspace#116
(`intermediary-must-attempt-to-send-outstanding-data-coming-from-closed-side-for-connect-request`
describes TCP/stream-layer tunnel teardown, which may not be an HTTP-layer conformance
statement at all).

## Test plan


BaggersIO:

BaggersIO:
**One commit added on top, `54c51906`, and the stack behind this branch was rebased and force-pushed.**

It fixes thymian-workspace#147, a defect this batch surfaced rather than caused. `cache-may-use-responses-to-head-for-satisfy-subsequent-head-requests` carried the **GET** rule's sentence verbatim — "a cache MAY use it to satisfy subsequent GET and HEAD requests" — which overstates the permission, since a HEAD response does not satisfy a later GET. Its `.url()` pointed at `#name-get` too. RFC 9110 §9.3.2 is now its description, and the anchor is `#name-head`.

Worth knowing: the impossibility note this batch wrote for that rule was already HEAD-only, written to the rule's id rather than to its stale description. So the file had been contradicting itself, and the reclassification is what made that visible.

Checking for the same copy-paste turned up a second instance: `origin-server-should-not-rely-on-private-agreements-for-head-requests` has the correct text but the same wrong anchor. Only its `.url()` changed. Every other directory under `methods/method-definitions/` is clean.

Nothing else moved: no severity, no context list, no impossibility code. At the tip — 402 rule files, 231 informational, 230 coded plus the one named #116 exception, `nx test rules-rfc-9110` 144/144, `core` 586/586, `thymian` 115/115.

matthyk (approved):

PR #411, opened 2026-09-14
feat(rules-rfc-9110): impossibility sweep 4/6: message-abstraction, conditional-requests (40 of 40)

BaggersIO:
## Summary

Batch 4 of 6 of the impossibility-reason sweep. Stacked on the message-context/routing batch.
Independent of the other five batches — they touch disjoint rule directories and can be
reviewed in any order.

Transcribes every informational rule in `message-abstraction` (21) and `conditional-requests`
(19) — 40 of 402 rules — into `.type('informational', code, note)`. Two prose conventions
converge on the same field: `message-abstraction`'s `//` comments sat above `httpRule()`
(rather than inline in the builder chain, as elsewhere in the corpus); `conditional-requests`'
19 carried a `/** ... */` JSDoc block above `httpRule()` instead. Both are removed entirely
once transcribed.

## Per-code tally

- `peer-not-observable`: 23
- `nothing-to-check`: 13
- `only-origin-knows`: 4
- `tool-limitation`: 0

Count matches the predicted 40 exactly; no mis-declaration findings; no new tracker issues
needed.

Where a removed comment also pointed at the sibling rule that enforces the externally-visible
consequence (most of the If-* precondition-ordering rules), that pointer moved to
`.explanation()` rather than being dropped.

## Test plan


BaggersIO:

matthyk (approved):

PR #413, opened 2026-09-14
feat(rules-rfc-9110): impossibility sweep 6/6 (range/auth/content-negotiation)

BaggersIO:
## Summary

Batch 6 of 6 (the last sweep batch) of the impossibility-reason sweep. Stacked on the
representation/identifiers batch. Independent of the other five batches — they touch disjoint
rule directories and can be reviewed in any order.

Transcribes every informational rule in `range-requests`, `authentication` and
`content-negotiation` — 20 of 402 rules — into `.type('informational', code, note)`. Includes
closing the corpus's one rule with no justification at all —
`user-agent-may-send-preference-headers-for-proactive-negotiation` — which is a MAY
(`nothing-to-check`).

## Per-code tally

- `only-origin-knows`: 2 (both authentication credential-validity rules)
- `peer-not-observable`: 9
- `nothing-to-check`: 9
- `tool-limitation`: 0

Count matches the predicted 20 exactly; no mis-declarations found in this batch; no new tracker issues
needed. Three rules' non-impossibility prose (a pointer to a companion/sibling rule) moved to
`.explanation()` instead of being dropped:
`recipient-should-process-q-parameter-as-weight`,
`server-may-ignore-or-reject-invalid-range-header`, `server-may-ignore-range-header`.

**This is the last of the six sweep batches** — with this merged, every one of the 402 rules
in `rules-rfc-9110` has a final `informational` classification (231 total). One rule corpus-wide
was found to be a genuine mis-declaration candidate (in the methods/fields batch, #409) and is
deliberately left without a reason pending a separate audit — tracked as
thymianofficial/thymian-workspace#116.

## Test plan


BaggersIO:

matthyk (approved):

